Issue #2, opened 2023-01-04
Add google sheet pipeline based on existing singer tap


rudolfix:
**Objective**
The task objective is to build a pipeline for google sheets with functionalities similar to existing singer tap (https://github.com/singer-io/tap-google-sheets). most of the requirements below make our pipeline more user friendly than a tap

In principle, the pipeline `dlt.source` method, for a givem `spreadsheet_id` or just url to a spreadsheet (copied from the browser) , should find all tabs containing data, convert them to separate resources and then yield. Look at the very simple example here (https://github.com/dlt-hub/dlt/blob/devel/docs/examples/google_sheets.py)

**Requirements**
* [x] create `dlt.source` function should accept spredsheet_id/url, a list of tab names (optional, if not present you should take all tabs), and other configuration see below
* [x] users should be able to pass credentials the `dlt` way. see example how google credentials are passed (https://github.com/dlt-hub/dlt/blob/devel/docs/examples/google_sheets.py ). 
* [x] use `replace` as write disposition for all tabs (we may do incremental loads using the state, but later)
* [x] figure out how the singer tap handles headers row. do they autodetect or always use first row as field names? As we yield json documents we need names of the fields so in case there are no headers we must create field names ourselves. let's talk on slack
* [x] you must be able to recognize data types of the fields. how singer tap does that? is there metadata support in google api? `dlt` is able to infer the data types correctly but let's try to do it better
* [x] dates must be handled correctly. we need a clever way to recognize that given field is a date. how singer tap does that?
* [x] name your pipeline `google_sheets`
* [x] remember to add required dependencies (1) with poetry to a right group (2) to requirements.txt. I think you'll need `google-api-python-client` to handle google

Testing
* [x] create a few spreadsheets for testing: with single tab, many tabs, with all possible data types etc.
* [x] create "problematic" datasheets for example with a tab where values are missing or where we have irregular data, charts etc. to see what happens
* [x] write tests that loads and checks if such data is loaded as expected
* [x] we'll have more tests once we start working

**How to start**
I'd start with creating a few test cases, try to run singer tap with them to see the results and then extend the example google sheets

rudolfix:
The naming convention for a table:

1. if the range is a whole sheet: resource is a sheet name
2. if the range is a named range: resource is the name of a range (I know we are not handling them right now, but it will be trivial to add them later)
3. if the range is a range expression then the name is `{sheet_name}_{range_expression}`

[...]

PR #4, opened 2023-01-09
google_sheet pipeline

rudolfix:
completes #2


pipelines/google_sheets/data_validator.py
```diff
@@ -0,0 +1,80 @@
...
+# this function takes an url to a Google spreadsheet and computes the spreadsheet id from it. returns an empty string if the url format is not correct
+# spreadsheet url formula: https://docs.google.com/spreadsheets/d/<spreadsheet_id>/edit
+# @input: url- str
+# @output spreadsheet_id: str
+def process_url(url: str) -> str:
+
+    # this string comes before the id
+    url_id_identifier = "d"
+    # split on the '/'
+    parts = url.split("/")
```
rudolfix, line 19: do not write your own parser. reuse what python has
1. parse the url with https://docs.python.org/3/library/urllib.parse.html
2. get the `path` part
3. use regular expression to get component after `spreadsheets/d/`

did you use chatGPT to generate code? it can actually generate the code you need. my advice is to use it to get a first version of a code that seems like a common problem
image

pipelines/google_sheets/data_validator.py
```diff
@@ -0,0 +1,80 @@
...
+    raise ValueError("Invalid URL. Cannot find spreadsheet ID")
+
+
+# This function receives a serial number which can be an int or float(depending on the serial number) and outputs a datetime object
+# @input: serial_number- int/float. The integer part shows the number of days since December 30th 1899, the decimal part shows the fraction of the day
+# @output: converted_date: datetime object for the same date as the serial number
+def serial_date_to_datetime(serial_number: Union[int, float]) -> datetime:
+
+    # convert starting date to ordinal date: number of days since Jan 1st, 1 AD
+    days_since = datetime(1899, 12, 30).toordinal() + serial_number
```
rudolfix, line 38: do not create this constant date every time function is called. it is a waste of cpu cycles.

rudolfix: btw. in singer code the epoch is different?
```python
excel_epoch = 25569 # 1970-01-01T00:00:00Z, Lotus Notes Serial Number for Epoch Start Date
```

pipelines/google_sheets/google_sheets.py
```diff
@@ -0,0 +1,133 @@
...
+except ModuleNotFoundError:
+    from google_sheets.data_validator import process_url, serial_date_to_datetime
+
+try:
+    from apiclient.discovery import build, Resource
+except ImportError:
+    raise MissingDependencyException("Google Sheets Source", ["google1", "google2"])
+
+# constants
+SPREADSHEET_ID = "1NVxFQYRqrGmur_MeIc4m4ewToF802uy2ObC61HOCstU"
```
rudolfix, line 23: this is test code I assume?

pipelines/google_sheets/google_sheets.py
```diff
@@ -0,0 +1,133 @@
...
+def process_range(sheet_val: Iterator[DictStrAny], sheet_val_sn: Iterator[DictStrAny]) -> Iterator[DictStrAny]:
+
+    # get headers - same for both arrays
+    headers = sheet_val[0]  # list
+    for row, row_sn in zip(sheet_val[1:], sheet_val_sn[1:]):
+        table_dict = {}
+        # process both rows and check for differences to spot dates
+        for val, val_sn, header in zip(row, row_sn, headers):
+            # check that they have different types and then process the dates
+            if type(val) != type(val_sn) and (isinstance(val_sn, int) or isinstance(val_sn, float)):
```
rudolfix, line 43: make it more efficient by
1. not comparing types but makeing sure val is instance of string.
2. `isinstance` takes (int, float) to check multiple types, see docs

pipelines/google_sheets/google_sheets.py
```diff
@@ -0,0 +1,133 @@
...
+    # authenticate to the service using the helper function
+    service = _initialize_sheets(cast(GcpClientCredentialsWithDefault, credentials))
+    sheet = service.spreadsheets()
+
+    # get the id from url if the id was left empty
+    if not spreadsheet_id:
+        spreadsheet_id = process_url(spreadsheet_url)
+
+    # if sheet names is left empty, make an api call and discover all the sheets in the spreadsheeet
+    if (not sheet_names) or len(sheet_names) == 0:
```
rudolfix, line 74: `not sheet_names` suffice here

pipelines/google_sheets/google_sheets.py
```diff
@@ -0,0 +1,133 @@
...
+        sheet_range = values[i]['values']
+        sheet_range_sn = values_date_sn[i]['values']
+
+        # create a resource from processing both sheet/range values
+        yield dlt.resource(process_range(sheet_val=sheet_range, sheet_val_sn=sheet_range_sn),
+                           name=sheet_range_name,
+                           write_disposition="replace")
+
+
+if __name__ == "__main__":
```
rudolfix, line 121: see `chess_pipeline.py` for an example of how to create usage examples. you should not have `__main__` here

pipelines/google_sheets/data_validator.py
redicane: According to the google sheets api documentation, the 1899 date is the correct one. This might be a different Lotus Serial Number format they use. Here is the documentation for google sheets' serial numbers https://developers.google.com/sheets/api/reference/rest/v4/DateTimeRenderOption

pipelines/google_sheets/google_sheets.py
redicane: Yes, it will be cleaned up and more appropriate tests will be set up.

tests/google_sheets/test_google_sheets_pipeline.py
```diff
@@ -0,0 +1,47 @@
...
+    assert 1 == 1
+
+
+@pytest.mark.parametrize('destination_name', ALL_DESTINATIONS)
+def test_full_load():
+    """"
+    Full run of a simple pipeline with all destinations
+    """
+    assert 1 == 1
+
```
rudolfix, line 47: we should test
1. a range that contains all possible sheet data types
2. a range that has only header
3. a range that has only one row that is not header (just single line of data)
4. a range that has some empty cells - for various data types
5. a range that has inconsistent data types ie. a date in first row and then something else (number, string etc.)
6. a range that has whole empty row
7. a range that has many empty rows
8. a range that has two "tables" with data like here https://docs.google.com/spreadsheets/d/11G95oVZjieRhyGqtQMQqlqpxyvWkRXowKE8CtdLtFaU/edit#gid=55249181

pipelines/google_sheets/google_sheets.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    Gets the metadata for the first 2 rows of every range specified. The first row is deduced as the header and the 2nd row specifies the format the rest of the data should follow
+    @:param: spreadsheet_id - the id of the spreadsheet
+    @:param: service - Resource object used by google-api-python-client to make api calls
+    @:param: ranges - List of ranges to get data from. If left empty, every sheet inside the spreadsheet will be included instead. named ranges not supported
+    @:return: ranges_data - A dict where all the range names are the key. The values for each key are the corresponding sheet metadata: sheet_name, headers, values
+    """
+
+    # process metadata ranges so only the first 2 rows are appended
+    ranges_data = {}
```
rudolfix, line 30: this is extremely complicated and hard to test. should we just call the API many times for each range we want to get metadata? it won't impact the performance significantly

imo this is premature optimization

pipelines/google_sheets/google_sheets.py
```diff
@@ -0,0 +1,248 @@
...
+            sheet_name = sheet_m['properties']['title']
+            sheet_names.append(sheet_name)
+        logging.info(f"Found the following sheets {sheet_names}")
+
+    # get metadata on the first 2 rows for every provided range
+    metadata_ranges_all = get_metadata(spreadsheet_id=spreadsheet_id, service=service, ranges=sheet_names)
+
+    # Make an api call to get the data for all sheets and ranges
+    # get dates as serial number
+    values = service.spreadsheets().values().batchGet(
```
rudolfix, line 187: as discussed move it to `process_range`

rudolfix (commented):
let's focus on the test cases. please create the test case documents and share with me

pipelines/google_sheets/data_validator.py
```diff
@@ -0,0 +1,177 @@
...
+def serial_date_to_datetime(serial_number: Union[int, float]) -> datetime:
+    """
+    This function receives a serial number which can be an int or float(depending on the serial number) and outputs a datetime object
+    @:param: serial_number- int/float. The integer part shows the number of days since December 30th 1899, the decimal part shows the fraction of the day
+    @:return: converted_date: datetime object for the same date as the serial number
+    """
+    # TODO: add timezone to data
+
+    # if called with a different data type, return with whatever input was, handled by the dlt pipeline later
+    if not isinstance(serial_number, (int, float)):
```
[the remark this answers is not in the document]
redicane: The first row of data will deduce which column has data types or not. Every field which is expected to be a data type will be directed to this function. If some of the data in a column is not a date despite it being supposed to be a data, it will cause the program to crash. The most common of this errors would be for the user to input a date value only for google_sheets not to recognise it as a date value and thus saving it as a string. I believe it would lead to better user experience to 'tolerate' these mistakes  and return them as the original data type since dlt will handle this anyway.

pipelines/google_sheets/data_validator.py
```diff
@@ -0,0 +1,177 @@
...
+        return serial_number
+
+    # convert starting date to ordinal date: number of days since Jan 1st, 1 AD
+    # the integer part is number of days
+    time_since = SERIAL_NUMBER_ORIGIN_ORDINAL + serial_number
```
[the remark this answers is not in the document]
redicane: That looks much better actually, thanks!

pipelines/google_sheets/data_validator.py
```diff
@@ -3,14 +3,16 @@
...
 
-
-# Date constant- n.o days from 1st Jan 1AD to 30th December 1899
-SERIAL_NUMBER_ORIGIN_ORDINAL = 693594
 # this string comes before the id
 URL_ID_IDENTIFIER = "d"
 # time info
 SECONDS_IN_DAY = 86400
+# TIMEZONE info
+DLT_TIMEZONE = "UTC"
```
rudolfix, line 13: pendulum provided by `dlt` should have UTC by default. does it make any difference when you remove that?

pipelines/google_sheets/data_validator.py
```diff
@@ -25,13 +27,10 @@ def get_data_type(value_list: list[DictStrAny]) -> list[bool]:
...
     for val_dict in value_list:
-        # TODO: try except
-        is_date = "effectiveValue" in val_dict and \
-                  "numberValue" in val_dict['effectiveValue'] and \
-                  "effectiveFormat" in val_dict and \
-                  "numberFormat" in val_dict['effectiveFormat'] and\
-                  "type" in val_dict["effectiveFormat"]["numberFormat"] and\
-                  ("DATE" in val_dict["effectiveFormat"]["numberFormat"]["type"] or "TIME" in val_dict["effectiveFormat"]["numberFormat"]["type"])
+        try:
+            is_date = "DATE" in val_dict["effectiveFormat"]["numberFormat"]["type"] or "TIME" in val_dict["effectiveFormat"]["numberFormat"]["type"]
```
rudolfix, line 31: why are you using `in`? "type" is a string. you could just compare, otherwise it is looking for substrings. is this your intention?

pipelines/google_sheets/data_validator.py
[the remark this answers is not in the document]
rudolfix: right!

pipelines/google_sheets/google_sheets.py
```diff
@@ -270,21 +271,25 @@ def process_range(sheet_val: Iterator[DictStrAny], sheet_meta: Iterator[DictStrA
     # get headers and first line data types which is just Datetime or not datetime so far
     headers = sheet_meta['headers']
     first_line_val_types = sheet_meta['values']
+
+    weird_stuff = [row for row in sheet_val]
```
rudolfix, line 275: what is "weird stuff"? :)

rudolfix (commented):
please see the comments and thanks for the tests!

pipelines/.dlt/example.secrets.toml
```diff
@@ -1,20 +1,19 @@
+### Sources
```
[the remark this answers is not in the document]
adrianbr: @redicane I am not sure what you discussed with Marcin - should this file be removed from the commit?

pipelines/docker-compose.yaml
```diff
@@ -0,0 +1,17 @@
+version: '3'
```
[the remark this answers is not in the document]
adrianbr: @redicane I am not sure what you discussed with Marcin - should this file be removed from the commit?

adrianbr (changes requested):
Looks good other than the 2 doubts, please confirm if those files were to be replaced or not

pipelines/docker-compose.yaml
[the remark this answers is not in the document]
redicane: Yeah, I thought I had removed it but it somehow is still there. Sorry!

redicane (approved):

adrianbr (approved):

PR #13, opened 2023-02-18
Add test for sql database pipeline

steinitzu:
Test case for sql pipeline verifying all tables are loaded.  
Adds a db in the docker postgres which is loaded with mock data and used for source.
https://github.com/dlt-hub/pipelines/issues/10

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -977,6 +977,18 @@ files = [
 ]
 
 [[package]]
+name = "mimesis"
+version = "7.0.0"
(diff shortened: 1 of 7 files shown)


rudolfix (approved):
Thanks for the test, I'm merging this for the same reason that #14 got merged.

my biggest issue is that this works only with postgres. we should run all the tests for all our destinations.

this will require that we install additional SQLAlchemy dialects but I think it is really worthy

I'll summarize it in a ticket

PR #11, opened 2023-02-15
SQL database pipeline

steinitzu:
SQL pipeline using SQLAlchemy. https://github.com/dlt-hub/pipelines/issues/10
So far a basic pipeline with customization for which schema and tables to load from.  
I'll be adding tests

@adrianbr does this look reasonable so far?

PR #14, opened 2023-02-18
Make SQL pipeline resumable.

steinitzu:
I'm experimenting with making the SQL pipeline resumable, so it only reads rows not included in the previous run.  https://github.com/dlt-hub/pipelines/issues/10

Works by setting a `cursor_column` and optionally also `unique_column` (could be pk or some other unique field) in config, e.g. `config.toml`

```toml
[sources.some_table_name]
cursor_column = "updated_at"
unique_column = "id"
```

In dlt.state I store the last value of the cursor column and all unique ids for rows with that same cursor value.  
Good for the case where e.g. cursor is a timestamp for last update, then the unique column makes sure we don't miss rows in case of updates with same timestamp without having to store id of every row loaded in `dlt.state`.

For subsequent runs of the pipeline it generates queries like this to filter out previously loaded:

```sql
SELECT * FROM table_name
WHERE updated_at >= '2023-02-18T01:02:03Z'
AND (updated_at, id) NOT IN (('2023-02-18T01:02:03Z', 1), ('2023-02-18T01:02:03Z', 2), ('2023-02-18T01:02:03Z', 3))
ORDER BY updated_at;
```

### DIFF
--- a/pipelines/sql_database/source.py
+++ b/pipelines/sql_database/source.py
@@ -1,24 +1,89 @@
-from typing import List, Iterator, Dict, Any, Optional
+from typing import List, Iterator, Dict, Any, Optional, Tuple, Union, Mapping, cast, TypedDict
+from collections import defaultdict
 
 import dlt
 from dlt.extract.source import DltResource
 from dlt.common import json
 
-from sqlalchemy import create_engine, MetaData, Table
+from sqlalchemy import create_engine, MetaData, Table, Executable, Column, tuple_
 from sqlalchemy.engine import Engine, Connection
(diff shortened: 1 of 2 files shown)


rudolfix (approved):
I like it and I think some of the ideas here (unique + cursor) we want in dlt core.
@steinitzu I have a few things that I want to improve but they mostly in #11 that already got merged.

here I'd mostly reorganize the code and add a standalone resource for a single table

I will merge this PR and make a ticket with all the improvements then assign to you. This will also help me to align with CI setup

PR #23, opened 2023-02-20
Adds pipedrive custom field's renaming capability

### LINKED ISSUE #9: Add pipedrive customization - Field rename
- Implement custom field rename for pipedrive.
- The fields should be renamed on the initial deployment and the rename “recipe” persisted in state and re-used on subsequent runs (so any field renames later do not change the schema). New custom fields should also be added to the state with their first-encountered name and this renaming would be reused on subsequent runs. In order do rename fields, the user would do it manually or by dropping the state.
- The auto-rename should be optional and it should be easy to toggle on/off for the end user.

nacho0605:
Closes #9

### DIFF
--- a/pipelines/pipedrive/__init__.py
+++ b/pipelines/pipedrive/__init__.py
@@ -99,6 +99,9 @@ def _paginated_get(url, headers, params):
             params['start'] = pagination_info.get("next_start")
 
 
+ENTITY_FIELDS_SUFFIX = "Fields"
+
+
 def _get_endpoint(entity, pipedrive_api_key, extra_params=None, munge_custom_fields=True):
     """
     Generic method to retrieve endpoint data based on the required headers and params.
@@ -119,12 +122,11 @@ def _get_endpoint(entity, pipedrive_api_key, extra_params=None, munge_custom_fie
     url = f'https://app.pipedrive.com/v1/{entity}'
     pages = _paginated_get(url, headers=headers, params=params)
     if munge_custom_fields:
-        fields_suffix = "Fields"
-        if fields_suffix in entity:  # checks if it's an entity fields' endpoint (e.g.: activityFields)
+        if ENTITY_FIELDS_SUFFIX in entity:  # checks if it's an entity fields' endpoint (e.g.: activityFields)
             munging_func = munge_push_func
         else:
             munging_func = pull_munge_func
-            entity = entity[:-1] + fields_suffix  # converts entities' endpoint into entity fields' endpoint
+            entity = entity[:-1] + ENTITY_FIELDS_SUFFIX  # converts entities' endpoint into entity fields' endpoint
         pages = map(functools.partial(munging_func, endpoint=entity), pages)
     yield from pages
 
--- a/pipelines/pipedrive/__init__.py
+++ b/pipelines/pipedrive/__init__.py
@@ -46,17 +46,17 @@ def pipedrive_source(pipedrive_api_key=dlt.secrets.value):
 
     # we need to add entity fields' resources before entities' resources in order to let custom fields data munging work properly
     entity_fields_endpoints = ['activityFields', 'dealFields', 'organizationFields', 'personFields', 'productFields']
-    endpoints_resources = [dlt.resource(_get_endpoint(endpoint, pipedrive_api_key), name=endpoint, write_disposition="replace") for endpoint in entity_fields_endpoints]
+    endpoints_resources = [dlt.resource(_get_endpoint(endpoint, pipedrive_api_key), name=endpoint, write_disposition='replace') for endpoint in entity_fields_endpoints]
 
     entities_endpoints = ['organizations', 'pipelines', 'persons', 'products', 'stages', 'users']
-    endpoints_resources += [dlt.resource(_get_endpoint(endpoint, pipedrive_api_key), name=endpoint, write_disposition="replace") for endpoint in entities_endpoints]
+    endpoints_resources += [dlt.resource(_get_endpoint(endpoint, pipedrive_api_key), name=endpoint, write_disposition='replace') for endpoint in entities_endpoints]
     # add activities
-    activities_resource = dlt.resource(_get_endpoint('activities', pipedrive_api_key, extra_params={'user_id': 0}), name='activities', write_disposition="replace")
+    activities_resource = dlt.resource(_get_endpoint('activities', pipedrive_api_key, extra_params={'user_id': 0}), name='activities', write_disposition='replace')
     endpoints_resources.append(activities_resource)
     # add resources that need 2 requests
 
     # we make the resource explicitly and put it in a variable
-    deals = dlt.resource(_get_endpoint('deals', pipedrive_api_key), name="deals")
+    deals = dlt.resource(_get_endpoint('deals', pipedrive_api_key), name='deals', write_disposition='replace')
 
     endpoints_resources.append(deals)
 
@@ -119,7 +119,13 @@ def _get_endpoint(entity, pipedrive_api_key, extra_params=None, munge_custom_fie
     url = f'https://app.pipedrive.com/v1/{entity}'
     pages = _paginated_get(url, headers=headers, params=params)
     if munge_custom_fields:
-        pages = map(functools.partial(munge_push_func if "Fields" in entity else pull_munge_func, endpoint=entity), pages)
+        fields_suffix = "Fields"
+        if fields_suffix in entity:  # checks if it's an entity fields' endpoint (e.g.: activityFields)
+            munging_func = munge_push_func
+        else:
+            munging_func = pull_munge_func
+            entity = entity[:-1] + fields_suffix  # converts entities' endpoint into entity fields' endpoint
+        pages = map(functools.partial(munging_func, endpoint=entity), pages)
     yield from pages
 
 
@@ -131,7 +137,7 @@ def deals_participants(deals_page, pipedrive_api_key=dlt.secrets.value):
     """
     for row in deals_page:
         endpoint = f"deals/{row['id']}/participants"
-        data = _get_endpoint(endpoint, pipedrive_api_key)
+        data = _get_endpoint(endpoint, pipedrive_api_key, munge_custom_fields=False)
         if data:
             yield from data
 
@@ -144,19 +150,6 @@ def deals_flow(deals_page, pipedrive_api_key=dlt.secrets.value):
     """
     for row in deals_page:
         endpoint = f"deals/{row['id']}/flow"
-        data = _get_endpoint(endpoint, pipedrive_api_key)
+        data = _get_endpoint(endpoint, pipedrive_api_key, munge_custom_fields=False)
         if data:
             yield from data
-
-
-if __name__ == '__main__':
-    # configure the pipeline with your destination details
-    pipeline = dlt.pipeline(pipeline_name='pipedrive', destination='bigquery', dataset_name='pipedrive')
-
-    #data = list(deals_participants())
-    #print(data)
-    # run the pipeline with your parameters
-    load_info = pipeline.run(pipedrive_source())
-
-    # pretty print the information on data that was loaded
-    print(load_info)
--- a/pipelines/pipedrive/__init__.py
+++ b/pipelines/pipedrive/__init__.py
@@ -11,13 +11,16 @@ To get an api key: https://pipedrive.readme.io/docs/how-to-find-the-api-token
 import dlt
 import requests
 
+from pipelines.pipedrive.custom_fields_munger import EndpointType, get_munge_endpoint, get_parse_mapping
+
 
 @dlt.source(name="pipedrive")
-def pipedrive_source(pipedrive_api_key=dlt.secrets.value):
+def pipedrive_source(pipedrive_api_key=dlt.secrets.value, munge_custom_fields=True):
     """
 
     Args:
     pipedrive_api_key: https://pipedrive.readme.io/docs/how-to-find-the-api-token
+    munge_custom_fields: bool
 
     Returns resources:
         activityFields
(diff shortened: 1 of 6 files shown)


pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _munge_push_func(data_item: Dict = None, endpoint: str = '') -> Dict:
+    """
+    Specific function to munge data and push changes to custom fields' mapping
+    The endpoint must be an entity fields' endpoint
+    """
+    if all([data_item, data_item.get('edit_flag'), data_item.get('name'), data_item.get('key'), endpoint]):
+        normalized_name = data_item['name'].strip()  # remove leading and trailing spaces (e.g. 'MULTIPLE OPTIONS ' -> 'multiple_options_')
+        normalized_name = normalized_name.lower()  # typos' workaround (e.g. 'CUSTOMFIELD_DEaL_USER' -> 'customfield_d_ea_l_user')
```
rudolfix, line 39: you can keep original names when renaming. dlt will normalize all those columns anyway. no need to do that

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Specific function to pull changes from custom fields' mapping and munge data
+    The endpoint must be an entity's endpoint
+    """
+    if all([data_item, endpoint]):
+        endpoint = f'{endpoint[:-1]}Fields'  # converts entity's endpoint into entity fields' endpoint
+        data_item_mapping = _custom_fields_mapping.get(endpoint)
+        if data_item_mapping:
+            for hash_string, normalized_name in data_item_mapping.items():
+                if data_item.get(hash_string, KeyError) is not KeyError:  # only check key existence
```
rudolfix, line 60: you do that as follows `hash_string in data_item`

rudolfix (approved):
thanks for contributing! I'd like to restructure the code. the implementation could be much simpler and shorter (IMO):
1. do not change the pipedrive_source() function
2. create another resource in which you get all the rename data for all the enpoints ie. in format used by get_parse_mapping
3. in this resource merge it with `dlt.state`
4. yield the whole mapping as a list of all endpoint mappings
5. plase this resource as first to be executed.

then:

in get_endpoint
```
pages = _paginated_get(url, headers=headers, params=params)
yield from pages
```

check if dlt.state contains rename mapping for `entity` if so map the pages through rename function
pages = map(rename_func, pages)

you can define it inline so it has access to entity and state. function will be similar to _pull_munge_func

I'm also leaving a few code style reviews

----------
we'll need tests for renames, add those to our test pipedrive account and assert in text
---------
please check this doc and `dlt.state` part
https://github.com/dlt-hub/dlt/blob/devel/docs/technical/general_usage.md

here is more on create pipeline
https://github.com/dlt-hub/dlt/blob/devel/docs/technical/create_pipeline.md

rudolfix (changes requested):
I mean: request changes not approve (yet :D)

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,85 @@
...
+    return data_item
+
+
+def get_parse_mapping() -> Generator[Dict, None, None]:
+    """
+    Specific function to parse custom fields' mapping, in order to be stored by dlt
+    """
+    for endpoint, data_item_mapping in _custom_fields_mapping.items():
+        for hash_string, normalized_name in data_item_mapping.items():
+            yield {endpoint: [{'key': normalized_name, 'hash_string': hash_string}]}
```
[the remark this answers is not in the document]
adrianbr: Indeed denormalising it would make it more easily readable in db @nacho0605

pipelines/pipedrive/__init__.py
```diff
@@ -11,13 +11,16 @@
 import dlt
 import requests
 
+from pipelines.pipedrive.custom_fields_munger import EndpointType, get_munge_endpoint, get_parse_mapping
```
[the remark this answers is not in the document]
nacho0605: sure, I forgot to check IDE's import

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,85 @@
...
+
+_custom_fields_mapping = {}
+
+
+class EndpointType(Enum):
+    ENTITY = 'ENTITY'
+    ENTITY_FIELDS = 'ENTITY_FIELDS'
+
+
+def get_munge_endpoint(endpoint: str = '', pipedrive_api_key: str = dlt.secrets.value, extra_params: Dict = None, endpoint_type: EndpointType = None) -> Generator[Dict, None, None]:
```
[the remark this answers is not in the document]
nacho0605: some defensive programming, I didn't realize that all arguments must be provided.
I'll do the refactoring.

pipelines/pipedrive/custom_fields_munger.py
nacho0605: I need to store normalized names in the mapping, in order to replace the hash strings with them in the entities and entity fields tables.
I'll remove the "typos' workaround" line but maybe it would be nice to document it somewhere.
Could it be possible to keep the "strip" line or maybe add it on dlt's "normalize_column_name" method?

pipelines/pipedrive/custom_fields_munger.py
nacho0605: Sure, thanks!

pipelines/pipedrive/custom_fields_munger.py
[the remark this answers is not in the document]
nacho0605: My bad, it should have been implemented this way:
````python
for endpoint, data_item_mapping in _custom_fields_mapping.items():
    yield {endpoint: [{'key': normalized_name, 'hash_string': hash_string} for hash_string, normalized_name in data_item_mapping.items()]}
````
But I'll do the denormalization anyway.

File: pipelines/pipedrive/custom_fields_munger.py [hunk 6]
```diff
@@ -0,0 +1,85 @@
...
+    for endpoint, data_item_mapping in _custom_fields_mapping.items():
+        for hash_string, normalized_name in data_item_mapping.items():
+            yield {endpoint: [{'key': normalized_name, 'hash_string': hash_string}]}
+
+
+class MappingFormat(Enum):
+    JSON = 'JSON'
+
+
+def get_mapping(mapping_format: MappingFormat = None) -> Union[str, Dict]:
```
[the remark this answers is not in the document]
nacho0605: It's just a function to get the mapping as a json string or a as a dictionary, in order to print it or whatever. Should I keep it or remove it?

pipelines/pipedrive/custom_fields_munger.py
rudolfix: I still do not understand why you normalize the names here. You can replace the hashes with non normalized name and then `dlt` will do exactly the same thing you are doing here in normalization stage.

Where you should indeed normalize the names is in `get_parse_mappings` because we want to have a table of normalized names to hashes (honestly I'd include normalized name, non normalized name and hash in this table!)

btw. thanks for pointing our the spaces... I actually change them into "_". I will strip them as you suggest

rudolfix (changes requested):
@nacho0605 thanks for addressing the code issues. we should still work on a code refactor. ping me on slack if you want to discuss my proposed changes. I think we can rename those columns with less code than it is now!
thanks!

pipelines/pipedrive/__init__.py
```diff
@@ -136,7 +148,8 @@ def deals_flow(deals_page, pipedrive_api_key=dlt.secrets.value):
         if data:
             yield from data
 
-if __name__=='__main__':
+
+if __name__ == '__main__':
```
rudolfix, line 152: we really should not have `__main__` in `__init__`! we have `pipedrive_pipeline.py` for that

pipelines/pipedrive/__init__.py
```diff
@@ -108,6 +118,8 @@ def _get_endpoint(entity, pipedrive_api_key, extra_params=None):
         params.update(extra_params)
     url = f'https://app.pipedrive.com/v1/{entity}'
     pages = _paginated_get(url, headers=headers, params=params)
+    if munge_custom_fields:
+        pages = map(functools.partial(munge_push_func if "Fields" in entity else pull_munge_func, endpoint=entity), pages)
```
rudolfix, line 122: explain this code in a comment. what this `Field` means? this line is not self readable

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,56 @@
...
+
+def pull_munge_func(data: List, endpoint: str) -> List:
+    """
+    Specific function to pull changes from custom fields' mapping stored in dlt's state and perform data munging
+    The endpoint must be an entities' endpoint
+    """
+    custom_fields_mapping = dlt.state().get('custom_fields_mapping')
+    if custom_fields_mapping:
+        for data_item in data:
+            endpoint = f'{endpoint[:-1]}Fields'  # converts entities' endpoint into entity fields' endpoint
```
rudolfix, line 40: this looks kind of hacky to me. couldn't you store the correct endpoint name in the state?

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,56 @@
...
+            data_item_mapping = custom_fields_mapping.get(endpoint)
+            if data_item_mapping:
+                for hash_string, names in data_item_mapping.items():
+                    if hash_string in data_item:
+                        data_item[names['normalized_name']] = data_item.pop(hash_string)
```
rudolfix, line 45: I'd still put name here and not `normalized_name`, should be used only in function below.

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,56 @@
+from dlt.common.normalizers.names.snake_case import normalize_column_name
+from typing import Dict, List, Generator, Optional
+
+import dlt
+
+
+def munge_push_func(data: List, endpoint: str) -> List:
```
rudolfix, line 6: if you want your pipeline to be fully typed please add `py.typed` file as explained in new README and run `make lint`. If you use generic types like `List` or `Dict` you should parametrize them

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Specific function to perform data munging and push changes to custom fields' mapping stored in dlt's state
+    The endpoint must be an entity fields' endpoint
+    """
+    custom_fields_mapping = dlt.state().setdefault('custom_fields_mapping', {})
+    for data_item in data:
+        if all([data_item.get('edit_flag'), data_item.get('name'), data_item.get('key')]):
+            if custom_fields_mapping.get(endpoint):
+                if data_item['key'] not in custom_fields_mapping[endpoint]:
```
rudolfix, line 15: I do not understand this line. what `key` does? why are we skipping the update if it exists? also you should update the mapping  in `data_item_mapping` with mapping from state to preserve the original values in state and just add new ones form `data_item_mapping`

pipelines/pipedrive/custom_fields_munger.py
```diff
@@ -0,0 +1,56 @@
...
+                    data_item_mapping = _normalize_map(data_item)
+                    custom_fields_mapping[endpoint].update(data_item_mapping)
+            else:
+                data_item_mapping = _normalize_map(data_item)
+                custom_fields_mapping[endpoint] = data_item_mapping
+            data_item['key'] = custom_fields_mapping[endpoint][data_item['key']]['normalized_name']
```
rudolfix, line 21: why we need this line?

rudolfix (changes requested):
the structure of the code is OK now! thanks!

Please also bump dlt version to 0.2.0a21 in pyproject.toml and include the changed files (also the lock)

pipelines/pipedrive/custom_fields_munger.py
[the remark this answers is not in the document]
nacho0605: Done.

pipelines/pipedrive/custom_fields_munger.py
nacho0605: We've already discussed it, but in a nutshell, because dlt does normalize column names but it doesn't normalize cells' data.

Done!

My pleasure, great!

pipelines/pipedrive/__init__.py
```diff
@@ -41,28 +44,34 @@ def pipedrive_source(pipedrive_api_key=dlt.secrets.value):
...
+    entities_endpoints = ['organizations', 'pipelines', 'persons', 'products', 'stages', 'users']
+    endpoints_resources += [dlt.resource(_get_endpoint(endpoint, pipedrive_api_key), name=endpoint, write_disposition="replace") for endpoint in entities_endpoints]
     # add activities
     activities_resource = dlt.resource(_get_endpoint('activities', pipedrive_api_key, extra_params={'user_id': 0}), name='activities', write_disposition="replace")
-    endpoint_resources.append(activities_resource)
+    endpoints_resources.append(activities_resource)
     # add resources that need 2 requests
 
     # we make the resource explicitly and put it in a variable
     deals = dlt.resource(_get_endpoint('deals', pipedrive_api_key), name="deals")
```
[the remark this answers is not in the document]
nacho0605: Write disposition mode remained unchanged from the beginning, but I'll make the change.

pipelines/pipedrive/__init__.py
nacho0605: Sure, I'll do it.

pipelines/pipedrive/custom_fields_munger.py
nacho0605: True, I'll try a slightly different approach.

pipelines/pipedrive/custom_fields_munger.py
nacho0605: I'll take a look.

pipelines/pipedrive/custom_fields_munger.py
nacho0605: Regarding custom fields, "key" field contains pipedrive's hash string representation of its name.
I assumed that pipedrive's hash strings are meant to be a univoque representation of custom fields names (per endpoint), so dlt's state shouldn't be updated if those values remained unchanged.
First of all I search on/update/add values on dlt's state, and then I end up updating data with dlt's state values.

pipelines/pipedrive/custom_fields_munger.py
nacho0605: Same as before.
First of all I search on/update/add values on dlt's state, and then I end up updating data with dlt's state values.

pipelines/pipedrive/__init__.py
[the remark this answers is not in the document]
rudolfix: @adrianbr you need to see this

pipelines/pipedrive/__init__.py
rudolfix: thanks!

pipelines/pipedrive/custom_fields_munger.py
rudolfix: OK I get it and this is good! please move the explanations above to the comments

pipelines/pipedrive/custom_fields_munger.py
rudolfix: right, but such feature will be useful, I'll added a ticket https://github.com/dlt-hub/dlt/issues/153

pipelines/pipedrive/custom_fields_munger.py
nacho0605: Sure!

PR #29, opened 2023-02-24
Draft: Add zendesk

sbakiu:


rudolfix:
@sbakiu or @redicane could you please share your zendesk credentials with us? We'll add those to CI so your tests will pass there.

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class ZendeskCredentials:
+    """
+    This class is used to store credentials from a dict of Zendesk Credentials
+    Subdomain is always required  and there are 3 alternative authentication ways
+    1. email + token
+    2. email + password
+    3. oauth_token - https://developer.zendesk.com/documentation/live-chat/getting-started/auth/#authorization-code-grant-flow
+    """
```
rudolfix, line 15: Optional[str] == Union[str, None]. better use Optional...

pipelines/zendesk/zendesk.py
```diff
@@ -0,0 +1,184 @@
...
+    @:returns: multiple dlt resources
+    """
+    # Authenticate
+    zen_credentials = ZendeskCredentials(credentials)
+    zendesk_client = auth_zendesk(credentials=zen_credentials)
+    return chats_table(zendesk_client=zendesk_client)
+
+
+@dlt.source
+def zendesk(credentials: Dict[str, str] = dlt.secrets.value, load_all: bool = True,
```
rudolfix, line 36: Please use `ZendeskCredentials` directly. Here you have examples (made for this case specifically). It will handle the auth  the options for you automatically

https://github.com/dlt-hub/dlt/blob/0.2.0a23/docs/technical/secrets_and_config.md#working-with-alternatives-of-credentials-union-types

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,120 @@
+from dlt.common import logger
+from dlt.common.typing import TDataItem, DictStrAny
+from typing import Iterator, Union, Dict, Any
+from zenpy import Zenpy
+from zenpy.lib.api_objects import Ticket
+
+
+class ZendeskCredentials:
```
rudolfix, line 8: define credentials as in this doc and a test mentioned therein
https://github.com/dlt-hub/dlt/blob/0.2.0a23/docs/technical/secrets_and_config.md#working-with-alternatives-of-credentials-union-types

rudolfix (changes requested):

adrianbr (approved):

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+    @:returns: Generator of dicts
+    """
+    # get dlt state
+    custom_tickets = dlt.state().setdefault("ticket_custom_fields", {})
+    # get all custom fields and update state if needed, otherwise just load dicts into tables
+    all_fields = zendesk_client.ticket_fields()
+    for field in all_fields:
+        field_id = str(field.id)
+        # grab id and update state dict if the id is new
+        if not (field_id in custom_tickets):
```
rudolfix, line 93: fyi: no need for parentheses here

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+            custom_tickets[field_id] = field.title
```
rudolfix, line 94: here we we update state only with the initial value. this is fine (preserve schema). but IMO we should yield this initial value in `return_dict` so people have full information in the database

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+    Dlt Resource that also writes to dlt state to update on custom ticket fields
+    @:param zendesk_client: Zenpy Object that is used to make API calls to Zendesk API
+    @:returns: Generator of dicts
+    """
+    # get dlt state
+    custom_tickets = dlt.state().setdefault("ticket_custom_fields", {})
+    # get all custom fields and update state if needed, otherwise just load dicts into tables
+    all_fields = zendesk_client.ticket_fields()
+    for field in all_fields:
```
rudolfix, line 90: there's no need to remove anything from state, right?

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+        field_id = str(field.id)
+        # grab id and update state dict if the id is new
+        if not (field_id in custom_tickets):
+            custom_tickets[field_id] = field.title
+        return_dict = field.to_dict()
+        yield return_dict
+
+
+@dlt.resource(name="tickets", write_disposition="replace")
```
rudolfix, line 99: having a `replace` on a resource that is incremental is questionable to me

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+def ticket_table(zendesk_client: Zenpy, pivot_fields: bool = True, per_page: int = 1000, start_time: DateTime = FIRST_DAY_OF_CURRENT_YEAR) -> Iterator[TDataItem]:
```
rudolfix, line 100: I have doubts if `start_time` is useful right how. It will not help you with incremental loads because you do not know what the start time for the next load should be.

for all tables accepting start time you should
1. if start time is None get it from dlt state or set to default (MILLENIUM)
2. after you yield all tickets write the last timestamp to state

incremental pipeline will just pass None to start time and get just new stuff with each run. @adrianbr was that the intention of it?

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+    all_tickets = zendesk_client.tickets.incremental(paginate_by_time=False,
+                                                     per_page=per_page,
+                                                     start_time=start_time,
+                                                     include=["users", "groups", "organisation", "brands"])
+    for ticket in all_tickets:
+        ticket_dict = process_ticket(ticket=ticket, custom_fields=fields_dict, pivot_fields=pivot_fields)
+        yield ticket_dict
+
+
+@dlt.resource(name="ticket_metric_events", write_disposition="replace")
```
rudolfix, line 121: same comments as above

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+    @:param start_time: Datetime on when to start loading for ticket metric events
+    @:returns: Generator returning dicts for every row of data
+    """
+    all_metric_events = zendesk_client.ticket_metric_events(start_time=start_time)
+    for metric_event in all_metric_events:
+        yield metric_event.to_dict()
+
+
+@dlt.resource(name="chats", write_disposition="replace")
+def chats_table(zendesk_client: Zenpy, start_date: DateTime = FIRST_DAY_OF_CURRENT_YEAR) -> Iterator[TDataItem]:
```
rudolfix, line 135: same as above

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+@dlt.source
+def zendesk_chat(credentials: Union[ZendeskCredentialsOAuth, ZendeskCredentialsToken, ZendeskCredentialsEmailPass] = dlt.secrets.value) -> DltResource:
+    """
+    The source for the dlt pipeline. It returns all the basic information.
+    @:param credentials: read as a dict, as filled in .dlt.secrets.toml
+    @:returns: multiple dlt resources
+    """
+    # Authenticate
+    zendesk_client = auth_zenpy(credentials=credentials)
+    return chats_table(zendesk_client=zendesk_client)
```
rudolfix, line 48: you do not pass `start_time` here. I think that chats must be incremental.

pipelines/zendesk/__init__.py
rudolfix: each incremental table should store its start time separately. and I know that we must do a resource decorator for start value....

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+    try:
+        # use the zenpy endpoint with methods to yield the data
+        resource_api_method = getattr(zendesk_client, f"{resource}")
+        resource_api = resource_api_method()[::per_page]
+    except TypeError:
+        # this resource doesn't support slicing for pagination
+        resource_api_method = getattr(zendesk_client, f"{resource}")
+        resource_api = resource_api_method()
+    except Exception as e:
+        logger.warning(f"Skipped resource: {resource}. Couldn't load!")
```
rudolfix, line 166: swallowing exceptions looks really dangerous. you should log and then raise

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+        # use the zenpy endpoint with methods to yield the data
+        resource_api_method = getattr(zendesk_client, f"{resource}")
+        resource_api_method_incremental = resource_api_method.incremental
+        resource_api = resource_api_method_incremental(start_date=start_date, per_page=per_page)
+    except TypeError:
+        # this resource doesn't support incremental so just load normally
+        resource_api_method = getattr(zendesk_client, f"{resource}")
+        resource_api = resource_api_method()
+    except Exception as e:
+        logger.warning(f"Skipped resource: {resource}. Couldn't load!")
```
rudolfix, line 193: swallowing exceptions looks really dangerous. you should log and then raise

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+            my_pages = [page for page in my_req]
+            for page in my_pages:
+                # multiple records
+                if isinstance(page, list):
+                    for record in page:
+                        yield record
+                else:
+                    yield page
+    except Exception as e:
+        logger.warning(f"Failed to load resource: {talk_endpoint}")
```
rudolfix, line 220: swallowing exceptions looks really dangerous. you should log and then raise

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,100 @@
...
+        # some basic resources return a dict instead of objects
+        for element in resource_api:
+            if isinstance(element, dict):
+                yield element
+                continue
+            try:
+                dict_res = element.to_dict()
+                yield dict_res
+            except AttributeError as e:
+                logger.warning(f"Error! Element {element} of type {type(element)} returned from api {resource_api} has no attribute to_dict")
```
rudolfix, line 99: this looks very dangerous. aren't we expecting this method in all cases?

pipelines/zendesk/helpers/talk_api.py
```diff
@@ -0,0 +1,115 @@
...
+    auth: Tuple[str, TSecretValue]
+
+    def __init__(self, credentials: Union[ZendeskCredentialsOAuth, ZendeskCredentialsToken, ZendeskCredentialsEmailPass]) -> None:
+        """
+        Initializer for the API client which is then used to make API calls to the ZendeskAPI
+        @:param credentials: ZendeskCredentials object which contains the necessary credentials to authenticate to ZendeskAPI
+        @:returns: Nothing since it just initializes API parameters
+        """
+        # set subdomain, this is always needed to configure endpoints
+        self.subdomain = credentials.subdomain
```
rudolfix, line 56: what was the problem with using `zenpy` here? we basically write same auth code twice. can't we use this client instead of zenpy then?

pipelines/zendesk/helpers/talk_api.py
```diff
@@ -0,0 +1,115 @@
...
+        while get_url:
+            try:
+                response = requests.get(get_url, headers=self.headers, auth=self.auth)
+                if response.status_code == ZENDESK_STATUS_CODES["ok"]:
+                    # check if there is a next page and yield the response,
+                    # usually all relevant data is stored in a key with same name as endpoint
+                    response_json = response.json()
+                    get_url = response_json.get("next_page", None)
+                    yield response_json[data_point_name]
+                elif response.status_code == ZENDESK_STATUS_CODES["rate_limit"]:
```
rudolfix, line 89: those are standard HTTP codes, you are OK with just using numbers

tests/zendesk/test_zendesk.py
```diff
@@ -0,0 +1,94 @@
...
+# list expected tables and the number of columns they are supposed to have
+SUPPORT_TABLES = ["ticket_fields", "tickets", "ticket_metric_events", "users", "sla_policies", "groups", "organizations", "brands", "activities", "automations", "custom_agent_roles", "dynamic_content",
+                  "group_memberships", "job_status", "macros", "organization_fields", "organization_memberships", "recipient_addresses", "requests", "satisfaction_ratings", "sharing_agreements", "skips",
+                  "suspended_tickets", "targets", "ticket_forms", "ticket_metrics", "triggers", "user_fields", "views", "tags"]
+CHAT_TABLES = ["chats"]
+TALK_TABLES = ["addresses", "agents_activity", "current_queue_activity", "greeting_categories", "greetings", "ivrs", "lines", "phone_numbers", "settings"]
+ALL_DESTINATIONS = ["postgres"]
+
+
+def create_pipeline(destination_name, dataset_name, full_refresh=True, include_support: bool = False, include_chat: bool = False, include_talk: bool = False):
```
rudolfix, line 17: this is good starting point. we must however test
1. for each incremental table that it loads incrementally. that could be done by inspecting state after running if the correct timestamp is there. and then running again to observe no new records
2. you must make sure that pivoting works

pipelines/zendesk_pipeline.py
```diff
@@ -0,0 +1,15 @@
+import time
+import dlt
+from zendesk import zendesk, zendesk_chat, zendesk_talk
+
+
+if __name__ == "__main__":
```
rudolfix, line 6: this script is really important. this is the place where people start working with your pipeline.

1. one example where we load all the data is ok
2. demonstrate incremental loads
3. demonstrate pivoting the custom fields
4. if we have any other customizations we must demonstrate them as well

pipelines/zendesk/requirements.txt
```diff
@@ -0,0 +1,2 @@
+zenpy==2.0.25
```
rudolfix, line 1: should the version be fixed to this exact value. also please add it to our dev dependencies group as explained here:
https://github.com/dlt-hub/pipelines#pipeline-specific-dependencies

rudolfix (changes requested):

pipelines/zendesk/__init__.py
redicane: As in add another key to the dictionary: "initial_field_name" : field_name?

pipelines/zendesk/__init__.py
redicane: Yes, I see no need to remove any of the original fields so far.

pipelines/zendesk/__init__.py
redicane: This was a temporary fix to make the resources failure resilient, since we don't want the pipeline to completely fail when a single resource fails.

pipelines/zendesk/helpers/api_helpers.py
redicane: we are, it's fixed now

pipelines/zendesk/__init__.py
rudolfix: we are working on requests drop in replacement that will do that for you

redicane: Ok, then I shall just remove this part in and similar parts in resource code.

pipelines/zendesk/helpers/talk_api.py
redicane: ZendeskTalk is not functional in Zenpy despite part of it showing up as implemented.

pipelines/zendesk/helpers/talk_api.py
redicane: I just left those for readability and making hackability of code easier for people who might not be as experienced with HTTP codes.

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+    ticket_custom_fields = dlt.state().setdefault("ticket_custom_fields", {})
+    # get all custom fields and update state if needed, otherwise just load dicts into tables
+    all_fields = zendesk_client.ticket_fields()
+    for field in all_fields:
+        return_dict = field.to_dict()
+        field_id = str(field.id)
+        # grab id and update state dict if the id is new, add a new key to indicate that this is the initial value for title
+        if not (field_id in ticket_custom_fields):
+            ticket_custom_fields[field_id] = field.title
+            return_dict["initial_title"] = field.title
```
rudolfix, line 109: 👍

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+    fields_dict = dlt.state().setdefault("ticket_custom_fields", {})
+    all_tickets = zendesk_client.tickets.incremental(paginate_by_time=False,
+                                                     per_page=per_page,
+                                                     start_time=start_time,
+                                                     include=["users", "groups", "organisation", "brands"])
+    for ticket in all_tickets:
+        ticket_dict = process_ticket(ticket=ticket, custom_fields=fields_dict, pivot_fields=pivot_fields)
+        yield ticket_dict
+    # update the last load time in the dlt state
+    dlt.state()["last_load_tickets"] = now().timestamp()
```
rudolfix, line 141: is this really now()? why not the last timestamp in the ticket? will it return duplicate tickets? even if so you have better chances of not losing anything if you add a milisecond to the last timestamp. or you just remove the last ticket in the list so you'll get it next time

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+@dlt.resource(name="ticket_metric_events", write_disposition="replace")
```
rudolfix, line 144: you use state below but resource is replace?

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+    # grab the start time from last dlt load if not filled, if that is also empty then use the first day of the millennium as the start time instead
+    # because of Zenpy this needs to be a Datetime instead of a timestamp
+    if not start_time:
+        # convert from timestamp to datetime object since it is required by Zenpy
+        start_time = from_timestamp(dlt.state().setdefault("last_load_ticket_metric_events", FIRST_DAY_OF_MILLENNIUM.timestamp()))
+    all_metric_events = zendesk_client.ticket_metric_events(start_time=start_time)
+    for metric_event in all_metric_events:
+        yield metric_event.to_dict()
+    # update the last load time in the dlt state, save as timestamp since datetime not supported by dlt state
+    dlt.state()["last_load_ticket_metric_events"] = now().timestamp()
```
rudolfix, line 162: same as above

pipelines/zendesk/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+    @:param talk_endpoint_name: The name of the talk_endpoint
+    @:param talk_endpoint: The actual url ending of this endpoint
+    @:returns: Generator of dicts
+    """
+    # grab the start time from last dlt load if not filled, if that is also empty then use the first day of the millennium as the start time instead
+    if not start_date:
+        final_start_date = dlt.state().setdefault(f"last_load_talk_{talk_endpoint_name}", FIRST_DAY_OF_MILLENNIUM.timestamp())
+    else:
+        final_start_date = start_date.timestamp()
+    my_req = zendesk_client.make_request_incremental(endpoint=talk_endpoint, data_point_name=talk_endpoint_name, start_date=final_start_date)
```
rudolfix, line 241: can't you just call `talk_resource` instead of duplicating code? parametrize it if necessary

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+    method of Ticket class doesn't return all the required information
+    @:param sample_ticket: Ticket Object returned by a Zenpy API call, contains info on a single ticket.
+    @:param custom_fields: A dict containing all the custom fields available for tickets.
+    @:param pivot_fields: Bool that indicates whether to pivot all custom fields or not.
+    @:return base_dict: A dict containing 'cleaned' data about a ticket.
+    """
+
+    base_dict: DictStrAny = ticket.to_dict()
+    # get tags as a string eliminating the square brackets from string
+    base_dict["tags"] = str(base_dict["tags"])[1:-1]
```
rudolfix, line 57: was this in the requirements? what is the problem with a list? if you want to generate a JSON type column that will be stored as struct in the destination add `tags` to `columns` in the resource declaration with type "complex" ie.
`columns={"tags": {"data_type": "complex"}}`

if that was your intention then tell me if the above does not work. I'd like to try it out :)

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # pivot custom field if indicated as such
+    # get custom fields
+    for custom_field in base_dict["custom_fields"]:
+        if pivot_fields:
+            cus_field_id = str(custom_field["id"])
+            field_name = custom_fields[cus_field_id]
+            base_dict[field_name] = custom_field["value"]
+        else:
+            custom_field["ticket_id"] = ticket.id
```
rudolfix, line 67: why are we setting constant value in a loop?

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+    elif isinstance(credentials, ZendeskCredentialsEmailPass):
+        zendesk_client = Zenpy(
+            email=credentials.email, subdomain=credentials.subdomain, password=credentials.password,
+            domain=domain, timeout=timeout, ratelimit_budget=ratelimit_budget, proactive_ratelimit=proactive_ratelimit, proactive_ratelimit_request_interval=proactive_ratelimit_request_interval,
+            disable_cache=disable_cache
+        )
+    return zendesk_client
+
+
+def process_ticket(ticket: Ticket, custom_fields: DictStrStr, pivot_fields: bool = True) -> DictStrAny:
```
rudolfix, line 45: rename `pivot_fields` to `pivot_custom_fields`

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+            base_dict[field_name] = custom_field["value"]
+        else:
+            custom_field["ticket_id"] = ticket.id
+
+    # delete fields that are not needed for pivoting
+    if pivot_fields:
+        del base_dict["custom_fields"]
+    else:
+        # un pivoting will simply save the dict as a json string
+        base_dict["custom_fields"] = str(base_dict["custom_fields"])
```
rudolfix, line 74: do the same trick on `custom_fields` as with `tags`. you'll get a nice queryable JSON in the destination :)

pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+    del base_dict["fields"]
+
+    # modify dates to return datetime objects instead
+    base_dict["updated_at"] = ticket.updated
```
rudolfix, line 78: why? what is in the "updated_at"?

pipelines/zendesk/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pendulum
```
rudolfix, line 1: please remove pendulum and python-dlt. they are in deps already

pipelines/zendesk/requirements.txt
```diff
@@ -0,0 +1,3 @@
...
+python-dlt
+zenpy==2.0.25
```
rudolfix, line 3: at least make it ^2.0.25 or even >= do not bind to exact version (if there's no reason)

tests/zendesk/test_zendesk.py
```diff
@@ -0,0 +1,144 @@
...
+TALK_TABLES = ["calls", "addresses", "agents_activity", "current_queue_activity", "greeting_categories", "greetings", "ivrs", "lines", "phone_numbers", "settings", "calls_incremental",
+               "legs_incremental"]
+# all the timezones saved in dlt state
+INCREMENTAL_SAVED_KEYS = ["last_load_tickets", "last_load_ticket_metric_events", "last_load_chats", "last_load_talk_calls", "last_load_talk_legs"]
+
+
+@pytest.mark.parametrize("destination_name", ALL_DESTINATIONS)
+def test_pivoting_tickets(destination_name: str) -> None:
+    """
+    Checks to see that the tickets table uses pivoting
```
rudolfix, line 24: this also tests the column rename for custom field right?

tests/zendesk/test_zendesk.py
```diff
@@ -0,0 +1,144 @@
...
+    # run pipeline
+    pipeline_incremental = dlt.pipeline(destination=destination_name, full_refresh=True, dataset_name="test_incremental")
+    data_support = zendesk_support(load_all=False)
+    data_chat = zendesk_chat()
+    data_talk = zendesk_talk()
+    info_incremental = pipeline_incremental.run([data_support, data_chat, data_talk])
+    # check that the expected keys are saved now in dlt state
+    for saved_timezone in INCREMENTAL_SAVED_KEYS:
+        assert isinstance(dlt.state()["zendesk"][saved_timezone], float)
+    assert_load_info(info_incremental, expected_load_packages=3)
```
rudolfix, line 64: at least load it again and check if there are no additional records.
see github for similar test:
https://github.com/dlt-hub/pipelines/blob/master/tests/github/test_github_pipeline.py#L36

and load_table_counts and load_table_distinct_counts that will help you to test against duplicates

tests/zendesk/test_zendesk.py
```diff
@@ -0,0 +1,144 @@
...
+
+    schema = pipeline.default_schema
+    user_tables = schema.all_tables()
+    num_proper_tables = 0
+    for table in user_tables:
+        table_name = table["name"]
+        if not ("__" in table_name):
+            assert table_name in tables
+            num_proper_tables += 1
+    assert num_proper_tables == len(tables)
```
rudolfix, line 144: you can also use `load_table_counts` to get data from destination and make sure all tables are loaded

rudolfix (changes requested):

pipelines/zendesk/__init__.py
redicane: The way I thought about this would be the most straightforward way of saving the last load time. Perhaps it is more accurate to save the timestamp when the query is initially ran and save that as the last load time so we don't lose any tickets created while the pipeline is running. Checking every ticket for the timestamps and updating to the latest one seems unnecessarily complex.

pipelines/zendesk/helpers/api_helpers.py
redicane: updated is a datetime object while updated_at is a string, due to the formatting of the string this doesn't seem to make a difference anyway

pipelines/zendesk/helpers/api_helpers.py
redicane: This is done so there is a reference to the parent ticket in every custom_field. I guess a reference to the parent is always saved with dlt_load_id but for querying purposes it might be better if ticket id is also in the child table. If it's too much we can always remove this.

pipelines/zendesk/__init__.py
rudolfix: this is how it is done though. we want to automate it btw, https://github.com/dlt-hub/dlt/issues/194
you need to put the last timestamp from the ticket in there.
then you need to check if you are getting any duplicates. I suggested a test to do that

rudolfix (approved):
looks good! thanks for the contribution 🚀

PR #30, opened 2023-02-27
Sql database pipeline fixes

steinitzu:
Addressing some of what's mentioned in https://github.com/dlt-hub/pipelines/issues/25  

* Tests for incremental loading and selected tables
* Using `ConnectionStringCredentials`  
* Some restructuring of the code, organized cursor and table loading logic in a separate module

### DIFF
--- a/pipelines/sql_database/sql_database.py
+++ b/pipelines/sql_database/sql_database.py
@@ -21,6 +21,16 @@ def sql_table(
     unique_column: Optional[str] = dlt.config.value,
     write_disposition: TWriteDisposition = 'append'
 ) -> DltResource:
+    """A dlt resource which loads data from an SQL database table using SQLAlchemy.
+
+    :param credentials: Database credentials or an `sqlalchemy.Engine` instance
+    :param table: Name of the table to load
+    :param schema: Optional name of the schema table belongs to (uses databse default schema by default)
+    :param metadata: Optional `sqlalchemy.MetaData` instance. `schema` argument is ignored when this is used.
+    :param cursor_column: Optional column name to use as cursor for resumeable loading
+    :param unique_column: Optional column that uniquely identifies a row in the table for resumeable loading
+    :param write_disposition: Write disposition of the resource
+    """
     engine = engine_from_credentials(credentials)
     engine.execution_options(stream_results=True)
     metadata = metadata or MetaData(schema=schema)
@@ -44,8 +54,9 @@ def sql_database(
     """A dlt source which loads data from an SQL database using SQLAlchemy.
     Resources are automatically created for each table in the schema or from the given list of tables.
 
-    :param credentials: Credentials for the database to load
+    :param credentials: Database credentials or an `sqlalchemy.Engine` instance
     :param schema: Name of the database schema to load (if different from default)
+    :param metadata: Optional `sqlalchemy.MetaData` instance. `schema` argument is ignored when this is used.
     :param table_names: A list of table names to load. By default all tables in the schema are loaded.
 
     :return: A list of dlt resources for each table to be loaded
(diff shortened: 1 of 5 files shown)


tests/sql_database/test_sql_database_pipeline.py
```diff
@@ -26,14 +30,81 @@ def test_load_sql_schema_loads_all_tables(sql_source_db: SQLAlchemySourceDB, des
...
+                row = cur.fetchone()
+                assert row[0] == info['row_count']
+
+
+@pytest.mark.parametrize('destination_name', ALL_DESTINATIONS)
+def test_load_sql_table_incremental(sql_source_db: SQLAlchemySourceDB, destination_name: str) -> None:
+    """Run pipeline twice. Insert more rows after first run
+    and ensure only those rows are stored after the second run.
+    """
+    os.environ['SOURCES__SQL_DATABASE__CHAT_MESSAGE__CURSOR_COLUMN'] = 'updated_at'
```
rudolfix, line 77: it is really cool that it works! another thing you can do with dlt that never came to my mind :)

pipelines/sql_database/sql_database.py
```diff
@@ -1,121 +1,43 @@
...
-                if cursor is not None:
-                    _set_last_value(
-                        table.name, cursor_column, row._mapping[cursor], row._mapping.get(unique.name) if unique is not None else None
-                    )
-                yield dict(row._mapping)
+from pipelines.sql_database.util import table_rows
 
 
 @dlt.source
 def sql_database(
```
rudolfix, line 13: the only thing that I feel is missing is the standalone resource. it would call `table_rows` once and let the user to provide the write_disposition, cursor_column and unique_column directly. I gave you a few links on how to do that in previous review

This will be useful for people that want to load a defined set of tables, with different settings and want to declare everything in python, without going to `config.toml` for each table...

pipelines/sql_database/sql_database.py
```diff
@@ -1,121 +1,43 @@
...
-    database_url: str = dlt.secrets.value,
+    credentials: ConnectionStringCredentials = dlt.secrets.value,
     schema: Optional[str] = dlt.config.value,
     table_names: Optional[List[str]] = dlt.config.value,
     write_disposition: str = 'replace',
```
rudolfix, line 17: this is still not being used. so in fact all the resources are "append", (or maybe I'm wrong?)
btw. if this is set to replace and the cursor column is specified then the past data will be deleted and only the increment will be kept. there may be use cases for that...

pipelines/sql_database/util.py
```diff
@@ -0,0 +1,116 @@
...
+        self.table_has_cursor = self.cursor_column is not None
+        self.table_has_unique = self.unique_column is not None
+
+    def make_query(self) -> Select[Any]:
+        table = self.table
+        query = table.select()
+        if not self.cursor:
+            return query
+        cursor_col = self.cursor_column
+        if cursor_col is None:
```
rudolfix, line 63: I'd raise exception if cursor/primary key do not exist

rudolfix (changes requested):
This is really cool. I still asked for a few improvements. If you want we can merge this version and do the fixes in another PR - or we can continue here

pipelines/sql_database/sql_database.py
steinitzu: This argument was left over from when I was experimenting initially. I removed it for now. Doesn't really make sense to have it there since you can pass it to `pipeline.run` anyway, right?

pipelines/sql_database/util.py
steinitzu: Added this in `Cursor.__init__` so it's raised early as possible

steinitzu:
Hey @rudolfix , I added some more updates. Great to get this feedback.  

Added a standalone `sql_table` resource  
And also allowing sqla Engine to be passed instead of credentials. For the table resource that's espcially crucial so you can re-use the connection pool, and when you need something special for the engine setup.  

More tests are needed for the resource and different combinations of cursor/unique column. I'll have to refactor that a bit so there's not so much repetition.   
And will also add batching in `table_rows` result.  
I can do this in another PR.

PR #31, opened 2023-02-28
Hubspot

mebusila:
Hubspot api implementation as a source for dlt


pipelines/hubspot/__init__.py
```diff
@@ -0,0 +1,68 @@
+from typing import Sequence, Iterator
+
+import dlt
+
+from dlt.common.typing import TDataItems
+from dlt.extract.source import DltResource
+
+from pipelines.hubspot.client import fetch_data
```
rudolfix, line 8: please use relative imports ie.
`from .hubspot.client import fetch_data`

we'll need it to properly distribute code.
https://github.com/dlt-hub/pipelines#python-module-import-structure

pipelines/hubspot/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        is passed to `fetch_data` as the `api_key` argument.
+
+        This function is decorated with `dlt.source`, which indicates that it is a Dlt source function.
+        When this function is included in a dlt pipeline, it will be executed to retrieve data for the
+        specified HubSpot API endpoints.
+
+    """
+
+    @dlt.resource(write_disposition="replace")
+    def companies() -> Iterator[TDataItems]:
```
rudolfix, line 41: you could generate resources dynamically. see the `pipedrive` pipeline. OFC the list here is also good - it is a matter of taste (to some degree)

pipelines/hubspot/client.py
```diff
@@ -0,0 +1,122 @@
...
+    if 'paging' in _data:
+        _next = _data['paging'].get('next', None)
+        if _next:
+            # Replace the base URL with an empty string to get the relative URL for the next page
+            next_url = _next['link'].replace(BASE_URL, '')
+            # Recursively call the `fetch_data` function to get the next page of results
+            yield from fetch_data(next_url, **kwargs)
+
+
+@retry(tries=3, delay=1, backoff=1.1)
```
rudolfix, line 82: good! we also work on https://github.com/dlt-hub/dlt/issues/147

the `retry` decorator is very crude (will retry HTTP 401 as well)

tests/hubspot/test_hubspot_pipeline.py
```diff
@@ -0,0 +1,86 @@
...
+
+import pytest
+import requests
+
+from pipelines.hubspot import fetch_data, CRM_CONTACTS_ENDPOINT, CRM_COMPANIES_ENDPOINT, CRM_DEALS_ENDPOINT, CRM_PRODUCTS_ENDPOINT, CRM_TICKETS_ENDPOINT, CRM_QUOTES_ENDPOINT
+from tests.hubspot.mock_data import mock_contacts_data, mock_companies_data, mock_deals_data, mock_products_data, mock_tickets_data, mock_quotes_data
+
+
+@pytest.fixture()
+def mock_response():
```
rudolfix, line 11: thanks for writing the tests with mocked data! besides this test could you add end to end test like the one here:
https://github.com/dlt-hub/pipelines/blob/master/tests/pipedrive/test_pipedrive_pipeline.py#L11
it is just a few lines

hubspot gives you dev accounts with test data. just ping me for a key. I'll add it to our CI as well

rudolfix (changes requested):
this looks really good! thx! please fix the import and consider end to end test as per comments. thanks!

pipelines/hubspot/__init__.py
mebusila: Yes, that is true, but i think in this way is more visible that resources are included, i will suggest keep it like this, at least for now when the number of resources is till manageable

pipelines/hubspot/client.py
mebusila: that;s good news, i like wrapping things around, you can get more control than just using a library

PR #32, opened 2023-03-03
Sql pipeline tests and batch loading

steinitzu:
https://github.com/dlt-hub/pipelines/issues/25
Test table resource incremental loading and column not found errors  
Yield rows in batches


rudolfix (approved):
looks good!

PR #43, opened 2023-03-18
Adds shopify pipeline

### LINKED ISSUE #42: shopify pipeline
# Planned Pipeline Info
Please provide info below when opening the issue to give use heads-up on what is planned.
- **name** of the pipeline to be used by `dlt init` and to be placed in `pipelines` folder: [e.g. pipedrive]
- **category and description** of the pipeline: [e.g. CRM, loads the relevant data from pipedrive api]

You can fill the data below when writing spec.

# Use Cases
Please provide descriptions up to 3 most important use cases that users of this pipeline do. Those use cases will be:
- implemented
- reviewed
- demonstrated in pipeline script
- documented
use case description is not only targeted at the developers but also by people creating test account and doing demos.

# Sources / Resources / Endpoints
Define pipeline interface to the user in terms of sources and resources.
- enumerate all the sources with information form which endpoints the data comes from
- ideally provide the config arguments to each source (ie. start dates, report ranges etc.)
- you can use pseudocode to show how you intend to use the source
- provide the default write disposition for resources in the source (all append, all replace?)
- in the sources identify the incremental and merge resources and specify them in a reasonable way (ideally by giving the cursor columns - what is the last value really?, primary keys and merge keys).

# Customization
Enumerate everything that goes beyond standard `dlt` building blocks. Suggest the implementation
- use of state
- in the code or as additional transform, filter or map function
- ask dlt team for help if it looks like complex software task

# Test account / test data
- specify what data you expect in the test dataset. otherwise refer to use cases
- specify what kind of account access you need, include tool's name, required plan or features needed

# Implementation tasks
Below you have a proposal for implementation tasks
* [ ] implement all the described sources, resources and endpoints
* [ ] make sure that the Use Cases can be easily executed by the pipeline's user by **providing demonstrations of all the use cases** in the `[pipeline_name]_pipeline.py`
* [ ] all the use cases are tested
* [ ] if there are any customizations you must test them, including unit tests for custom code

nacho0605:
Closes #42
Tests are pending

### DIFF
--- a/pipelines/shopify/__init__.py
+++ b/pipelines/shopify/__init__.py
@@ -6,10 +6,10 @@ To get an admin access token: https://shopify.dev/docs/apps/auth/admin-app-acces
 """
 
 import dlt
-import requests
 
 from dlt.common.typing import TDataItems
 from dlt.extract.source import DltResource
+from dlt.sources.helpers import requests
 from time import sleep
 from timeit import default_timer
 from typing import Any, Dict, Iterator, Optional, Sequence
--- a/pipelines/shopify/__init__.py
+++ b/pipelines/shopify/__init__.py
@@ -10,13 +10,15 @@ import requests
 
 from dlt.common.typing import TDataItems
 from dlt.extract.source import DltResource
+from time import sleep
+from timeit import default_timer
 from typing import Any, Dict, Iterator, Optional, Sequence
 
 from .incremental_loading_helpers import get_linked_url, RelationType, get_since_timestamp, set_last_timestamp
 
 
 @dlt.source(name='shopify')
-def shopify_source(access_token: str = dlt.secrets.value, store_name: str = dlt.secrets.value, api_version: str = dlt.config.value) -> Sequence[DltResource]:
+def shopify_source(access_token: str = dlt.secrets.value, store_name: str = dlt.secrets.value, api_version: str = dlt.config.value, rate_limit: int = dlt.config.value) -> Sequence[DltResource]:
     """
 
     Args:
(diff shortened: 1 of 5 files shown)


pipelines/shopify/incremental_loading_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+import dlt
+import re
+
+
+class RelationType(str, Enum):
+    NEXT = 'next'
+    PREVIOUS = 'previous'
+
+
+def get_linked_url(link_header: str, relation_type: str) -> str:
```
rudolfix, line 17: `requests` do that automatically. see `response.links` and https://github.com/dlt-hub/pipelines/blob/master/pipelines/github/__init__.py#L239

pipelines/shopify/incremental_loading_helpers.py
```diff
@@ -0,0 +1,97 @@
+from datetime import datetime, timedelta, timezone
+from dateutil import parser
```
rudolfix, line 2: please import `pendulum` from `from dlt.common import pendulum` and use it
str(pendulum.now()) - to iso format
pendulum.parse(str(pendulum.now())) from iso format. please remove your own parsing

pipelines/shopify/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+        # yield data only
+        data = page[endpoint]
+        if data:
+            last_timestamp_str = data[-1].get('created_at', '')
+            yield data
+        # check if next page exists
+        link_header = response.headers.get('Link')
+        linked_url = get_linked_url(link_header, RelationType.NEXT) if link_header else ''
+        # is_next_page is set to True or False
+        is_next_page = linked_url != ''
```
rudolfix, line 93: just break here.
```
if not linked_url:
   break
```

pipelines/shopify/incremental_loading_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    return linked_url
+
+
+@with_config
+def get_since_timestamp(endpoint: str, step: int = dlt.config.value, max_retries: int = dlt.config.value, backoff_delay: float = dlt.config.value, days_back: int = dlt.config.value) -> str:
+    """
+    Specific function to generate 'since_timestamp' based on last timestamp stored in dlt's state/destiny (if available)
+    """
+    since_timestamp_str = ''
+    if all([isinstance(step, int), step > 0, isinstance(days_back, int), days_back > 0]):
```
rudolfix, line 38: without explanation I'm not able to review it. what are those `days_back`

pipelines/shopify/incremental_loading_helpers.py
```diff
@@ -0,0 +1,97 @@
...
+        last_timestamp = _get_last_timestamp_from_state(endpoint)
+        if not last_timestamp:
+            last_timestamp = _get_last_timestamp_from_destiny(endpoint, max_retries, backoff_delay)
```
rudolfix, line 41: why do you need to access destination?

rudolfix (changes requested):

pipelines/shopify/incremental_loading_helpers.py
nacho0605: Done.
Thanks!

pipelines/shopify/incremental_loading_helpers.py
nacho0605: It tries to look up for endpoint's last timestamp on dlt's state, then on dlt's destiny, and finally loads a default value (set up to 30 days back by default) if necessary.
'step' value (1 second by default) is added to 'last timestamp' in order to avoid requesting/storing duplicated data (last/first data item).
'max_retires' and 'backoff_delay' are associated to destiny's requests.

pipelines/shopify/incremental_loading_helpers.py
nacho0605: Just in case dlt's state gets erased from disk or unavailable for any another reason, and destiny still remains loaded with data.
@adrianbr

adrianbr: it should not be necessary. I think this code will not run. I added some docs to state, maybe they help? https://dlthub.com/docs/concepts/state

nacho0605: I've removed that block and updated the rest.

pipelines/shopify/incremental_loading_helpers.py
nacho0605: I've removed dlt's destiny looking up part.

rudolfix (changes requested):

pipelines/shopify/incremental_loading_helpers.py
nacho0605: Done.

pipelines/shopify/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+    last_timestamp_str = ''
+    while is_next_page:
+        start_time = default_timer()
+        response = requests.get(url, headers=headers, params=params)
+        response.raise_for_status()
+        page = response.json()
+        # yield data only
+        data = page[endpoint]
+        if data:
+            last_timestamp_str = data[-1].get('created_at', '')
```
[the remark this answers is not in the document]
nacho0605: Done.

PR #59, opened 2023-03-31
adjust pipeline default behavior

adrianbr:


pipelines/zendesk_pipeline.py
```diff
@@ -54,7 +54,7 @@ def incremental_pipeline():
 if __name__ == "__main__":
     # simple run where everything is loaded
     start = time.time()
-    load_info = load_all()
+    load_info = incremental_pipeline()
```
redicane, line 57: I actually think this should be load_all() instead of incremental_pipeline(). load_all() will take all the data in the pipeline since the last load date in dlt state (if empty it will take all data this millennium) while incremental_pipeline() will take all the data since the chosen start time which by default will be 1st Jan 2023 everytime. That was meant as an example that this behaviour could also be enabled. I wrote that in the comments but maybe it should also be made clearer in the function names.

Issue #72, opened 2023-04-11
Zendesk Support not working with Nested Objects in response


redicane:
Zendesk Support had an issue where a response from Zenpy would return an api object inside an api object which caused the pipeline to crash since generated dicts were not json serialisable.

PR #74, opened 2023-04-11
Clean- Fix issue with Zendesk Support

redicane:
Fixes issue #72

### DIFF
--- a/pipelines/zendesk/helpers/api_helpers.py
+++ b/pipelines/zendesk/helpers/api_helpers.py
@@ -1,10 +1,11 @@
+import json
 from typing import Iterator, Any, Optional, Union
 from dlt.common import logger, pendulum
 from dlt.common.exceptions import MissingDependencyException
 from dlt.common.typing import DictStrAny, DictStrStr, TDataItem, TDataItems
 try:
     from zenpy import Zenpy
-    from zenpy.lib.api_objects import Ticket
+    from zenpy.lib.api_objects import Ticket, JobStatus
 except ImportError:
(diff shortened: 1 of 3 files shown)


pipelines/zendesk/helpers/api_helpers.py
```diff
@@ -92,14 +91,14 @@ def basic_load(resource_api: Iterator[Any]) -> Iterator[TDataItem]:
     """
     # sometimes there is a single element which is a dict instead of a generator of objects
     if isinstance(resource_api, dict):
-        yield resource_api
+        yield _make_json_serializable(resource_api)
```
rudolfix, line 94: every Zenpy object derives from `BaseObject` that implements this:
```
    def to_json(self, indent=2):
        """ Return self formatted as JSON. """
        return json.dumps(self, default=json_encode_for_printing, indent=indent)
        ```
why not to use it? `Ticket` does that as well

rudolfix (changes requested):

pipelines/zendesk/helpers/api_helpers.py
redicane: Hey @rudolfix,

From what I can see from experimenting with it, the to_json method will convert the whole object to a string which would then need to be converted to a dict with the correct data types for all fields. This is a bit redundant because the process is done by the to_dict method. In my opinion, using to_json and then encoding ourselves does more harm then good.

redicane: I had a look at that implementation for to_dict. I think using to_dict with the default options should be ok for our use case.

PR #80, opened 2023-04-14
uses incremental in github

rudolfix:


steinitzu (commented):
This looks great!

One question: The state isn't scoped per repo (wasn't before either) so you can only load 1 repo per schema, right? Is that intentional?

steinitzu (approved):

steinitzu:
Yes makes sense you'd rename the source if needed. Very cool with the dynamic tables!

PR #83, opened 2023-04-14
google analytics 4 pipeline

rudolfix:
we continue #54 here which got merged by accident and had to be reverted


adrianbr:
end date: always yesterday, do not load "partial, current period"

Run 1. 
If start date exists: Load from start date until end date
Else: Load from start of all time (2000 or similar)  until end date

Run 2.
Load from last_loaded at until end date

rudolfix:
@adrianbr cool we are getting there! What happens if user specifies a new date in run 2? Should we raise error? Should we ignore it (maybe with a warning)? Shpuld we do a full refresh?

adrianbr:
> @adrianbr cool we are getting there! What happens if user specifies a new date in run 2?
I would ignore it. If the user really wants to start loading from earlier, they can just start with a full load.

adrianbr:
After talking with Redi, it seems the most straightforward way to handle it is to only use the start date the first time. If the user resets this to something else, we should warn and carry on

redicane (approved):

redicane (approved):

redicane (dismissed):

Issue #77, opened 2023-04-14
Matomo basic pipeline


adrianbr:
# Description
Matomo basic pipeline

# Use Cases
- Request tracking event data that was tracked with Matomo. I should have this data in a table that can be incremented on any schedule so I can keep it near live. I should be able to optionally also get the stateful user and update (merge) this data with events.
- Request aggregated report data on data tracked with Matomo. I expect this to always be aggregated on date, and incremented on date, so I can see the data up to end of yesterday, today. it is loaded incrementally in a nightly load.
- Request user defined report. I expect that it has a date column on which it is loaded incrementally in a nightly load.

# Sources / Resources / Endpoints
We want TWO sources with the folowing resources:

We want the sources to be split by schedule:
matomo_reports: Contains Reporting and Custom reporting
matomo_events: Contains events and dependent users

### Reporting resource
- we want the reporting api, the part where we can send a query and get back some data like in ga4
https://developer.matomo.org/api-reference/reporting-api
Requirements:
- Loading mode -  **let's always request date with our queries.** We want to load incrementally based on this date by using "append" mode based on last value. We should only load complete periods such as up till yesterday
- User should have some example queries he can issue
- User should be able to change the query
- we should increment on date by using the append/last value based increment described here https://dlthub.com/docs/general-usage/incremental-loading#incremental-loading-with-last-value

### Raw data "events"
- Raw data: https://developer.matomo.org/api-reference/reporting-api#Live we want this "raw data" api too - this should allow to export raw data from matomo
Resource:
-  Live.getLastVisitsDetails: Named "events" - This should be incremented as above. Always start from all time (which is a last 10d rss) and increment from last load.
-  Live.getVisitorProfile. named "visitors" This should be a dependent resource on events and request the profile for those users that showed up in the event increment. The visitors profile should be merged with each increment.

### Custom reports
We want to get custom reports too, as a method alternative to the first reports resource.
- CustomReports.getCustomReport (idSite, period, date, idCustomReport, segment = '', expanded = '', flat = '', idSubtable = '', columns = '') [ No example available ]
- I am not sure if we can get the report name automatically for the table name - if not provide the user with an option
- we can always request and increment on date so please use the same increment method as above.

# Customization
I believe there should be no customisation used for this pipeline. 
If it looks like we should use something, please notify the dlt team.

# Test account / test data
Use the public demo
[...]

PR #100, opened 2023-04-26
Draft: Add matomo

redicane:
Adding basis for Matomo Pipeline.

# Relevant issue

issue #77


pipelines/matomo/helpers/data_processing.py
```diff
@@ -0,0 +1,84 @@
...
+FIRST_DAY_OF_MILLENNIUM_TIMESTAMP = 946684800.0
+
+
+def process_report(report: Iterator[TDataItem]) -> Iterator[TDataItem]:
+    """
+    Helper, loops through multiple formats of method_data and processes them into dlt resources
+    :param report: Response from Matomo API containing data for a single method.
+    :returns: generator of dicts.
+    """
+    if isinstance(report, dict):
```
rudolfix, line 19: decomposing reports into single lines yielded separately will slow your pipeline substantially. the best way is to add elemenets to a list and then return a list (not yield). there's really no need to yield here. I know it looks cool but it is rather impractical

pipelines/matomo/helpers/data_processing.py
```diff
@@ -0,0 +1,84 @@
...
+                    el["date"] = pendulum.parse(key)
+                    yield el
+            elif isinstance(value, dict):
+                value["date"] = pendulum.parse(key)
+                yield value
+            else:
+                yield report
+    else:
+        try:
+            for value in report:
```
rudolfix, line 32: here just add dates and return `report`

pipelines/matomo_pipeline.py
```diff
@@ -0,0 +1,71 @@
...
+    print(info)
+
+
+def run_reports():
+    """
+    Runs the pipeline only loading reports.
+    :return:
+    """
+    pipeline_events = dlt.pipeline(dataset_name="matomo_reports", full_refresh=False, destination="postgres", pipeline_name="matomo")
+    data = matomo_reports()
```
rudolfix, line 50: i'd expect most of the people will pass site id explicitly. maybe you could do that here, not relying on configuration

rudolfix (changes requested):
this looks good! I have a few usability and performance issues. thanks for the tests!

pipelines/matomo/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+import dlt
+import pendulum
+from dlt.common.typing import DictStrAny, TDataItem
+from dlt.extract.source import DltResource
+from .helpers.matomo_client import MatomoAPIClient
+from .helpers.data_processing import FIRST_DAY_OF_MILLENNIUM, get_matomo_last_load_timestamp, get_matomo_date_range, process_report, process_visitors
+
+
+@dlt.source(max_table_nesting=2)
+def matomo_reports(credentials: Dict[str, str] = dlt.secrets.value, queries: List[DictStrAny] = dlt.config.value) -> List[DltResource]:
```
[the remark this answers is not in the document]
redicane: I agree. I was also thinking of moving url to config since it is not actually a secret value.

rudolfix:
@redicane one question about paging (forgot to ask). what do you mean by inconsistent? there's a guide how to use paging for raw events here https://matomo.org/faq/how-to/faq_24536/

is this information incorrect/outdated?

redicane:
Hey @rudolfix, filter_limit and filter_offset were flat out not working for the methods we tried. Additionally, there is no info in the response indicating that there is another page of data. That being said, I can experiment a bit more with live visits data.

pipelines/matomo_pipeline.py
redicane: @rudolfix, For reports site_id is inside the queries. I guess you also want to move site_id out of the queries and have the same site_id for every query?

pipelines/matomo/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+    :param visits: List of dicts containing information on last visits in the given timeframe
+    :param client: Used to make calls to Matomo API
+    :param site_id: Every site in Matomo has a unique id
+    :returns: Dict containing information about  the visitor
+    """
+
+    for visit in visits:
+        visitor_id = visit["visitorId"]
+        method_data = client.get_method(site_id=site_id, method="Live.getVisitorProfile", extra_params={"visitorId": visitor_id})
+        yield method_data
```
rudolfix, line 116: `yield from` because `get_method` yields. otherwise you yield a generator. which `dlt` will deal with by creating additional step in pipeline that we do not need

pipelines/matomo/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Dlt transformer. Receives information about visits from get_last_visits
+    :param visits: List of dicts containing information on last visits in the given timeframe
+    :param client: Used to make calls to Matomo API
+    :param site_id: Every site in Matomo has a unique id
+    :returns: Dict containing information about  the visitor
+    """
+
+    for visit in visits:
+        visitor_id = visit["visitorId"]
```
rudolfix, line 114: you said there's way to batch it?

tests/matomo/test_matomo_pipeline.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@pytest.mark.parametrize("destination_name", ALL_DESTINATIONS)
+def test_incrementing_reports(destination_name: str) -> None:
+    """
+    Checks that incremental loading works for reports
+    @:param: destination_name - redshift/bigquery/postgres
+    """
+    # Define all end_dates to be used for incremental loading
+    first_end_date = pendulum.datetime(year=2020, month=1, day=1)
```
rudolfix, line 88: where those dates are coming from? they should split the existing dataset in several parts

rudolfix (changes requested):
looks good, one issue to fix + 2 questions. also we do not have access to your test account. we'll ping you regarding that

tests/matomo/test_matomo_pipeline.py
redicane: Those are some random dates for the test data. With the demo matomo site every split has some data.

pipelines/matomo/__init__.py
redicane: @rudolfix  I was talking about getting last visits not last visitors. I pushed some changes that try to batch this but are limited by the api calls and still not really optimised really.

rudolfix (approved):

Issue #96, opened 2023-04-24
Stripe verified pipeline


adrianbr:
# Description

Stripe is an online payment company that offers software and APIs for processing payments and business management. 

This pipeline uses Stripe API and dlt to load data such as Customer, Subscription, Event and etc. to the database and to calculate the MRR and churn rate.

- **name** of the pipeline is "stripe_analytics"
- **category and description** of the pipeline: loads the relevant data from Stripe API and calculates some metrics. 

# Use Cases

1. Pipeline provides automatic data collection using Stripe API and storing them in a database (DuckDB, Postgres, Redshift, BigQuery). You will be able to store information about customers, all their purchases, subscriptions, as well as information about your products and etc.

> You can load data from any endpoint `load_data(endpoints=("Customer", "Product"))`.

> You can load data in Incremental mode `load_incremental_endpoints(endpoints=("Event", "Invoice"))`.

2. With the pipeline you can calculate the most important metrics (MRR and subscription churn rate) and store them in a database as a _resource_. Use `load_data_and_get_metrics()`.

# Sources / Resources / Endpoints

We have two sources: stripe_source and incremental_stripe_source. 

1. **stripe_source**:
Retrieves data from the Stripe API for the specified endpoints. For almost all endpoints, Stripe API responses do not provide key "updated", so in most cases we are forced to load the data in 'replace' mode. This source is suitable for all types of endpoints, including 'Events', 'Invoice' etc, but these endpoints can also be loaded in incremental mode (see source _incremental_stripe_source_).

2. **incremental_stripe_source**:
As Stripe API does not include the "updated" key in its responses, we are only able to perform incremental downloads from endpoints where all objects are uneditable. This source yields the resources with incremental loading based on "append" mode. You will load only the newest data without duplicating and without downloading a huge amount of data each time.

To calculate the metrics in `load_data_and_get_metrics()`, we need endpoints Subscription and Event. All endpoints become resources automatically.

Resource Subscription:

- With [Subscriptions](https://stripe.com/docs/billing/subscriptions/creating), customers make recurring payments for access to a product. Subscriptions require you to retain more information about your customers than one-time purchases do because you need to charge customers in the future.
- _Updated Endpoint_. Object states can change over time, for example, customers can unsubscribe, so that the status of the subscription changes from active to canceled.
[...]

PR #110, opened 2023-05-03
Features/stripe pipeline

AstrakhantsevaAA:
# Tell us what you do here

- contributing new pipeline for Stripe
- provide mrr and churn metrics calculation as a resource
- tests

# Relevant issue

issue #96

# More PR info


pipelines/stripe_analytics/stripe_analytics.py
```diff
@@ -0,0 +1,107 @@
...
+        created={"gte": start_date, "lt": end_date}, **kwargs
+    )
+    return dict(resource_dict)
+
+
+@dlt.source
+def stripe_source(
+    stripe_secret_key: str = dlt.secrets.value,
+    limit: int = 100,
+    get_all_data: bool = False,
```
rudolfix, line 48: what is the use of `get_all_data`? is it useful to our users? on first look it stops loading after receiving first page of results. this you can do directly in `dlt` by doing `stripe_source().add_limit(1)`. https://dlthub.com/docs/general-usage/source#load-data-partially
IMO we do not need this option

pipelines/stripe_analytics/stripe_analytics.py
```diff
@@ -0,0 +1,107 @@
...
+    start_date: Optional[datetime] = None,
+    end_date: Optional[datetime] = None,
+) -> Generator[DltResource, Any, None]:
+    stripe.api_key = stripe_secret_key
+    stripe.api_version = "2022-11-15"
+
+    def get_resource(
+        endpoint: Endpoints,
+        created: Optional[Any] = dlt.sources.incremental("created", initial_value=-3600),
```
rudolfix, line 57: IMO you should use `start_date` as initial value of incremental. then you will be consistent with other pipelines. we had a long discussion with @adrianbr on how we should deal with initial values in incremental load. and this is what we decided:
1. if the start_date is passed it is used as initial value for a first load (full load)
2. on subsequent loads we continue from the `last_value` and ignore start date
3. people can reset the resource/source by doing `write_dispositon="replace"` as described here https://dlthub.com/docs/general-usage/source#do-a-full-refresh

pipelines/stripe_analytics/stripe_analytics.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> Generator[Dict[Any, Any], Any, None]:
+        get_more = True
+        starting_after = None
+        start_value = created.last_value
+
+        while get_more:
+            response = stripe_get_data(
+                endpoint,
+                start_date=start_value if start_date is None else start_date,
```
rudolfix, line 66: you pass start_value here. or just `created.start_value` (`dlt` keeps this value for you). the construct here is overriding incremental load when start_date is specified

pipelines/stripe_analytics/stripe_analytics.py
```diff
@@ -0,0 +1,107 @@
...
+            if len(response["data"]) > 0:
+                starting_after = response["data"][-1]["id"]
+
+            yield response["data"]
+
+    for endpoint in Endpoints:
+        yield dlt.resource(
+            get_resource,
+            name=endpoint.value,
+            write_disposition="merge",
```
rudolfix, line 82: do we need merge here? this endpoint is returning updates to objects right? do same "id" can be retruned many times?

rudolfix (changes requested):
this looks really good! thanks for the tests.
please see the review. some questions and suggestions

pipelines/stripe_analytics/metrics.py
```diff
@@ -0,0 +1,72 @@
+from datetime import datetime, timedelta
```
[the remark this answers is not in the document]
AstrakhantsevaAA: I didn't want to pull one more dependency to this script. okay I'll change it
should I use pendulum for typing?

pipelines/stripe_analytics/stripe_analytics.py
AstrakhantsevaAA: You suggest to load all data by default?
sure, I can use this `stripe_source().add_limit(1)` feature additionally

pipelines/stripe_analytics/stripe_analytics.py
AstrakhantsevaAA: probably you are right. i'm just afraid of "append" mode XD

AstrakhantsevaAA: good question. I just realized that we need to reload almost all data, because, for example, endpoint Subscriptions doesn't create new object, just update old. we need to update our data every time

pipelines/stripe_analytics/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@dlt.source
+def incremental_stripe_source(
+    endpoints: list,
+    stripe_secret_key: str = dlt.secrets.value,
+    start_date: Optional[DateTime] = None,
+    end_date: Optional[DateTime] = None,
+) -> Generator[DltResource, Any, None]:
+    stripe.api_key = stripe_secret_key
```
rudolfix, line 22: we need docstrings! your explanations from demo script should be also included here. and:
- what are valud endpoint names? point to the stripe docs? otherwise how people can guess what goes here?
- the start data is rather `initial_start_date` - it is only applied on first run. you must explain that
- what is the role of end date? you actually do not demonstrate usage 

no one is going to guess that you filter dates on created at. you may even put it in the name `initial_created_at`

pipelines/stripe_analytics/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+            incremental_resource,
+            name=endpoint,
+            write_disposition="append",
+            primary_key="id",
+        )(endpoint)
+
+
+@dlt.source
+def updated_stripe_source(
+    endpoints: list,
```
rudolfix, line 47: same

pipelines/stripe_analytics/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+    stripe_secret_key: str = dlt.secrets.value,
+    start_date: Optional[DateTime] = None,
+    end_date: Optional[DateTime] = None,
+) -> Generator[DltResource, Any, None]:
+    stripe.api_key = stripe_secret_key
+    stripe.api_version = "2022-11-15"
+
```
rudolfix, line 54: docstrings: same

pipelines/stripe_analytics/helpers.py
```diff
@@ -0,0 +1,63 @@
+from enum import Enum
+from typing import Any, Dict, Generator, Optional, Union
+
+import stripe
+from dlt.common import pendulum
+from pendulum import DateTime
+
+
+class UpdatedEndpoints(Enum):
```
rudolfix, line 9: better to use strings and tuples ie DEFAULT_UPDATED_ENDPOINTS. see facebook ads

rudolfix (changes requested):

rudolfix (approved):
I'd say this is now top! more like SDK for building stripe pipelines but my take is that we should go even more in this direction

rudolfix (approved):

PR #114, opened 2023-05-04
Shopify pipeline

sh-rp:
# Tell us what you do here

# More PR info
We could not use the Shopify python sdk but use our own requests implementation, using the sdk produces nicer code, using our requests probably guards against a couple of problems.

### DIFF
--- a/pipelines/shopify_dlt/__init__.py
+++ b/pipelines/shopify_dlt/__init__.py
@@ -1,0 +1,76 @@
+
+"""Fetches Shopify Orders and Products."""
+
+from typing import Any, Dict, Iterator, Iterator, Optional
+
+import shopify
+
+import dlt
+
+from dlt.extract.source import DltResource
+
+FIRST_DAY_OF_MILLENNIUM = "2000-01-01"
+DEFAULT_API_VERSION = "2023-01"
+
+@dlt.source(name="shopify")
+def shopify_source(
+    private_app_password: str = dlt.secrets.value,
+    api_version: str = DEFAULT_API_VERSION,
+    shop_url: str = dlt.config.value,
+    start_date: Optional[str] = FIRST_DAY_OF_MILLENNIUM,
+) -> Iterator[DltResource]:
+    """
+    The source for the shopify pipeline. Available resources are products, orders and customers.
+    @:param: private_app_password - The app password to the app on your shop
+    @:param: api_version - The api version to use
+    @:param: shop_url - The url of your shop
+    @:param: start_date - The date from which to import items, will import items initially created on or after this date
+    @:return: data_resources_list - list of dlt resources
+    """
+
+    # build client
+    session = shopify.Session(shop_url, api_version, private_app_password)
+    shopify.ShopifyResource.activate_session(session)
+
+    # define resources
+    @dlt.resource(primary_key="id", write_disposition="merge")
+    def products(updated_at = dlt.sources.incremental("updated_at", initial_value=FIRST_DAY_OF_MILLENNIUM)) -> Iterator[Dict[str, Any]]:
+        """
+        The resource for products on your shop, supports incremental loading and pagination
+        @:param: updated_at - The saved state of the last updated_at value
+        @:return: generator of products
+        """
+        page = shopify.Product.find(updated_at_min=updated_at.last_value, created_at_min=start_date)
+        while page:
+            for product in page:
+                yield product.to_dict()
+            page = page.next_page() if page.has_next_page() else None
+
+    @dlt.resource(primary_key="id", write_disposition="merge")
+    def orders(updated_at = dlt.sources.incremental("updated_at", initial_value=FIRST_DAY_OF_MILLENNIUM)) -> Iterator[Dict[str, Any]]:
+        """
+        The resource for orders on your shop, supports incremental loading and pagination
+        @:param: updated_at - The saved state of the last updated_at value
+        @:return: generator of orders
+        """
+        page = shopify.Order.find(updated_at_min=updated_at.last_value, created_at_min=start_date)
+        while page:
+            for order in page:
+                yield order.to_dict()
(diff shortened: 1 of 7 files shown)


pipelines/shopify_dlt/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+import dlt
+
+from dlt.extract.source import DltResource
+
+
+@dlt.source(name="shopify")
+def shopify_source(
+    private_app_password: str = dlt.secrets.value,
+    api_version: str = dlt.config.value,
```
rudolfix, line 16: please use a default value for api version. default values can still be configured in `config.toml` but optionally

pipelines/shopify_dlt/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    @:return: data_resources_list - list of dlt resources
+    """
+
+    # build client
+    session = shopify.Session(shop_url, api_version, private_app_password)
+    shopify.ShopifyResource.activate_session(session)
+
+    # define resources
+    @dlt.resource(primary_key="id", write_disposition="merge")
+    def products(updated_at = dlt.sources.incremental("updated_at", initial_value="1970-05-04")) -> Iterator[Dict[str, Any]]:
```
rudolfix, line 33: `updated_at` is a ISO datetime string, right? then  good.

pipelines/shopify_dlt/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+import dlt
+
+from dlt.extract.source import DltResource
+
+
+@dlt.source(name="shopify")
+def shopify_source(
+    private_app_password: str = dlt.secrets.value,
+    api_version: str = dlt.config.value,
+    shop_url: str = dlt.config.value,
```
rudolfix, line 17: please add one more parameter `start_date` or `start_timestamp` that you can then pass as `initial_value` to `incremental`. we'll have people with millions of orders spanning many years. and maybe they just need to load last year of data

pipelines/shopify_dlt_pipeline.py
```diff
@@ -0,0 +1,21 @@
+"""Pipeline to load shopify data into BigQuery.
+"""
+
+import dlt
+from typing import List
+from shopify_dlt import shopify_source
+
+def load(resources: List[str]) -> None:
```
rudolfix, line 8: if you add start_date let's demonstrate it

tests/shopify_dlt/test_shopify_pipeline.py
```diff
@@ -0,0 +1,20 @@
...
+    assert_load_info(load_info)
+    table_names = [t["name"] for t in pipeline.default_schema.data_tables()]
+    table_counts = load_table_counts(pipeline, *table_names)
+
+    # for now only check main tables
+    expected_tables = ['products', 'orders', 'customers']
+    assert set(table_counts.keys()) > set(expected_tables)
+    assert table_counts['products'] == 17
+    assert table_counts['orders'] == 10
+    assert table_counts['customers'] == 3
```
rudolfix, line 20: load again to make sure incremental loading works (counts should not change)

pyproject.toml
```diff
@@ -48,3 +48,6 @@ asana = "^3.2.1"
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.poetry.group.shopify_dlt.dependencies]
+shopifyapi = "^12.3.0"
```
rudolfix, line 53: 👍

rudolfix (changes requested):

pipelines/shopify_dlt/__init__.py
sh-rp: Yes

pipelines/shopify_dlt/__init__.py
sh-rp: done

sh-rp: (not quite sure though if I did it right. intuitively the way I did it now it would not get the version from the config file, I did not find anything in the docs about default values of config items though)

pipelines/shopify_dlt_pipeline.py
sh-rp: done

tests/shopify_dlt/test_shopify_pipeline.py
sh-rp: done (I think)

pipelines/shopify_dlt/__init__.py
sh-rp: I took a slightly different approach, but one that makes more sense imho. lmk what you think.

rudolfix: this means that we do not load entities created after `start_date` even if they are updated after `start_date`? it is hard to say if this makes more sense or not... I personally would just filter by `updated_at` but I'm ok with what you did if you explain this in the docstrings below

pipelines/shopify_dlt_pipeline.py
rudolfix: where? :)

rudolfix: maybe the change is not comitted?

pipelines/strapi/readme.md
```diff
@@ -1,18 +1,78 @@
-## Strapi
+---
```
rudolfix, line 1: why strapi readme is here?

pipelines/shopify_dlt_pipeline.py
rudolfix: ahh ok. it is in the tests. but this script is distributed with the pipeline and used for copy and paste. so we should show how start date is used as well.

rudolfix (changes requested):
all (almost) good. thx for test update!

rudolfix (approved):

Issue #87, opened 2023-04-19
Google Ads basic pipeline


matthauskrzykowski:
# Planned Pipeline Info
Please provide info below when opening the issue to give use heads-up on what is planned.
- **name** google_ads
- **category and description** of the pipeline: Ad network

# Use Cases
Use case 1: Custom query source
1. User can pass one or multiple queries + associated table names to a function to create dynamically multiple resources.  It should be demonstrated for the user how to pass more than 1 query.
- User already has 2 examples: 1. Keyword report as in the google docs https://developers.google.com/google-ads/api/samples/get-keyword-stats, and anything else.

Use case 2: Pre-selected tables source
User can select to consume this source and they will get a preconfigured set of resources/queries/reports.

dimension tables: 
These seem to be tables with descriptive data, that should be considered stateful dimensions We should load them with replace load disposition.
[accounts](https://developers.google.com/google-ads/api/fields/v11/customer)
[ad_group_ads](https://developers.google.com/google-ads/api/fields/v11/ad_group_ad)
[ad_group_ad_labels](https://developers.google.com/google-ads/api/fields/v11/ad_group_ad_label)
[ad_groups](https://developers.google.com/google-ads/api/fields/v11/ad_group)
[ad_group_labels](https://developers.google.com/google-ads/api/fields/v11/ad_group_label)
[campaign_labels](https://developers.google.com/google-ads/api/fields/v11/campaign_label)
[click_view](https://developers.google.com/google-ads/api/reference/rpc/v11/ClickView)
[keyword](https://developers.google.com/google-ads/api/fields/v11/keyword_view)
[geographic](https://developers.google.com/google-ads/api/fields/v11/geographic_view)

Report Tables: These always contain a date, we should append increment on that date
[campaigns](https://developers.google.com/google-ads/api/fields/v11/campaign)
[account_performance_report](https://developers.google.com/google-ads/api/docs/migration/mapping#account_performance)
[ad_group_ad_report](https://developers.google.com/google-ads/api/docs/migration/mapping#ad_performance)
[display_keyword_report](https://developers.google.com/google-ads/api/docs/migration/mapping#display_keyword_performance)
[display_topics_report](https://developers.google.com/google-ads/api/docs/migration/mapping#display_topics_performance)
[shopping_performance_report](https://developers.google.com/google-ads/api/docs/migration/mapping#shopping_performance)
[user_location_report](https://developers.google.com/google-ads/api/fields/v11/user_location_view)

[...]

PR #115, opened 2023-05-04
google ads pipeline

redicane:
# Tell us what you do here

- [] contributing new pipeline (please create an issue [here]() and link it below)
- [x] implementing verified pipeline (please link a relevant issue labelled as `verified pipeline`)

#87  Relevant issue


pipelines/google_ads/__init__.py
```diff
@@ -0,0 +1,141 @@
+from typing import Dict, List, Union
+import dlt
+from dlt.common.configuration.specs import GcpClientCredentialsWithDefault
+from pipelines.google_analytics.helpers.credentials import GoogleAnalyticsCredentialsOAuth
```
rudolfix, line 4: please do not import cross-pipelines. we have good support for OAUTH now. Import `GcpOAuthCredentialsWithDefault` from specs. it is similar to yours

pipelines/google_ads/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+    :param credentials:
+    :param dev_token:
+    :param query_list:
+    :return:
+    """
+
+    # generate access token for credentials if we are using OAuth2.0
+    if isinstance(credentials, GoogleAnalyticsCredentialsOAuth):
+        credentials.auth("https://www.googleapis.com/auth/analytics.readonly")
+        credentials = {
```
rudolfix, line 38: you can just use `to_native_representation` if you import from our specs (I hope - if not then tell me)

rudolfix: or just use `to_google_credentials` that will return google credentials object directly

pipelines/google_ads/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+            "developer_token": dev_token,
+            "refresh_token": credentials.refresh_token,
+            "client_id": credentials.client_id,
+            "client_secret": credentials.client_secret
+        }
+        client = GoogleAdsClient.load_from_dict(config_dict=credentials)
+    # use service account to authenticate if not using OAuth2.0
+    else:
+        pass
```
rudolfix, line 47: not yet done I assume? both credentials have `to_google_credentials` so you can just do it at the end?

pipelines/google_ads/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+               dev_token: str = dlt.secrets.value):
+    """
+    Loads default tables for google ads in the database.
+    :param credentials:
+    :param dev_token:
+    :return:
+    """
+    # generate access token for credentials if we are using OAuth2.0
+    if isinstance(credentials, GoogleAnalyticsCredentialsOAuth):
+        credentials.auth("https://www.googleapis.com/auth/analytics.readonly")
```
rudolfix, line 73: wrong scope I assume?

pipelines/google_ads/generate_user_credentials.py
```diff
@@ -0,0 +1,195 @@
...
+import socket
+import sys
+from urllib.parse import unquote
+
+# If using Web flow, the redirect URL must match exactly what’s configured in GCP for
+# the OAuth client.  If using Desktop flow, the redirect must be a localhost URL and
+# is not explicitly set in GCP.
+from google_auth_oauthlib.flow import Flow
+
+_SCOPE = "https://www.googleapis.com/auth/adwords"
```
rudolfix, line 32: why are we using the web flow not a desktop flow? we had nice script for google analytics. any reason?

rudolfix (changes requested):
Are you planning to implement incremental?

rudolfix:
here's the feedback to the current state:
## Authentication
A developer token is needed on top of regular google authentication. Only OAuth is currently implemented as a way of authenticating. The OAuthCredential Class in dlt core might need to be better
___
## Pipeline Script and Sources
These are setup currently. There are 2 sources: 1 for user queries and 1 for default tables. User queries are saved in config or given as input for the user queries source.
## Helpers and testing
Helpers are set up in helpers/data_processing.py to process the responses from the google api client. The problem being that responses are mostly Objects which have no implemented to_dict() or to_json() method. Some testing sample functions are also set up.
___
## Resources
For both the resources of the default and user query tables major changes are needed. A query is needed to get this data and in Google Ads Query language “Select *” is not a thing, so column names need to be included in the config which will make the whole thing bloated.

pipelines/google_ads/__init__.py
sh-rp: done

pipelines/google_ads/__init__.py
sh-rp: we need a dict here to inject the developer key, somehow google ads works a bit different. we can discuss

pipelines/google_ads/__init__.py
sh-rp: fixed

PR #126, opened 2023-05-10
adding mux pipeline

TyDunn:
# Tell us what you do here
contributing new pipeline: https://github.com/dlt-hub/pipelines/issues/125


pipelines/mux/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+@dlt.resource(write_disposition="merge")
+def assets_resource(MUX_API_ACCESS_TOKEN=dlt.secrets.value, MUX_API_SECRET_KEY=dlt.secrets.value):
+    """Passes metadata about every asset to be loaded"""
+    url = f"{API_BASE_URL}/video/v1/assets"
+    limit = 100 # loads only 100 assets
+    params = { "limit": limit }
+
+    response = requests.get(url, params=params, auth=HTTPBasicAuth(MUX_API_ACCESS_TOKEN, MUX_API_SECRET_KEY))
+    response.raise_for_status()
+    for item in response.json()["data"]:
```
rudolfix, line 27: you can just yield a list. I recommend it to everyone: dlt will process them much faster and transparent to the user.

pipelines/mux/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    yest_end = int(time.mktime(today.timetuple()))
+
+    while True:
+        params = { "limit": limit, "page": page, "timeframe[]": [yest_start, yest_end] }
+        response = requests.get(url, params=params, auth=HTTPBasicAuth(MUX_API_ACCESS_TOKEN, MUX_API_SECRET_KEY))
+        response.raise_for_status()
+        data = response.json()["data"]
+        if data == []:
+            break
+        for item in data:
```
rudolfix, line 48: same thing. also your code will be simpler :)

pipelines/mux/requirements.txt
```diff
@@ -0,0 +1,2 @@
+dlt
+requests
```
rudolfix, line 2: requests are part of dlt, you do not need to specify them

rudolfix (changes requested):
regarding merging this pipeline. we'll probably move the community pipelines somewhere else. your pipeline does not really follow the structure of pipelines repo. probably the SHARE_PIPELINE doc was not good enough. but instead of fixing this we'll rather create a place where `dlt init` project can be pushed directly and then we can resubmit this without changes

still please see my review and improve the yielding

rudolfix:
@TyDunn I've found an account that @ahoffix created and was able to add test for it. so now we can merge it. pipeline is really neat and a good start for someone to add more resources to it. we we'll document it and keep with the others. thx!

rudolfix (approved):

PR #128, opened 2023-05-10
firebase pipeline initial commit

Anggi-Permana-Harianja:
# Tell us what you do here

- [ ] implementing verified pipeline (please link a relevant issue labelled as `verified pipeline`)

# Relevant issue

issue #
[link](https://github.com/dlt-hub/pipelines/issues/109)

# More PR info


Anggi-Permana-Harianja:
WIP, currently only covers realtime database, next would be cover crashlytic.

pipelines/firebase/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+import dlt
+from dlt.extract.source import DltResource
+from dlt.common.typing import TDataItem
+
+from firebase_admin import credentials, db
+
+
+@dlt.source(name="firebase")
+def firebase_source(
+    project_id: str =  dlt.secrets.value,
```
rudolfix, line 14: we have a building block for google service credentials, please see for example:
If your source/resource needs google credentials, just use dlt built-in credentials as we do in [google sheets](https://github.com/dlt-hub/pipelines/blob/master/pipelines/google_sheets/__init__.py#L26) and [google analytics](https://github.com/dlt-hub/pipelines/blob/master/pipelines/google_analytics/__init__.py#L32). Also note how credentials.to_native_credentials() is used to initialize google api client.

pipelines/firebase/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+
+        # initialize app
+        try:
+            firebase_admin.initialize_app(
+                credentials, 
+                {"databaseUrl": database_url}
+            )
+        except ValueError:
+            firebase_admin.get_app(name="[DEFAULT]")
```
rudolfix, line 48: could you explain what the code does here? what is `get_app`? should the `name` be configurable?

pipelines/firebase/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+        ref = db.reference("/")
```
rudolfix, line 50: is the path configurable?

pipelines/firebase/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        data = json.loads(ref.get())
```
rudolfix, line 51: this is getting the whole database at once. any way to select or filter data? does the concept of incremental load apply here?

pipelines/firebase/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+        """Please adjust the loop part according to your firebase database json schema"""
+        for item in data["discography"]:
```
rudolfix, line 54: this should be placed outside of the pipeline. see https://dlthub.com/docs/general-usage/resource#filter-transform-and-pivot-data and demonstrate this transformer in demo script

rudolfix (changes requested):
this pipeline is interesting. we should let people easily extract a stream of data from firebase and transform it separately. see my review.

pipelines/firebase/__init__.py
Anggi-Permana-Harianja: unfortunately this is a buildt-in method from firebase-admin module and I dont think it can be configureable

pipelines/firebase/__init__.py
Anggi-Permana-Harianja: I have a lil bit confusion here, if we put transformation/parsing outside the pipeline, how the pipeline will store the parsed data? is there any example of pipeline that does it?

pipelines/firebase/__init__.py
rudolfix: OK! another question: do those objects contain unique id? does it depend on database schema?

Anggi-Permana-Harianja: we can set the object name using `db.child("collection_name")`, otherwise it will be random

pipelines/.dlt/example.secrets.toml
```diff
@@ -123,4 +123,32 @@ password=""
...
+
+[sources.firebase]
+project_id="set me up!"
+private_key_id="set me up!"
+private_key="set me up!"
+client_email="set me up!"
+client_id="set me up!"
+database_url="set me up!"
+
+[sources.facebook_ads]
```
rudolfix, line 136: most of those look duplicated, please remove the duplicates

pipelines/firebase/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+from dlt.sources.credentials import GcpOAuthCredentials, GcpServiceAccountCredentials
+from dlt.common.exceptions import MissingDependencyException
+
+try:
+    from firebase_admin import credentials, db
+except ImportError:
+    raise MissingDependencyException("Firebase Admin Python SDK", ["firebase-admin"])
+
+
+_DATA =  {
```
rudolfix, line 17: this is a demo  case. please move it to demo script. best way would be to create another inline source in firebase_pipleine.py that just creates the app and fixture and then call it (like you call this one). please make this pipeline generic

rudolfix (changes requested):
A few things are still missing to get it merged:
1. typing is wrong / does not work in Python 3.8. if you want that mypy checks it follow this: https://github.com/dlt-hub/pipelines#typing-and-linting
2. the least is to make linter work: https://github.com/dlt-hub/pipelines#typing-and-linting
3. add the firebase dependency to poetry group as described here https://github.com/dlt-hub/pipelines#pipeline-specific-dependencies

rudolfix:
@Anggi-Permana-Harianja looks like pyproject toml is incorrectly merged and still the test data is in pipeline code. we reorganized the pipelines code and I've moved your fork as our branch. https://github.com/dlt-hub/pipelines/tree/rfix/firebase-attempt
we'll continue to work on it later.

PR #135, opened 2023-05-11
Sql database examples

### LINKED ISSUE #75: Add examples to sql pipeline
As usage documentation, we need some examples in the SQL pipeline that show people how to configure the pipeline

Existing example: How to load all database 
usual use case
- Load some tables with "replace"
- load non stateful tables with "append"
- load some tables with merge

So we would like to show the following pipeline examples:
1. "load all tables" A case where we have a list of tables. The user will configure incremental loading for some tables and leave the rest to load as "replace"
2 "load selected tables" Same as before, but the user must specify "replace" tables as well, and anything else defaults to not load. 

We should demonstrate two methods:
1. with `sql_database` where we reflect the tables and selectively apply hints as specified above and load
2. with `sql_table` standalone resource where user knows the name of the tables and configures each one then loads together.

We may actually use one of public mysql databases as a demo source: ie. test `test_load_mysql_data_load` and 
```python
# reflect a database
credentials = ConnectionStringCredentials("mysql+pymysql://rfamro@mysql-rfam-public.ebi.ac.uk:4497/Rfam")
database = sql_database(credentials)
```
where we also demonstrate how to pass credentials explicitly...

steinitzu:
# Tell us what you do here

- [x] improving, documenting, customizing existing pipeline (please link relevant issue)

Adds usage examples for the sqlalchemy pipeline.   
* Using the sql_database source in multiple configuration with explicit credentials
* sql_database source to load entire database
* Using the standalone resource to load some known tables

Changed the standalone table resource to be a normal function wrapped in `with_config`. The nested resource causes issues with the incremental wrapper

# Relevant issue

closes #75

### DIFF
--- a/pipelines/.dlt/example.secrets.toml
+++ b/pipelines/.dlt/example.secrets.toml
@@ -56,7 +56,12 @@ pipedrive_api_key="set me up!"
 
 # Example database for source in sql_database/source.py
 [sources.sql_database]
-database_url = "postgresql://my_user:my_password@localhost:5439/my_database"
+drivername = "mysql+pymysql"
+database = "Rfam"
+username = "rfamro"
+# password = ""
+host = "mysql-rfam-public.ebi.ac.uk"
+port = "4497"
 
 # not all these credentials are needed, you can choose between
 # email + password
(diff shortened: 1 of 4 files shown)


rudolfix (commented):
looks good to me! thanks!
@adrianbr please take a look if this demo is enough

adrianbr:
Looks good!

adrianbr (approved):

Issue #136, opened 2023-05-11
Notion verified pipeline


burnash:
# Planned Pipeline Info

- **name:** notion
- **category and description**: CRM, loads the database data from Notion API

You can fill the data below when writing spec.

# Use Cases
- Retrieve and load CRM data for sales lead analysis
- Extracting project management data: Import project tasks, assignees, due dates, and progress

# Sources / Resources / Endpoints
- `notion_databases(database_ids: Optional[List[Dict[str, str]]] = None)` - receives a list of dictionaries each containing a database id and a name to use as a table in a destination.

# Test account / test data
- A notion account with an an [integration](https://www.notion.so/my-integrations) set up is needed.

# Implementation tasks
Below you have a proposal for implementation tasks
* [ ] implement all the described sources, resources and endpoints
* [ ] make sure that the Use Cases can be easily executed by the pipeline's user by **providing demonstrations of all the use cases** in the `[pipeline_name]_pipeline.py`
* [ ] all the use cases are tested
* [ ] if there are any customizations you must test them, including unit tests for custom code

PR #137, opened 2023-05-11
Notion pipeline

burnash:
Implementing a verified pipeline 

# Relevant issue

issue #136


pipelines/notion/client.py
```diff
@@ -0,0 +1,156 @@
...
+        return response.json()
+
+    def search(
+        self,
+        query: Optional[str] = None,
+        filter_criteria: Optional[Dict[str, Any]] = None,
+        sort: Optional[Dict[str, Any]] = None,
+        start_cursor: Optional[str] = None,
+        page_size: Optional[int] = None,
+    ) -> Dict[str, Any]:
```
rudolfix, line 114: this is Generator so correct type would be `Iterator[...]`.

if you want your pipeline to be type-checked, place a file called `py.typed` in the `./notion` folder. this is how our Makefile selects pipelines for type checking

pipelines/notion/client.py
```diff
@@ -0,0 +1,156 @@
...
+                'page_size': page_size,
+            }
+
+            filtered_payload = self._filter_out_none_values(payload)
+
+            response = self.send_payload(
+                'search', '', payload=filtered_payload
+            )
+
+            for result in response.get('results', []):
```
rudolfix, line 151: it is way faster to yield lists, not item by item. `dlt` will handle both lists and items same way, transparently for the user. list goes through pipeline as single item. if you use transformers, it is really way more efficient

pipelines/notion/database.py
```diff
@@ -0,0 +1,79 @@
...
+        while has_more:
+            response = self.notion_client.send_payload(
+                'databases',
+                self.database_id,
+                subresource='query',
+                query_params=filter_properties,
+                payload=payload,
+            )
+
+            for result in response.get('results', []):
```
rudolfix, line 75: same thing

rudolfix (changes requested):
Thanks for the test, docs and docstrings, this is really high quality!

regarding the mock tests. I see some use for them for example to reproduce some edge cases from  customers that are hard to add to our test account. we can discuss how to use them

still I see a lot of value in testing on real account. ie. authorization, api changes, adding new test cases is IMO easier.

as we have many test in mocks, you can add just one test where we load a few tables from notion. please try many data types for tables (in Notion) so we can see how they are mapped to python values

pipelines/notion/client.py
burnash: I use `search()` in `notion_databases()` to filter database ids myseld here:

```python
if database_ids is None:
    search_results = notion_client.search(
        filter_criteria={'value': 'database', 'property': 'object'}
    )
    database_ids = [
        {'id': result['id'], 'use_name': result['title'][0]['plain_text']}
        for result in search_results
    ]
```

If I yield a list, the usability of `search()` could suffer, since the calling code would need to handle chaining. For this particular case I'm not sure if the performance gains would be that visible. Happy to adjust if I'm wrong.

pipelines/notion/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+import dlt
+from dlt.extract.source import DltResource
+
+from .client import NotionClient
+from .database import NotionDatabase
+
+
+@dlt.source
+def notion_databases(
+    database_ids: Optional[List[Dict[str, str]]] = None,
```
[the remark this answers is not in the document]
burnash: > if you set it to dlt.config.value the argument is required and None value is not allowed

What's the optimal way to let the value be None? The way I put it, when the `database_ids` is missing (and defaults to None), `notion_databases` queries all available databases.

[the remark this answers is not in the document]
rudolfix: ahhh then it is good. nothing to change there. what I mean is that you can still use `config.toml` to pass `database_ids`

pipelines/notion/client.py
rudolfix: yes you are right (I feel like GPT-4 now ;p) you are using search not to query data, just the database ids... then all is good and we should keep yielding single elements

rudolfix (approved):

Issue #131, opened 2023-05-10
Workable verified pipeline


AstrakhantsevaAA:
# Planned Pipeline Info
- **name** of the pipeline: workable
- **category and description** of the pipeline: loads the relevant data from Workable API

# Use Cases
- load data from any endpoint with **replace** mode, except endpoint "candidates" - we can load it **incrementally**
- load data from dependent endpoints  with `@dlt.transformer`
- load filtered data by date

# Sources / Resources / Endpoints

Sources:

**workable_source** 
Retrieves data from the Workable API for the specified endpoints.
For almost all endpoints, Workable API responses do not provide keys "updated_at", so in most cases we are forced to load the date in 'replace' mode. 'Сandidates' are the only endpoints that have a key 'updated_at', which means that we can update the data incrementally in 'merge' mode.

Resources:

- "members",
- "recruiters",
- "stages",
- "requisitions",
- "jobs",
- "custom_attributes",
- "candidates"

Resources that depend on another resource are implemented as transformers, so they can re-use the original resource data without re-downloading:

resources with write_disposition='merge'
- "candidates_activities"
- "candidates_offer"

resources with write_disposition='replace'
- "jobs_activities"
- "jobs_application_form"
- "jobs_questions"
- "jobs_stages"
- "jobs_custom_attributes"
- "jobs_members"
- "jobs_recruiters"

# Test account / test data

Workable test account provide test data.

rudolfix:
done :)

PR #143, opened 2023-05-12
Workable pipeline

AstrakhantsevaAA:
# Tell us what you do here
Created pipeline to load data from Workable API:
- load data from any endpoint with replace mode, except endpoint "candidates" - we can load it incrementally
- load data from dependent endpoints with @dlt.transformer
- load filtered data by date

# Relevant issue

issue #131 

# More PR info


rudolfix:
@AstrakhantsevaAA if you want your code to be type checked add `py.typed` empty files and do `make lint`. currently you have invalid typing and `init` linter will not pass (probably that's python 3.8 on which linter runs but we support it officially)

PR #149, opened 2023-05-22
add salesforce pipeline

sh-rp:
# Tell us what you do here

- [ ] implementing verified pipeline (please link a relevant issue labelled as `verified pipeline`)

# Relevant issue

issue #

# More PR info


pipelines/salesforce/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+    password: str = dlt.secrets.value,
+    security_token: str = dlt.secrets.value,
+):
+    client = Salesforce(username, password, security_token)
+    # You can add more sobjects here, custom ones too
+    # See standard objects here:
+    # https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_list.htm
+    # See the `sobject_factory` function docstring for info on the arguments
+    return (
+        sobject_factory(name="sf_user", sobject="User", client=client, deferrable=True),
```
sh-rp, line 156: the table name now is sf_user, because I could not get the table count query to work with all destinations at the same time. for duckdb a table named user must be escaped, for bigquery the escapes do not work, but an unescaped table named user does work.

pipelines/salesforce/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+
+def get_records(
+    sf: Salesforce,
+    sobject: str,
+    state: t.Optional[dict] = None,
+    replication_key: t.Optional[str] = None,
+    fields: t.Optional[t.List[str]] = None,
+) -> t.Iterable[dict]:
+    """Return an iterable of records from the Salesforce API."""
+    print(f"Starting {sobject} extraction...")
```
rudolfix, line 34: pls remove the print statements. now we have a progress monitoring via `progress` argument to pipeline

pipelines/salesforce/__init__.py
rudolfix: BigQuery uses different quotation (``) to escape identifiers. we can use have `user` if you use `sql_client.make_qualified_table_name` to escape the table name.

yeah but let's keep sf_user. it will make writing queries easier

pipelines/salesforce/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+
+    You can override this function based on your own setup. It will
+    add a LIMIT 100 clause to the queries.
+    """
+    return True
+
+def get_records(
+    sf: Salesforce,
+    sobject: str,
+    state: t.Optional[dict] = None,
```
rudolfix, line 29: we should not use state directly. it is better to create `dlt.sources.incremental` argument on the `replication_key`, there are plenty of examples on how to do that

pipelines/salesforce/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+        f["name"] for f in desc["fields"] if f["type"] in ("datetime",) and f["name"]
+    }
+    # If no fields are specified, use all fields except compound fields
+    if not fields:
+        fields = [f["name"] for f in desc["fields"] if f["name"] not in compound_fields]
+
+    # Generate a predicate to filter records by the replication key
+    predicate, order_by, n_records = "", "", 0
+    if replication_key:
+        bookmark = state.get("last_replication_key")
```
rudolfix, line 60: this is `last_value` in incremental

pipelines/salesforce/__init__.py
```diff
@@ -0,0 +1,226 @@
...
+    return (
+        sobject_factory(name="sf_user", sobject="User", client=client, deferrable=True),
+        sobject_factory(
+            name="user_role", sobject="UserRole", client=client, deferrable=True
+        ),
+        sobject_factory(
+            name="opportunity",
+            sobject="Opportunity",
+            client=client,
+            write_disposition="append",
```
rudolfix, line 164: looks like all the append resources are on modification date. this means that those objects can be retrieved many times and we'll generate duplicates. IMO we need to change it to `merge`
https://help.salesforce.com/s/articleView?id=000387261&type=1

rudolfix (changes requested):
our account is locked out. I'm afraid we must decrease the parallelism in CI workflows to 1.

pipelines/salesforce/__init__.py
sh-rp: done

pipelines/salesforce/__init__.py
sh-rp: yes, I went through the pipeline code and rewrote it a bit to be nicer and more compact, this is taken into account now

pipelines/salesforce/__init__.py
sh-rp: done

pipelines/salesforce/__init__.py
sh-rp: done, do I need to specify the primary key if I use merge_ it seems to accept it without

pipelines/salesforce/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+    You can override this function based on your own setup. It will
+    add a LIMIT 100 clause to the queries.
+    """
+    return True
+
+def get_records(
+    sf: Salesforce,
+    sobject: str,
+    last_state: t.Optional[dict] = None,
```
rudolfix, line 30: this is last value, not last state, and is not a dict

please add `py.typed` to your pipeline folder so mypy will check the typing. see other pipelines for example

sh-rp: Done, why do we not enforce linting by default over the entire repo?

pyproject.toml
```diff
@@ -10,6 +10,7 @@ packages = [{include = "pipelines"}]
 [tool.poetry.dependencies]
 python = "^3.8.1"
 dlt = {version = "^0.2.7", extras = ["redshift", "bigquery", "postgres", "duckdb"]}
+simple-salesforce = "^1.12.4"
```
rudolfix, line 13: this should be removed from normal dependencies and placed in the respective group (this is also done!)

rudolfix (changes requested):
what is happening on windows: the gtihub runner defines env var `username` which gets injected over the google secrets configuration in salesforce pipeline. env var provider has the highest priority in `dlt`. best way to fix that is to change the argument in the @dlt.source to `user_name` :) that will fix the problem

tests/salesforce/test_salesforce_pipeline.py
```diff
@@ -0,0 +1,46 @@
...
+import dlt
+from pipelines.salesforce import salesforce_source
+
+@pytest.mark.parametrize('destination_name', ALL_DESTINATIONS)
+def test_all_resources(destination_name: str) -> None:
+
+    expected_tables = ['account', 'campaign', 'contact', 'lead', 'opportunity', 'pricebook_2', 'pricebook_entry', 'product_2', 'sf_user', 'user_role']
+
+    pipeline = dlt.pipeline(
+        pipeline_name="salesforce", destination=destination_name, dataset_name="salesforce_data",
```
rudolfix, line 12: add full_refresh=True. now you load data to the same dataset, there may be race condition on CI

rudolfix (changes requested):
@sh-rp add full_refresh=True to your test. now you load data to the same dataset, there may be race condition on CI

rudolfix (approved):

PR #153, opened 2023-05-29
[cleanup] asana pipeline

sh-rp:
Example cleanup of a pipeline in our repo

### DIFF
--- a/pipelines/asana_dlt/__init__.py
+++ b/pipelines/asana_dlt/__init__.py
@@ -9,6 +9,7 @@ functions are meant to be used as part of a data loading pipeline.
 import typing as t
 from typing import Sequence, Iterable, Dict, Any
 import dlt
+from dlt.common.typing import TDataItem
 
 from dlt.extract.source import DltResource
 from .settings import (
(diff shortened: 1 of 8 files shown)


pipelines/asana_dlt/__init__.py
```diff
@@ -34,48 +49,26 @@ def workspaces(access_token: str = dlt.secrets.value) -> t.Iterator[dict]:
...
+    Fetches and returns a list of projects for a given workspace from Asana.
+
+    Args:
+        workspace (dict): The workspace data.
+        access_token (str): The access token to authenticate the Asana API client, provided in the secrets file
+
+    Returns:
+        list[dict]: The project data for the given workspace.
+    """
     print(f"Fetching projects for workspace {workspace['name']}...")
```
rudolfix, line 65: we can remove those prints. we have `progress` argument to pipeline that does exactly that

pipelines/asana_dlt/__init__.py
```diff
@@ -34,48 +49,26 @@ def workspaces(access_token: str = dlt.secrets.value) -> t.Iterator[dict]:
     write_disposition="replace",
 )
 @dlt.defer
-def projects(workspace, access_token: str = dlt.secrets.value) -> t.Iterator[dict]:
-    """Returns a list of projects for a given workspace"""
+def projects(
+    workspace: Dict[str, Any], access_token: str = dlt.secrets.value
+) -> Iterable[Dict[str, Any]]:
```
rudolfix, line 54: there are two types to mark returned items `TDataItems` in this case fits

pipelines/asana_dlt/__init__.py
```diff
@@ -286,28 +240,47 @@ def teams(workspace, access_token: str = dlt.secrets.value) -> t.Iterator[dict]:
...
-                ),
+                timeout=TIMEOUT,
+                opt_fields=",".join(USER_FIELDS),
             )
         ]
     finally:
         print(f"Done fetching users for workspace {workspace['name']}.")
 
 
 @dlt.source
```
rudolfix, line 270: IMO @source should be at the top

pipelines/asana_dlt/const.py
```diff
@@ -0,0 +1,142 @@
...
+    "followers",
+    "name",
+    "color",
+    "notes",
+    "permalink_url",
+    "workspace",
+]
+
+# List of task fields to be retrieved from Asana API
+TASK_FIELDS = [
```
rudolfix, line 68: better to use tuples becaue you can make them defaults in the function signatures

pipelines/asana_dlt/const.py
```diff
@@ -0,0 +1,142 @@
+# Default start date for Asana API requests, only tasks started after this date will be collected
+DEFAULT_START_DATE = "2010-01-01T00:00:00.000Z"
+
+# Asana API request timeout
+TIMEOUT = 300
```
rudolfix, line 5: should be passed as an argument maybe?

sh-rp: hm, I would have this as a setting across all endpoints and keep it like this for now

PR #159, opened 2023-05-31
[cleanup] Facebook ads pipeline

sh-rp:


pipelines/facebook_ads/fb.py
```diff
@@ -24,80 +24,107 @@
...
 
-    if 'error' in data:
+    if "error" in data:
         raise Exception(f"Error refreshing token: {data['error']}")
 
     return data["access_token"]
 
 
-def get_ads_account(account_id: str, access_token: str, request_timeout: float) -> AdAccount:
+def get_ads_account(
```
rudolfix, line 75: looks like this should go to `helpers.py`. things in `fb.py` are useful utils for the user but actually not used by pipeline code. (we can rename that file to utils.py - and mention this in README)

pipelines/facebook_ads/fb.py
```diff
@@ -114,34 +141,42 @@ def execute_job(
     job: JobsJob,
```
rudolfix, line 141: same here

PR #171, opened 2023-06-01
[cleanup] notion pipeline

sh-rp:


burnash (changes requested):
Much better than it was before the cleanup. However, I'm not sure if change from `yield response.get('results', [])` to `yield from response.get("results", [])` is needed.

PR #178, opened 2023-06-12
Remove full_refresh args in Hubspot pipeline

mermi:
# Tell us what you do here

# Relevant issue

issue #

# More PR info
- We found out when the  then the pipeline will try ti create the dataset with a suffix as a timestamp, and we want to enable people to use their own datasets or schemas (depends on the destination)


adrianbr:
The test without full refresh is not needed here, the test simply asserts that data loaded successfully via load info, so it should have the same outcome

PR #186, opened 2023-06-20
Created Matomo README.md

dat-a-man:
# Tell us what you do here

- [ X] improving, documenting, customizing existing source (please link an issue or describe below)

# Relevant issue

issue #

# More PR info
Created Matomo README.md

### DIFF
--- a/sources/matomo/README.md
+++ b/sources/matomo/README.md
@@ -1,0 +1,74 @@
+# Matomo
+
+Matomo is a free and open-source web analytics platform that allows website owners and businesses to gain detailed insights into their website and application performance. With this verified source, you can easily extract and seamlessly load data from Matomo to your preferred destination. This verifies source supports the following endpoints:
+
+| Endpoint | Description |
+| --- | --- |
+| visits | the visits in the given site ID for the given period |
+| matomo_reports | the data is retrieved according to the given queries. For example, in the pipeline example |
+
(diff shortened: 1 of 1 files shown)


sources/matomo/README.md
```diff
@@ -0,0 +1,74 @@
+# Matomo
+
+Matomo is a free and open-source web analytics platform that allows website owners and businesses to gain detailed insights into their website and application performance. With this verified source, you can easily extract and seamlessly load data from Matomo to your preferred destination. This verifies source supports the following endpoints:
+
+| Endpoint | Description |
+| --- | --- |
+| visits | the visits in the given site ID for the given period |
```
AstrakhantsevaAA, line 7: these are not endpoints, but sources

it is not easy to understand what's happening in this project, so i suggest to keep it like that, but rename  `This verifies source supports the following endpoints:` to `This verifies source supports the following sources:`

AstrakhantsevaAA: or may be it makes sense to remove it altogether

adrianbr: yeah maybe just describe what resources be found, afaik
event data:
- visit events
- users of those visit events, with info about their session
aggregated data:
- reports

dat-a-man: please check 👍

dat-a-man:
Closed

sources/matomo/README.md
adrianbr: the intention here was to separate
- daily running (aggregated reports on date)
from 
- any-time top up with events
- but users require merge so maybe those not that often for cost reason

Issue #148, opened 2023-05-19
remove the zenpy from zendesk source


rudolfix:
**Pipeline name**
zendesk

**Make sure that**
* [x] I checked the README in the pipeline folder
* [] If I have problems with the credentials I checked the [credentials docs](https://dlthub.com/docs/general-usage/credentials) and [troubleshooting](https://dlthub.com/docs/walkthroughs/run-a-pipeline#4-detect-and-handle-problems)

**Describe the bug**
Zenpy is causing countless problems when using to load data. The most problematic are
- seems that caching is out of control and happens even if disabled. our source 
- zenpy object cannot be serialized to json. the built in methods do not work. they still contain non serializable objects
Current code uses brute force approach to serialize zenpy objects to Python dicts. Properties are json serialized and if that does not work they are converted to string with `str`.
Here's better approach from Singer: https://github.com/singer-io/tap-zendesk/blob/master/tap_zendesk/sync.py

**Suggested fix**
1. * [x] Remove zenpy dependence. Use requests to get data. One of the sources does that anyway. Mind that zendesk is throttling requests very often so make sure we handle that (zendesk talk does that)
2. * [x] Clients report 502 http errors that do not get retried. let's retry 50x errors (I think it is default for our requests): "502 bad gateway on the same page of tickets appearing after 3h of loading - but we cannot manage it as its managed by zenpy - could be partly resolved by the chunked backloading( without backloading, blocking issue)"
3. * [x] we should support loading ranges of tickets, chats etc. the problem description from the customer: "initial load too large and slow - should we support backloading with start/end intervals? should we have a general backloading concept?"

PR #197, opened 2023-06-25
Zendesk without zenpy

steinitzu:
# Tell us what you do here

# Relevant issue

issue #148 

# More PR info

Removes zenpy dependency and implements zenpy talk/support/chat API with requests.   
Proactive request throttling is not implemented yet. Will add this option in the general requests helper. But retries on 429 + retry-after are already working.


adrianbr (dismissed):

sources/zendesk/helpers/backloading.py
```diff
@@ -0,0 +1,24 @@
+from typing import Dict, Any
+
+from dlt.common import pendulum
+
+from dlt.common.configuration.specs import BaseConfiguration, configspec
+
+
+@configspec
+class SliceConfig(BaseConfiguration):
+    start: pendulum.DateTime = None
```
rudolfix, line 10: by this we are creating a new patter how to handle backloading and incremental. if we do this we should update all the other pipelines. currently we have following pattern:
- you can pass `start_date` or `initial_load_past_days` which will be used to initialize `initial_value` of incremental
- if not present incremental loads from the beginning
- the `start_date` is obviously ignored on any next pipeline run - incremental will use the last value to continue
- we didn't have option for `end_date`. that obviously would need to disable incremental altogether

from the user perspecitve the biggest advangtage is that you can jus run pipeline in a loop. the backloading and subsequent incremental runs use the same pipeline code

when I look how current thing works
- you can load a time period (from date to date). incremental is disabled
- or you can load from the beginning with incremental. so there's no way to backload and then use incremental
- or you need to build some custom DAGs where you do backloading and then incremental. but even in that case you need to pass the end time of last backload to the incremental run as initial value

maybe we could combine this patterns
- if we have end date then no incremental
- if we have start data but no end date then we use the start date as initial and incremental is enable
- if no range specified then incremental starts from the beginning (or some other default which needs to be documented)

rudolfix (changes requested):
I think that code is good. I'm just making sure that this backloading mechanism works as we want and we can then port it to other pipelines. please see what I suggest.

sources/zendesk/helpers/backloading.py
steinitzu: Considering this, probably these should be `incremental` params and we add this logic there.  

I.e. add to `dlt.sources.incremental` both `start_value` and `end_value` arguments.  
The start value would always supersede the state value (`initial_value` behaviour doesn't change)

Usage as e.g.

```python
@source
def zendesk_support(start_date: DateTime=SOME_DEFAULT, end_date: DateTime = None):
    if end_date:
        incremental_args = dict(start_value=start_date, end_value=end_date)
    else::
        incremental_args = dict(initial_value=start_date)
    
    yield ticket_resource(updated_at=dlt.sources.incremental[DateTime](**incremental_args))
```

The filtering in incremental would be pretty much the same I think, couple of differences when start/end values are supplied:

1. Only write `end_value` to state

    Instead of writing the current row value to state, we only write the highest `end_value` to state   
    e.g.  
     ```python
     new_state_value = self.last_value_func([end_value, current_state_value])
     ```
    -> Makes it possible to switch from chunked backloading to incremental. It just starts from end of highest range  
    
2. Stop the generator when end value is reached

    tbd how to do this, something like this is available in sources `add_limit`   
    Check e.g.
    
    ```python
    if row_value == end_value or self.max_value_func([row_value, self.end_value]) != self.end_value:
        # Reached end value, stop
        kill_generator_somehow()
    ```
    
This would mean you can backload, and then stop supplying the start/end args to go back to incremental.  
Or vice versa you can load incrementally new data and then backfill later.  

    
The only iffy part about this I can think of is if you're gonna be loading chunks in parallel. Since it's writing to state and there's no synchronization there. But solvable by e.g. loading the highest chunk first.

adrianbr: parallel will not be possible in all cases and that's OK - in airflow, the user would regulate that with https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#config-core-max-active-runs-per-dag

Issue #205, opened 2023-06-29
[Ustructured] verified source


AstrakhantsevaAA:
# Quick source info
- Name of the source: Unstructured [temporary]
- What is the data source: [Unstructured data: Invoices, PDF, HTML, other]

# What source does/will do

Implement the PDF/txt to structured data “verified source” that converts unstructured data from a specified data resource to structured data using provided queries.

# Test account / test data

For local folder:

Add test data to verified-sources/tests/filesystem and verified-sources/tests/unstructured data.
Test invoices are generated with ChatGPT and saved with different extensions:
- invoice_1.pdf
- invoice_2.txt
- invoice_3.jpg

For Google Drive:

Create test Google account  and put there generated invoices:
- invoice_1.pdf
- invoice_2.txt
- invoice_3.jpg

# Additional context

PR #209, opened 2023-07-04
Features/unstructured source

AstrakhantsevaAA:
# Tell us what you do here

# Relevant issue

issue #205 

# More PR info
Use langchain to extract structured data from unstructured (txt, pdf).

- Created verified source `filesystem`, which contains `local_folder` and `google_drive` resources.
- Created verified source `unstructured_data` with source `unstructured_to_structured`, which contains transformer-type resource `structured_resource`.


sources/filesystem/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+    DATA_DIR,
+    FOLDER_IDS,
+    STORAGE_FOLDER_PATH,
+    ClIENT_SECRET_PATH,
+)
+
+
+@dlt.resource(write_disposition="replace")
+def local_folder(
+    extensions: Optional[Sequence[str]] = None,
```
rudolfix, line 21: we'll convert it into fsspec destination so we can also read from buckets. to make it production grade:
- should descend into folders
- should do incremental loading using "modified" timestamp on the file
- should delete processed files as an option (this one is tricky but doable transactionally)

sources/filesystem/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+
+    for file in files:
+        if file.is_file():
+            yield {"file_path": file.as_posix()}
+
+
+@dlt.resource(write_disposition="replace")
+def google_drive(
+    extensions: Sequence[str] = (".txt", ".pdf"),
+    client_secret_path: str = ClIENT_SECRET_PATH,
```
rudolfix, line 60: we have really good support for google credentials. see `google_analytics` source. all your problems below are solved... I should probably review this way earlier (sorry if you lost time)

sources/filesystem/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+    Args:
+        extensions (Sequence[str]): A sequence of file extensions to filter the files. Only files with these extensions will be retrieved.
+            Defaults to (".txt", ".pdf").
+        client_secret_path (str): The path to the client secret JSON file for authenticating with Google Drive API.
+            Defaults to ClIENT_SECRET_PATH (see settings.py).
+        token_path (str): The path to the token JSON file for storing and reusing the authentication token.
+            Defaults to AUTHORIZED_USER_PATH (see settings.py).
+        credentials (Dict[str, Any]): The authorized user info in Google format for authenticating with Google Drive API.
+            Defaults to dlt.secrets.value.
+        folder_ids (Sequence[str]): A sequence of folder IDs from which to retrieve the files.
```
rudolfix, line 79: this is cool. maybe we should do that in local filesystem as well.

sources/filesystem/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+            Defaults to FOLDER_IDS (see settings.py).
+        storage_folder_path (str): The path to the local folder where the downloaded files will be stored.
+            Defaults to STORAGE_FOLDER_PATH (see settings.py).
+        download (bool): Indicates whether to download the files or not. If True, the files will be downloaded and stored locally.
+            If False, only the file names and IDs will be yielded without downloading.
+            Defaults to False.
+
+    Yields:
+        TDataItem: A dictionary representing a file. If download is True, the dictionary contains the following keys:
+            - "file_path" (str): The local path of the downloaded file.
```
rudolfix, line 89: again - modification time is missing. gdrive returns plenty other props that we should include

sources/unstructured_data/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+import dlt
+from dlt.extract.source import DltResource, TDataItem
+
+from .helpers import filetype_mapper, process_file_to_structured
+
+logging.basicConfig(format="%(asctime)s WARNING: %(message)s", level=logging.WARNING)
+
+
+@dlt.source(name="unstructured_to_structured")
+def unstructured_to_structured_source(
```
rudolfix, line 17: you do not really need that source. standalone resource is enough

sources/unstructured_data/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Returns:
+        Iterable[DltResource]: An iterable containing a resource with original unstructured data and resource with structured data.
+
+    """
+    if openai_api_key:
+        os.environ["OPENAI_API_KEY"] = openai_api_key
+
+    @dlt.transformer(
+        data_from=unstructured_resource,
```
rudolfix, line 40: do not bind resource here. you can use late binding (with pipe | operator). just ask our GPT-4 (how to late bind resource and transformer?)

sources/unstructured_data/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        name=f"structured_data_from_{unstructured_resource.name}",
+    )
+    def structured_resource(unstructured_item: TDataItem) -> Iterable[TDataItem]:
+        file_path = unstructured_item["file_path"]
```
rudolfix, line 44: so here is the biggest challenge: the format of "unstructured_item". we'll have many sources of possibly unstructured data:
- file system (files)
- gdrive (files)
- inboxes
- message queues
- web page scrapers
- webhooks
the point is that we should be able to connect all of those to this transformer. (it needs to know that there is a file in "file_path" with certain type ("mime-type") etc.)
let's discuss it when I'm back.

sources/unstructured_data/helpers.py
```diff
@@ -0,0 +1,42 @@
+import logging
+from typing import Any, Dict
+
+from langchain.document_loaders import UnstructuredFileLoader, UnstructuredPDFLoader
+from langchain.indexes import VectorstoreIndexCreator
+
+logging.basicConfig(format="%(asctime)s WARNING: %(message)s", level=logging.WARNING)
+
+filetype_mapper = {
+    ".txt": UnstructuredFileLoader,
```
rudolfix, line 10: cool! but IMO langchain has something that will pick the parser for a given file type?

sources/filesystem/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+from .settings import (
+    AUTHORIZED_USER_PATH,
+    DATA_DIR,
+    FOLDER_IDS,
+    STORAGE_FOLDER_PATH,
+    ClIENT_SECRET_PATH,
+)
+
+
+@dlt.resource(write_disposition="replace")
```
rudolfix, line 19: I think it should be `selected=False` by default. then the data will be passed to transformer and not stored as separate table

sources/unstructured_data/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        extension = Path(file_path).suffix
+        if extension in filetype_mapper:
+            loader = filetype_mapper[extension](file_path)
+            yield process_file_to_structured(loader, queries)
```
rudolfix, line 48: you should include the "unstructured_item" in the data that you yield. ie.
```
info = process_file_to_structured(loader, queries)
info["metadata"] = unstructured_item
yield info
```

sources/unstructured_data/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    Returns:
+        Iterable[DltResource]: An iterable containing a resource with original unstructured data and resource with structured data.
+
+    """
+    if openai_api_key:
+        os.environ["OPENAI_API_KEY"] = openai_api_key
+
+    @dlt.transformer(
+        data_from=unstructured_resource,
+        name=f"structured_data_from_{unstructured_resource.name}",
```
rudolfix, line 41: there is a trick that you can use to set the name of resource when it is created. look for `sql_table` resource. you just declare a transformer with generic name and it returns a transformer with a dynamic name.

sources/unstructured_data/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+        loader: (Any) The loader that uses unstructured to load files. E.g. UnstructuredFileLoader, UnstructuredPDFLoader.
+        queries (Dict[str, str]): A dictionary of queries to be applied to the loaded file.
+            Each query maps a field name to a query string that specifies how to process the field.
+
+    Returns:
+        Dict[str, str]: A dictionary containing the processed structured data from the loaded file.
+            The dictionary includes a "file_path" key with the path of the loaded file and
+            additional keys corresponding to the queried fields and their processed values.
+    """
+    index = VectorstoreIndexCreator().from_loaders([loader])
```
rudolfix, line 38: we could probably pick the vector database as well via config?

sources/unstructured_data/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+    response = {"file_path": loader.file_path}
+    for k, query in queries.items():
```
rudolfix, line 40: we can ran this in parallel or async, (I hope that langchain supports async directly)

sources/unstructured_data/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+    ".pdf": UnstructuredPDFLoader,
+}
+
+
+def safely_query_index(index: Any, query: str) -> Any:
+    try:
+        answer = index.query(query).strip()
+        return answer
+    except Exception as e:
+        logging.warning(e)
```
rudolfix, line 20: ohhh we cannot just ignore exceptions. we need to see when they happen and how to handle them. ie. if openAPI call fails we can use tenacity to retry

sources/filesystem/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+        done = False
+        # Download the next chunk of data and check if the download is complete
+        while not done:
+            _, done = downloader.next_chunk()
+        # Write the downloaded file content to the specified file path
+        with open(file_path, "wb") as f:
+            f.write(fh.getvalue())
+
+    except HttpError as error:
+        print(f"An error occurred: {error}")
```
rudolfix, line 106: IMO you should raise here

sources/filesystem/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+from google.auth.transport.requests import Request
+from google.oauth2.credentials import Credentials
+from google_auth_oauthlib.flow import InstalledAppFlow
+from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError
+from googleapiclient.http import MediaIoBaseDownload
+
+from .settings import MIME_TYPE_MAPPER, SCOPES
+
+logging.basicConfig(format="%(asctime)s WARNING: %(message)s", level=logging.WARNING)
```
rudolfix, line 15: dlt has a logger:
```
from dlt.common import logger
```
and then use `logger` in the same way as `logging`

rudolfix (changes requested):
hey this looks amazing. thanks for docs and tests. my take is that we should use it to showcase how you can put langchain in production. to do that:
- our filesystem and gdrive must be upgraded (see my review)
- we must specify the interface to the transformer so people can easily connect different "transport resources" without writing custom adapters

next steps:
- fix the issues but please skip changing the filesystem and gdrive until we agree how they should work
- keep the unstrucutured item format as it is now

sources/unstructured_data/helpers.py
AstrakhantsevaAA: langchain has an extended list of vector bases, do we want to use them or create our own? https://python.langchain.com/docs/modules/data_connection/vectorstores/

sources/unstructured_data/helpers.py
AstrakhantsevaAA: honestly I didn't have any exception, so I don't know how to catch and handle e.g. call fails, I deleted this try-except for now

sources/unstructured_data/__init__.py
AstrakhantsevaAA: what about ?
```
unstructured_item.update(process_file_to_structured(loader, queries))

yield unstructured_item
```

sources/filesystem/__init__.py
AstrakhantsevaAA: it was stored because I returned it in unstructured_source. I was planned

sources/unstructured_data/helpers.py
AstrakhantsevaAA: right! actually UnstructuredFileLoader checks all files with `libmagic-dev` library.I should keep only this loader for all types of files

sources/unstructured_data/helpers.py
AstrakhantsevaAA: I gonna use some of the langchains vectorstores for now

rudolfix (approved):
OK this is so good. I have two remarks regarding the mime type detection. There's actually a function in Python to guess fhose from extensions. Please do not use MAGIC lib. it is a huge c library. we'll think about that later.

you can also merge this. we can fix the mime types later

rudolfix (approved):

Issue #214, opened 2023-07-14
add backloading and fix the custom fields state dependency in zendesk


rudolfix:
**Tasks**
1. * [x] accept `end_date` argument to incremental sources that will be used in incremental backloading
2. * [x] demonstrate backloading in the demo script: typical case is that you backload a range and then run incremental loading from the end of the range
3. * [x] test the `end_date` in some essential endpoints ie. tickets
4. * [x] if `pivot_custom_fields`  is true we always must fetch the custom fields and update the state! the reason: if someone uses `with_resources` to select just tickets we'll not get the `ticket_fields_table` resource and the state is not updated
5. * [x] in `process_ticket`:
```
for custom_field in ticket["custom_fields"]:
        if pivot_custom_fields:
            cus_field_id = str(custom_field["id"])
            field = custom_fields[cus_field_id]
```
allow for a field to not exist! just log a warning and fallback to the action we do when pivots are not enabled. why? it is quite possible that custom fields may change when pipeline runs. just prevent failure on that

PR #215, opened 2023-07-14
Zendesk backload and custom fields

steinitzu:
# Tell us what you do here

# Relevant issue

https://github.com/dlt-hub/verified-sources/issues/214

# More PR info

* Added `end_value` option to all sources. But this needs dlt updated with the fix in https://github.com/dlt-hub/dlt/pull/495  , otherwise it's mostly broken

* Always create the custom fields at start of ticket resource when `pivot_fields` is enabled. The `ticket_fields` as a resource which creates a table is still the same and must be selected if you want the db table  

* Fallback for non-existent custom fields -> processed same as without pivoting


sources/zendesk/__init__.py
```diff
@@ -28,13 +29,16 @@
...
     credentials: TZendeskCredentials = dlt.secrets.value,
     incremental_start_time: Optional[pendulum.DateTime] = DEFAULT_START_DATE,
+    incremental_end_time: Optional[pendulum.DateTime] = None,
 ) -> Iterable[DltResource]:
     """
     Retrieves data from Zendesk Talk for phone calls and voicemails.
 
     Args:
         credentials (TZendeskCredentials): The credentials for authentication. Defaults to the value in the `dlt.secrets` object.
         incremental_start_time (Optional[pendulum.DateTime]): The start time for incremental loading. Defaults to DEFAULT_START_DATE.
```
rudolfix, line 39: let's rename that to `start_time` and give updated description ie. The start time of the range for which to load data

sources/zendesk/__init__.py
```diff
@@ -28,13 +29,16 @@
...
+        incremental_end_time: Optionally load records only until this date/time.
```
rudolfix, line 40: also rename. and give the correct description: `The end time of the range for which to load data. If end time is not provided, the incremental loading will be enabled and after initial run, only new data will be retrieved"

sources/zendesk/__init__.py
```diff
@@ -122,13 +130,16 @@ def talk_incremental_resource(
...
     credentials: ZendeskCredentialsOAuth = dlt.secrets.value,
     incremental_start_time: Optional[pendulum.DateTime] = DEFAULT_START_DATE,
+    incremental_end_time: Optional[pendulum.DateTime] = None,
 ) -> Iterable[DltResource]:
     """
     Retrieves data from Zendesk Chat for chat interactions.
 
     Args:
         credentials (ZendeskCredentialsOAuth): The credentials for authentication. Defaults to the value in the `dlt.secrets` object.
         incremental_start_time (Optional[pendulum.DateTime]): The start time for incremental loading. Defaults to DEFAULT_START_DATE.
```
rudolfix, line 140: also here

sources/zendesk/helpers/api_helpers.py
```diff
@@ -12,6 +14,9 @@
 )
 
 
+log = logging.getLogger(__name__)
```
rudolfix, line 17: better if you import logging module from dlt? what do you think?

sources/zendesk_pipeline.py
```diff
@@ -52,7 +59,7 @@ def incremental_load_all_start_time() -> Any:
     # Choosing starting point for incremental load - optional, the default is the last load time. If no last load time
     # the start time will be the 1st day of the millennium
     # start time needs to be a pendulum datetime object
-    start_time = pendulum.DateTime(year=2023, month=1, day=1)
+    start_time = pendulum.DateTime(year=2023, month=1, day=1).in_timezone("UTC")
```
rudolfix, line 62: uh, is it really necessary? pendulum imported from dlt module is patched to always use UTC

tests/zendesk/test_zendesk_source.py
```diff
@@ -166,6 +167,32 @@ def test_incrementing(destination_name: str) -> None:
...
+    assert_load_info(info)
+
+    with pipeline.sql_client() as client:
+        rows = [
+            pendulum.instance(row[0])
+            for row in client.execute_sql("SELECT updated_at FROM tickets")
+        ]
+
+    assert all(value < incremental_end_time for value in rows)
+
```
rudolfix, line 194: what happens if we run the pipeline again, starting from `incremental_end_time`? will we get all records? could you test it?

another thing we should test is if incremental_end_time is exactly a timestamp of an ticket, then are we getting it twice? this is more like is the range open or closed. please add such information to docs after you figure it out (I think the answer is in the zendesk api)

rudolfix (changes requested):
@steinitzu the biggest problem I see is that in `talk_incremental_resource` and other helpers you are not passing the `end_time` to the zendesk API. so we load all the data and use incremental to filter parts we need. is this API not supporting `until` or similar param?

sources/zendesk_pipeline.py
steinitzu: I thought so too,  parsed tz naive dates and such all assume UTC, but does not seem to apply to the `DateTime` constructor:

```python
In [6]: pendulum.parse("2022-03-03T01:02:03")
Out[6]: DateTime(2022, 3, 3, 1, 2, 3, tzinfo=Timezone('UTC'))

In [7]: pendulum.DateTime(2023, 1, 1, 1, 1, 1)
Out[7]: DateTime(2023, 1, 1, 1, 1, 1)
```

Not sure if there's some simple way to prevent naive dates entirely. Maybe we can patch the `DateTime` class

steinitzu:
> @steinitzu the biggest problem I see is that in `talk_incremental_resource` and other helpers you are not passing the `end_time` to the zendesk API. so we load all the data and use incremental to filter parts we need. is this API not supporting `until` or similar param?

I think not. There's only `start_time` param mentioned in the docs  https://developer.zendesk.com/api-reference/voice/talk-api/incremental_exports/#request-parameters  , support api also https://developer.zendesk.com/api-reference/ticketing/ticket-management/incremental_exports/#query-string-parameters  

So it's relying on the "stop iteration" thing in incremental now. But could do the same checks here instead, i.e. check first item per page.

rudolfix:
OK you are right. no end_date in the API...

tests/zendesk/test_zendesk_source.py
steinitzu: Added second run.  

The open/closed range is tested in dlt. So don't think it needs another test here, right? But I added the info to the docstring in the sources.

pyproject.toml
```diff
@@ -10,7 +10,7 @@ packages = [{include = "sources"}]
 [tool.poetry.dependencies]
 python = "^3.8.1"
 black = "^23.3.0"
-dlt = {version = "^0.3.0", allow-prereleases = true, extras = ["redshift", "bigquery", "postgres", "duckdb"]}
+dlt = {git = "https://github.com/dlt-hub/dlt.git", branch="fix/incremental-end-value-with-batches", allow-prereleases = true, extras = ["redshift", "bigquery", "postgres", "duckdb"]}
```
rudolfix, line 13: let's set the pre-release here. stay tuned.

sources/github/__init__.py
```diff
@@ -104,8 +104,7 @@ def repo_events(
 
             # stop requesting pages if the last element was already older than initial value
             # note: incremental will skip those items anyway, we just do not want to use the api limits
-            if page and page[-1]["created_at"] < last_created_at.initial_value:
-                # do not get more pages, we overlap with previous run
+            if last_created_at.start_out_of_range:
```
rudolfix, line 107: cool!

sources/zendesk/helpers/api_helpers.py
rudolfix: this is what I meant in here:
```
from dlt.common import logger
```

logger has the same interface as `log`

tests/zendesk/test_zendesk_source.py
rudolfix: we do not need another test, just:
1. set `start_time` to exact timestamp of first ticker in our test data and make sure that we have this ticket in our db
2. set `end_time` to exact timestamps of one of the tickets in the middle of our test data and make sure it is in our db exactly once

I know this is a little bit of PITA but we are testing the API behavior (and also double testing our incremental exclusiveness)

rudolfix (changes requested):
I'll ping you with pre-release version of dlt here

tests/zendesk/test_zendesk_source.py
steinitzu: Aha got you, good to have this. I'll add it!

sources/pipedrive/__init__.py
```diff
@@ -184,16 +184,13 @@ def leads(
...
     )
     for page in pages:
-        if last_value:
-            # Just check whether first item is lower, worst case we load 1 redundant page before break
-            first_item = page[0] if page else None
-            if first_item and first_item["update_time"] < last_value:
-                return
         yield rename_fields(page, fields_mapping)
+        # flag is set when first lower than start time is hits, no need to request more data
+        if update_time.start_out_of_range:
```
rudolfix, line 195: to my knowledge this pipeline is already in production. we should keep the old version or detect and run the supported method

rudolfix:
@steinitzu `0.3.5a0` released

sources/pipedrive/__init__.py
steinitzu: Hmm ok best to leave the old version for now then. These changes in zendesk are incompatible also, should we be doing checks there?

rudolfix (approved):
LGTM! thank you!

PR #225, opened 2023-07-24
Features/unstructured source inbox

AstrakhantsevaAA:
# Tell us what you do here

# Relevant issue

issue #

# More PR info
Load all emails from gmail (not only with attachments)
- filter by emails
- incremental loading
- optional attachments downloading


sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+            if name and name in HEADERS_DEFAULT:
+                response[name] = header.get("value")
+                if name == "Date":
+                    response[name] = pendulum.parse(header.get("value"), strict=False)
+
+        body = part.get("body")
+
+        if part.get("filename"):
+            if "data" in body:
+                file_data = message.encode_attachment_data(body["data"])
```
rudolfix, line 103: this will typically be message body, please extract mime type so we can filter just pdfs later

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+            elif "attachmentId" in body:
+                response["attachment_id"] = body["attachmentId"]
```
rudolfix, line 105: here the same. extract mime types. pdf has `application/pdf` mime type.

we can filter for pdf by adding `add_filter` to the resource or you can add allowed_mimetypes argument.

sources/unstructured_data/inbox/helpers.py
```diff
@@ -0,0 +1,57 @@
...
+        self.client = gmail_client
+        self.user_id = user_id
+        self.message_id = message_id
+
+    def content(self) -> TDataItem:
+        return self.client.messages.get(
+            userId=self.user_id, id=self.message_id
+        ).execute()
+
+    def attachments(self, attachment_id: str) -> TDataItem:
```
rudolfix, line 48: this is not used, right?

rudolfix (changes requested):
this looks good! the biggest potential problem is again authorization and `user_id` which is always "me'. I'm not sure we'll use the invoices account to get oauth token or service account with domain delegation to read the e-mail. probably the latter... so user_id should be argument to resources.

also if we ever take this as a separate source I'd take a look at IMAP client and use it instead so it works with all inboxes (and can be authorized with user/password)

sources/unstructured_data/inbox/helpers.py
AstrakhantsevaAA: i use it in row 108

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+                attachment = message.attachments(attachment_id=body["attachmentId"])
+                file_data = message.encode_attachment_data(attachment["data"])
+            else:
+                file_data = None
+
+            if file_data is not None:
+                response["file_name"] = part["filename"]
```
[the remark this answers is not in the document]
AstrakhantsevaAA: no, but I can add it. we don't use it, just store

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+                    save_path = os.path.join(
+                        storage_folder_path, message.message_id + "_" + part["filename"]
+                    )
+                    os.makedirs(os.path.dirname(save_path), exist_ok=True)
+                    with open(save_path, "wb") as f:
+                        f.write(file_data)
+
+                    response["file_path"] = os.path.abspath(save_path)
+
+        yield response
```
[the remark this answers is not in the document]
AstrakhantsevaAA: because it was planned as gmail resource which retrieves all messages and attachments, but now I see, as you said above I should create resource messages and resource-transformer attachments

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+                res = parse_message(message, download, storage_folder_path)
+                yield from res
+
+        page_token = messages_info.get("nextPageToken")
+
+        if not page_token:
+            break
+
+
+def parse_message(
```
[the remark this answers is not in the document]
AstrakhantsevaAA: now it returns all messages and contents, I'll change it

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+                        filename = part.get_filename()
+                        if filename:
+                            attachment_data = part.get_payload(decode=True)
+                            attachment_path = os.path.join(storage_folder_path, message_uid.decode("utf-8")  + filename)
+                            os.makedirs(os.path.dirname(attachment_path), exist_ok=True)
+
+                            with open(attachment_path, 'wb') as f:
+                                f.write(attachment_data)
+
+                            item.update({"file_name": filename, "file_path": os.path.abspath(attachment_path)})
```
rudolfix, line 109: clone item first with copy(). why? because you are yielding and then overwriting the file name. and this is still the same dict - if you have many attachments per email
also please add content type!

rudolfix (changes requested):
this is really good! and dealing with emails is not easy. also I had no idea that imap is built in in Python :D

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def read_messages(client_: imaplib.IMAP4_SSL, criteria_: Sequence[Any]):
+        _, messages = client_.search(None, *criteria_)
+        message_ids = messages[0].split()
+
+        if not message_ids:
+            logger.warning("No emails found.")
+
+        for message_id in message_ids:
+            _, data = client_.fetch(message_id, "(RFC822)")
```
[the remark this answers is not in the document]
rudolfix: OK here people say how to search for UIDs past last value UID and also make search to return UIDs not message sequence numbers:
https://stackoverflow.com/questions/46453569/python-imaplib-get-uids-greater-than

FYI: there are two message ids in IMAP. only UID makes sense in our case

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+            raw_email = data[0][1]
+            msg = email.message_from_bytes(raw_email)
+            email_data = {
+                'message_uid': message_id,
+                'message_id': msg['Message-ID'],
+                'from': msg['From'],
+                'subject': msg['Subject'],
+                'date': pendulum.parse(msg['Date'], strict=False),
+                'content_type': msg.get_content_type(),
+                'body': get_email_body(msg),
```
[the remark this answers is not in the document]
AstrakhantsevaAA: so much information that is not human readable:D
Screenshot from 2023-07-27 11-32-55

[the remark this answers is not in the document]
AstrakhantsevaAA: image

sources/unstructured_data/inbox/__init__.py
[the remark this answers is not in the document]
AstrakhantsevaAA: result, data = client_.uid('search', 'UID ' + str(int(message_ids[0])) + ':*')

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+
+@dlt.resource
+def messages_uids(
+    host: str = dlt.secrets.value,
+    email_account: str = dlt.secrets.value,
+    password: str = dlt.secrets.value,
+    filter_emails: Sequence[str] = FILTER_EMAILS,
+    folder: str = "INBOX",
+    start_date: pendulum.DateTime = pendulum.datetime(2000, 1, 1),
+    initial_message_num: Optional[Any] = dlt.sources.incremental(
```
rudolfix, line 41: Optional[int]

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        "message_uid", initial_value=1
```
rudolfix, line 42: are getting the message with uid == 1? I think we have inclusive lower bound so you should

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        client.login(email_account, password)
+        client.select(folder, readonly=True)
+        criteria = [f'(SINCE {start_date.strftime("%d-%b-%Y")})']
+
+        if filter_emails:
+            logger.info(f"Load emails only from: {filter_emails}")
+            for email_ in filter_emails:
+                criteria.extend([f'(FROM {email_})'])
+
+        status, messages = client.uid('search', 'UID ' + str(int(last_message_num)) + ':*', *criteria)
```
rudolfix: hmmm just add it to `criteria` with SINCE?
`'UID ' + str(int(last_message_num)) + ':*'`

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+
+        status, data = client.uid('fetch', message_uid, '(RFC822)')
+
+        if status == 'OK':
+            raw_email = data[0]
+            if raw_email:
+                raw_email = data[0][1]
+                msg = email.message_from_bytes(raw_email)
+
+                email_data = dict(msg)
```
rudolfix: let's remove all properties starting with `X-` and move this code to a separate function

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+                email_data['message_uid'] = int(message_uid)
+                email_data["date"] = pendulum.parse(msg['Date'], strict=False)
```
rudolfix: it is `Date`

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if not message_uids:
+            logger.warning("No emails found.")
+        else:
+            yield from [{"message_uid": int(message_uid)} for message_uid in message_uids]
+
+
+@dlt.transformer(name="messages", write_disposition="replace")
+def read_messages(
+    item: TDataItem,
```
rudolfix, line 72: it is `int`

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+                email_data['message_uid'] = int(message_uid)
+                email_data["date"] = pendulum.parse(msg['Date'], strict=False)
+                email_data['content_type'] = msg.get_content_type()
+                email_data['body'] = get_email_body(msg)
+                yield email_data
+
+
+@dlt.transformer(name="attachments", write_disposition="replace")
+def get_attachments_by_uid(
+    item: TDataItem,
```
rudolfix, line 94: `int`

sources/unstructured_data/inbox/__init__.py
rudolfix: if that does not type check then keep the old type...

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+                        filename = part.get_filename()
+                        if filename:
+                            attachment_data = part.get_payload(decode=True)
+                            attachment_path = os.path.join(storage_folder_path, message_uid + filename)
+                            os.makedirs(os.path.dirname(attachment_path), exist_ok=True)
+
+                            with open(attachment_path, 'wb') as f:
+                                f.write(attachment_data)
+
+                            result = deepcopy(item)
```
rudolfix: since item is an `int` you do not need to deepcopy

rudolfix: extract all message details like in `read_messages`

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+                            result.update(
+                                {
+                                    "file_name": filename,
```
rudolfix: OK so this is our future standard to represent files :)
1. let's add `mtime` which is last modification time, you can take it from INTERNAL DATE
2. place the message information you get from `read_messages` in a field named `envelope`

yeah the things is getting organized

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+            raw_email = data[0]
+            if raw_email:
+                raw_email = data[0][1]
+                msg = email.message_from_bytes(raw_email)
+
+                email_data = dict(msg)
+                email_data['message_uid'] = int(message_uid)
+                email_data["date"] = pendulum.parse(msg['Date'], strict=False)
+                email_data['content_type'] = msg.get_content_type()
+                email_data['body'] = get_email_body(msg)
```
rudolfix: please also add internal date

rudolfix: add argument to fetch body or not. default is false

sources/unstructured_data/inbox/settings.py
```diff
@@ -0,0 +1,2 @@
+STORAGE_FOLDER_PATH = "temp"
```
rudolfix, line 3: use `tempfile` to get temp folder

rudolfix (changes requested):
a lot of small comments. but the overall structure is 👍

sources/unstructured_data/inbox/__init__.py
AstrakhantsevaAA: it is the lowest, I cannot take 0 or -1 it raises errors

sources/unstructured_data/inbox/__init__.py
AstrakhantsevaAA: mmm, I'm not sure it is gonna work.. how it can be int? it is transformer function

sources/unstructured_data/inbox/__init__.py
AstrakhantsevaAA: we can use it like  ` uids | read_messages | get_attachments_by_uid(storage_folder_path=storage_folder_path)` :D

AstrakhantsevaAA: changed

sources/unstructured_data/inbox/settings.py
AstrakhantsevaAA: i can use it in unstructured_data_pipeline.py script

AstrakhantsevaAA: I use it in tests

sources/unstructured_data/inbox/__init__.py
rudolfix: yes - now you yield dictionaries. before you yielded integers. (or I'm completely wrong)

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+            ]
+
+
+@dlt.transformer(name="messages", write_disposition="replace")
+def read_messages(
+    item: TDataItem,
+    host: str = dlt.secrets.value,
+    email_account: str = dlt.secrets.value,
+    password: str = dlt.secrets.value,
+    body: bool = False,
```
rudolfix, line 90: `include_body` sounds better

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+) -> TDataItem:
+    message_uid = str(item["message_uid"])
+
+    with imaplib.IMAP4_SSL(host) as client:
+        client.login(email_account, password)
+        msg = get_message_obj(client, message_uid)
+        if msg:
+            result = deepcopy(item)
+            result.update(extract_email_info(msg, body=body))
+            result.update({"internal_date": get_internal_data(client, message_uid)})
```
rudolfix, line 100: `get_internal_date`

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+                        attachment_path = os.path.join(
+                            storage_folder_path, message_uid + filename
+                        )
+                        os.makedirs(os.path.dirname(attachment_path), exist_ok=True)
+
+                        with open(attachment_path, "wb") as f:
+                            f.write(attachment_data)
+
+                        result = deepcopy(item)
+                        result.update(email_info)
```
rudolfix, line 137: maybe you could put it into `envelope` field not to mix attachment data with message (envelope in this context)

sources/unstructured_data/inbox/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+                        result.update(
+                            {
+                                "file_name": filename,
+                                "file_path": os.path.abspath(attachment_path),
+                                "content_type": part.get_content_type(),
+                                "internal_date": internal_date,
```
rudolfix, line 143: hmmm... I'd name it `modification_time` this is what you get for a file

sources/unstructured_data/inbox/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+        if raw_email:
+            raw_email = data[0][1]
+            msg = email.message_from_bytes(raw_email)
+
+    return msg
+
+
+def get_internal_data(client: imaplib.IMAP4_SSL, message_uid: str) -> Optional[Any]:
+    client.select()
+    status, data = client.uid("fetch", message_uid, "(INTERNALDATE)")
```
rudolfix, line 37: so INTERNALDATE is not included in the message? here we do a roundtrip to server IMO (RFC822 INTERNALDATE) will work

sources/unstructured_data/inbox/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    date = None
+    if status == "OK":
+        str_datetime = data[0].decode().split('"')[1]
```
rudolfix, line 40: https://docs.python.org/3/library/imaplib.html#imaplib.Internaldate2tuple

parse with this but then convert to pendulum datetime (this will create a local datetime without timezone)

rudolfix (changes requested):

sources/unstructured_data/inbox/__init__.py
AstrakhantsevaAA: only attachment part? or whole info (don't think it makes sense)?

Issue #224, opened 2023-07-23
Mypy can't type check sources in pipeline example scripts


steinitzu:
**Source name**
All sources

**Make sure that**
* [x] I checked the README in the source folder
* [] If I have problems with the credentials I checked the [credentials docs](https://dlthub.com/docs/general-usage/credentials) and [troubleshooting](https://dlthub.com/docs/walkthroughs/run-a-pipeline#4-detect-and-handle-problems)

**Describe the bug**

The pipeline scripts (e.g. [zendesk_pipeline.py](https://github.com/dlt-hub/verified-sources/blob/master/sources/zendesk_pipeline.py) ) are not fully type checked because when mypy is run from the root dir it can't resolve imports from the respective source package:

```python3
from zendesk import zendesk_chat
```
Correct import in this context would be:

```python3
from .zendesk import zendesk_chat
# or
from sources.zendesk import zendesk_chat
```

(former import is still needed and correct after generating the pipeline with `dlt init`)

Result is that mypy has no type information for sources in this script so errors from wrong source arguments and such are ignored.

**To Reproduce**
Clone the repo and install and check out this commit which has type errors, run from repo root:

```shell
git checkout 95b8a58
mypy sources/zendesk_pipeline.py
```
Mypy shows no errors

**Expected behavior**
Mypy should show multiple errors. Can verify this by running mypy from the sources dir instead:
```shell
cd sources
mypy zendesk_pipeline.py
```
-> 
```shell
zendesk_pipeline.py:70: error: Unexpected keyword argument "incremental_start_time" for "zendesk_support"  [call-arg]
zendesk_pipeline.py:71: error: Unexpected keyword argument "incremental_start_time" for "zendesk_chat"  [call-arg]
zendesk_pipeline.py:72: error: Unexpected keyword argument "incremental_start_time" for "zendesk_talk"  [call-arg]
zendesk_pipeline.py:99: error: Unexpected keyword argument "incremental_start_time" for "zendesk_support"  [call-arg]
zendesk_pipeline.py:99: error: Unexpected keyword argument "incremental_end_time" for "zendesk_support"  [call-arg]
zendesk_pipeline.py:107: error: Unexpected keyword argument "incremental_start_time" for "zendesk_support"  [call-arg]
Found 6 errors in 1 file (checked 1 source file)
```
**Possible solutions**
We may be able to simply delete `sources/__init__.py` so mypy doesn't read it as a package. Check whether that has any side effects, but I think that should be fine.

PR #230, opened 2023-07-28
Improve type checking in sources

steinitzu:
# Tell us what you do here

# Relevant issue

Related to https://github.com/dlt-hub/verified-sources/issues/224

# More PR info

* `sources/__init__.py` file is deleted, don't think it's needed -> mypy can now follow imports in pipeline scripts and check them properly.   
* Disable `ignore_missing_imports` by default. Enabling it globally silently ignores any misspelled module names and such -> instead ignore only the specific packages which are missing type hints.

### DIFF
--- a/check-package.sh
+++ b/check-package.sh
@@ -19,7 +19,7 @@ while IFS= read -r d; do
       error="yes"
     fi
   fi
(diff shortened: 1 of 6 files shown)


rudolfix (approved):
this is so good :)

Issue #212, opened 2023-07-14
Mongodb verified source


adrianbr:
# Quick source info
- Name of the source: Mongo
- What is the data source: Mongo db

# Current Status
- [ x] I plan to write it

# What source does/will do
enable copying mongo to dbs easily
unpack depth should be configurable

# Test account / test data
- [x ] I'd like `dltHub` to create a test account before we start

# Additional context
Anything we should know about your source

PR #239, opened 2023-08-10
Add mongodb source (WIP)

sehnem:
# Tell us what you do here

# Relevant issue

Issue #212

# More PR info

Added a Mongo verified Source as it was suggested by @adrianbr , most of the things are working, just `apply_hints` is not inserting the incremental for some reason, it also need tests that were not implemented yet.
As it is my first verified source it would be good if someone give me some feedback about possible issues in the current implementation, and, if it is simple, what I need for the `apply_hints` to work.
The implementation is based on the sql_database.


sources/mongodb_pipeline.py
```diff
@@ -0,0 +1,95 @@
...
+    analytics = mongodb(connection_url, database=database).with_resources("transactions", "accounts")
+    info = pipeline.run(analytics, write_disposition="replace")
+    print(info)
+
+    database = "sample_airbnb"
+    # Load a table incrementally with append write disposition
+    # this is good when a table only has new rows inserted, but not updated
+    airbnb = mongodb(connection_url, database=database).with_resources("listingsAndReviews")
+    # TODO: Not working
+    airbnb.listingsAndReviews.apply_hints(incremental=dlt.sources.incremental("last_scraped"))
```
rudolfix, line 41: look at sql_database example
```
from typing import List

import dlt
from dlt.sources.credentials import ConnectionStringCredentials
from dlt.common.time import ensure_pendulum_datetime
from dlt.common import pendulum

from sql_database import sql_database, sql_table


def load_select_tables_from_database() -> None:
    """Use the sql_database source to reflect an entire database schema and load select tables from it.

    This example sources data from the public Rfam MySQL database.
    """
    # Create a pipeline
    pipeline = dlt.pipeline(
        pipeline_name="rfam", destination="postgres", dataset_name="rfam_data", full_refresh=True
    )

    # Credentials for the sample database.
    # Note: It is recommended to configure credentials in `.dlt/secrets.toml` under `sources.sql_database.credentials`
    credentials = ConnectionStringCredentials(
        "mysql+pymysql://rfamro@mysql-rfam-public.ebi.ac.uk:4497/Rfam"
    )

    # Configure the source to load a few select tables incrementally
    source_1 = sql_database(credentials).with_resources("clan")
    # Add incremental config to the resources. "updated" is a timestamp column in these tables that gets used as a cursor
    # source_1.family.apply_hints(incremental=dlt.sources.incremental("updated"))
    source_1.clan.apply_hints(incremental=dlt.sources.incremental("updated", initial_value=ensure_pendulum_datetime("2014-06-02 15:19:28+00:00")))

    # Run the pipeline. The merge write disposition merges existing rows in the destination by primary key
    info = pipeline.run(source_1, write_disposition="append")
    print(info)
```
this works here: pulls just record after the date in `initial_value`

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+        )(client, collection, incremental=incremental)
+
+
+@dlt.common.configuration.with_config(
+    sections=("sources", "mongodb"), spec=MongoDbCollectionResourceConfiguration
+)
+def mongodb_collection(
+    connection_url: str = dlt.secrets.value,
+    database: str = dlt.config.value,
+    collection: str = dlt.config.value,
```
rudolfix, line 62: we should also add `write_disposition` here. if `sql_database` does not have it we'll add it as well

sources/mongodb/helpers.py
```diff
@@ -0,0 +1,113 @@
...
+        """Clean document to be json serializable"""
+        output = {}
+        if not isinstance(document, dict):
+            return document
+        for key, value in document.items():
+            if isinstance(value, dict):
+                output[key] = self.clean_document(value)
+            elif isinstance(value, list):
+                output[key] = [self.clean_document(item) for item in value]
+            elif isinstance(value, datetime):
```
rudolfix, line 62: does it always have timezone? if not we should ensure that timezone is present (we have util function for that)

sources/mongodb/helpers.py
```diff
@@ -0,0 +1,113 @@
...
+                output[key] = value
+            else:
+                output[key] = str(value)
```
rudolfix, line 65: hmmm why? we are losing types here. dlt will create schema from common python types so they should stay (including bytes if mongo supports binary blobs).
some problems with bson having their own types?

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    MongoDbCollectionConfiguration,
+    MongoDbCollectionResourceConfiguration,
+)
+
+@dlt.source
+def mongodb(
+    connection_url: str = dlt.secrets.value,
+    database: str = dlt.config.value,
+    collection_names: Optional[List[str]] = dlt.config.value,
+    incremental: Optional[dlt.sources.incremental[Any]] = dlt.config.value,
```
rudolfix, line 20: you do not need this at source level. people will configure that per table

sources/mongodb_pipeline.py
rudolfix: IDK what "not working" means... most common problems:
- defining the cursor on non existing column (will raise exception with explanation)
- using `initial_value` with type incompatible with what is in cursor column - exception on musing max function

if by not working you mean that it is not filtering data then let's talk on slack

rudolfix (changes requested):
hey! this is good stuff! thanks for all docstrings and documentation. also the code is clear and well organized.

I raised a few issues and had some questions

the thing that is still missing are tests. here, before you starts, we'll create a mongo instance for you where you could put your test data - then we'll pass this to our CI

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    Returns:
+        Iterable[DltResource]: A list of DLT resources for each collection to be loaded.
+    """
+    client = client_from_credentials(connection_url)
+    database = client[database]
+
+    collection_obj = database[collection]
+
+    return dlt.resource(
+        collection_documents, name=collection_obj.name, primary_key="_id"
```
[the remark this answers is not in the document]
sehnem: Yes, it is always `_id`, you can customize it but it will transform the field into an object.

sources/mongodb/helpers.py
sehnem: I set an option `tz_aware=True` in the client that ensure that we have timezone.

sources/mongodb/__init__.py
sehnem: It will never run on source level? I based it on the sql one, should I remove it? Sorry I am not sure if I understood this part.

sources/mongodb_pipeline.py
sehnem: I found that the incremental is applied with the `apply_hints` but when extracting the data, the incremental is removed by the function [`apply_hint_args`](https://github.com/dlt-hub/dlt/blob/devel/dlt/pipeline/pipeline.py#L818) from dlt. I could not figure out yet if I am passing something wrong or what, will keep looking at it.

rudolfix: Let me also check. maybe you've found some kind of a bug

tests/mongodb/test_mongodb_source.py
```diff
@@ -0,0 +1,30 @@
...
+    print(load_info)
+    assert_load_info(load_info)
+    table_names = [t["name"] for t in pipeline.default_schema.data_tables()]
+    table_counts = load_table_counts(pipeline, *table_names)
+
+    # for now only check main tables
+    expected_tables = ["collection1", "collection2"]
+    assert set(table_counts.keys()) >= set(expected_tables)
+
+    assert table_counts["collection1"] == 20
```
rudolfix, line 25: how does this test work? do we need to use mongo mock to create collections being tested? I do not see that here

sources/mongodb/helpers.py
```diff
@@ -0,0 +1,122 @@
...
+        for key, value in document.items():
+            if isinstance(value, dict):
+                output[key] = self.clean_document(value)
+            elif isinstance(value, list):
+                output[key] = [self.clean_document(item) for item in value]
+            elif isinstance(value, Decimal128):
+                output[key] = value.to_decimal()
+            elif isinstance(value, ObjectId):
+                output[key] = str(value)
+            elif isinstance(value, (str, int, datetime, float)):
```
rudolfix, line 65: those I'll put at the top of the checks - most of the values are of those types

sources/mongodb/helpers.py
```diff
@@ -0,0 +1,122 @@
...
+        """Clean document to be json serializable"""
+        output = {}
+        if not isinstance(document, dict):
+            return document
+        for key, value in document.items():
+            if isinstance(value, dict):
+                output[key] = self.clean_document(value)
+            elif isinstance(value, list):
+                output[key] = [self.clean_document(item) for item in value]
+            elif isinstance(value, Decimal128):
```
rudolfix, line 61: I hope there are no more decimals like decimal256 etc. used by mongo? maybe there's just generic type decimal in bson?

tests/mongodb/test_mongodb_source.py
rudolfix: when we create mock collections let's make sure all data types (including dates, decimals, nested jsons are there)

rudolfix (changes requested):
this looks really good! thanks! just small issues to existing code. what we are missing is more tests and some setup:

1. there are merge conflicts, could you rebase/merge the master branch into yours?
2. we should use your demo script as test cases - you can just copy the code or call those functions and then assert that tests are passing
3. we are working hard for the CI to run correctly from forks - we'll need to add some test mongo instances to our secrets in order to run tests. you are using some hosted mongo right? with airbnb test data?
4. is incremental load working now? what was wrong before?

sources/mongodb/helpers.py
sehnem: I checked it, and `Decimal128` is the only decimal in [bson](https://github.com/mongodb/mongo-python-driver/blob/master/bson/__init__.py#L136).

pyproject.toml
```diff
@@ -27,7 +27,9 @@ types-requests = "^2.28.11.7"
 mimesis = "^7.0.0"
 requests-mock = "^1.10.0"
 pandas-stubs = "^2.0.0"
-black = "^23.3.0"
```
rudolfix, line 30: was there a merge conflict? I think that
- you removed black
- the [tool.black] must be at the end. this is poetry/tomlkit bug and we can't mix sections...

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    MongoDbCollectionConfiguration,
+    MongoDbCollectionResourceConfiguration,
+    client_from_credentials,
+    collection_documents,
+)
+
+
+@dlt.source
+def mongodb(
+    connection_url: str = dlt.secrets.value,
```
rudolfix, line 18: what if database is part of `connection_url`? do we still need `database` as string? if so we could make optional

sources/mongodb_pipeline.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Run the pipeline. For a large db this may take a while
+    info = pipeline.run(source, write_disposition="replace")
+
+    return info
+
+
+if __name__ == "__main__":
+    # Credentials for the sample database.
+    # Note: It is recommended to configure credentials in `.dlt/secrets.toml` under `sources.sql_database.credentials`
```
rudolfix, line 114: here it is 
```toml
[sources.mongodb]
connection_url=...
```
or just connection_url="..."

tests/mongodb/test_mongodb_source.py
```diff
@@ -0,0 +1,46 @@
...
+    pipeline = dlt.pipeline(
+        pipeline_name="mongodb",
+        destination=destination_name,
+        dataset_name="mongodb_data",
+        full_refresh=True,
+    )
+
+    connection_url = "mongodb://admin:123456@localhost:27017"
+    # Load selected collection with different settings
+    load_info = load_select_collection_db(connection_url)
```
rudolfix, line 22: ohh this is clever, we must start running our demo scripts in tests!

rudolfix (changes requested):
this is really good! the only important issue is to fix pyproject.toml (probably merge conflict). thanks!

sources/mongodb/__init__.py
sehnem: Just gave a look at the [docs](https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html) to confirm, apparently you cannot pass database with the connection string.

sources/mongodb/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Setup verified source
+
+To setup the Mongo Database Verified Source read the [full documentation here.](https://dlthub.com/docs/dlt-ecosystem/verified-sources/mongodb)
+
+## Add credentials
+1. Open `.dlt/secrets.toml`.
+2. In order to continue, you can use the a connection URL and database that looks like this:
+    ```bash
+    connection_url = "mongodb://dbuser:passwd@host.or.ip:27017"
```
rudolfix, line 18: Please link to some documentation on supported connection strings

sources/mongodb/README.md
```diff
@@ -0,0 +1,54 @@
...
+    database = "local"
```
rudolfix, line 19: explain that database is optional and if one is provided in connection string - that one will be used

sources/mongodb/README.md
```diff
@@ -0,0 +1,54 @@
...
+    ```bash
+    python3 mongodb_pipeline.py
+    ```
+    
+3. To ensure that everything loads as expected, use the command:
+    ```bash
+    dlt pipeline <pipeline_name> show
+    ```
+    
+    For example, the pipeline_name for the above pipeline example is `local_mongo`, you can use any custom name instead.
```
rudolfix, line 49: you named the pipelines differently than `loacal_mongo` in the demo scripts

sources/mongodb/README.md
```diff
@@ -0,0 +1,54 @@
...
+    
+
```
rudolfix, line 51: You could add another chapter ie
## How to use

and here you explain that by default we reflect all collections in database, we add primary keys and you can select the collections before loading as demonstrated in demo script

or you can load single collection by calling ... function

we'll improve this doc anyway but this is essential information

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    incremental: Optional[dlt.sources.incremental] = None,  # type: ignore[type-arg]
+    write_disposition: Optional[str] = dlt.config.value,
+) -> Iterable[DltResource]:
+    """
+    A DLT source which loads data from a mongo database using PyMongo.
+    Resources are automatically created for each collection in the database or from the given list of collection.
+
+    Args:
+        connection_url (str): Database connection_url.
+        database (str): Selected database name.
```
rudolfix, line 30: explain that it is optional and taken from connection url if not present

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+        collection (Optional[List[str]]): The list of collections `pymongo.collection.Collection` to load.
```
rudolfix, line 31: `collection`?
please pay attention to docstrings

sources/mongodb/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    collection: str = dlt.config.value,
+    incremental: Optional[dlt.sources.incremental] = None,  # type: ignore[type-arg]
+    write_disposition: Optional[str] = dlt.config.value,
+) -> Any:
+    """
+    A DLT source which loads a collection from a mongo database using PyMongo.
+
+    Args:
+        connection_url (str): Database connection_url.
+        database (str): Selected database name.
```
rudolfix, line 77: same

sources/mongodb/helpers.py
```diff
@@ -0,0 +1,115 @@
...
+import dlt
+from bson.decimal128 import Decimal128
+from bson.objectid import ObjectId
+from dlt.common.configuration.specs import BaseConfiguration, configspec
+from dlt.common.typing import TDataItem
+from pymongo import MongoClient
+
+
+CHUNK_SIZE = 10000
+
```
rudolfix, line 15: let's try to type client, collection and cursor anyway by doing that only when typechecking

from typing import TYPE_CHECKING
if TYPE_CHECKING:
    TMongoClient = MongoClient[Any]
    TCollection = Collection[Any]  # type: ignore
    TCursor = Cursor[Any]
else:
    TMongoClient = Any
    TCollection = Any
    TCursor = Any

and use TMongoClient, TCollection and TCursor below. this should work!

sources/mongodb/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pymongo>=4.4.1
```
rudolfix, line 1: I think a version from like 6 months ago also will work? does not make sense to put the newest dep

sources/mongodb_pipeline.py
```diff
@@ -0,0 +1,109 @@
+from typing import List
+
+import dlt
+from dlt.common import pendulum
+from dlt.pipeline.pipeline import Pipeline
+from dlt.common.pipeline import LoadInfo
+
+try:
+    from .mongodb import mongodb, mongodb_collection  # type: ignore
```
rudolfix, line 9: please explain the trick in a comment: that we want to make this script to work as part of the tests and standalone so we try two import methods. first being one used for the tests

sources/mongodb_pipeline.py
```diff
@@ -0,0 +1,109 @@
...
+
+def load_select_collection_db(pipeline: Pipeline = None) -> LoadInfo:
+    """Use the mongodb source to reflect an entire database schema and load select tables from it.
+
+    This example sources data from a sample mongo database data from [mongodb-sample-dataset](https://github.com/neelabalan/mongodb-sample-dataset).
+    """
+    if pipeline is None:
+        # Create a pipeline
+        pipeline = dlt.pipeline(
+            pipeline_name="pipeline1", destination="duckdb", dataset_name="mongo_data"
```
rudolfix, line 22: maybe you can name pipeline as in docs?

rudolfix: you can use the same pipeline name everywhere

but set the dataset name so it corresponds the data being loaded (ie. collections, databases)

rudolfix (changes requested):

rudolfix (approved):

PR #240, opened 2023-08-10
Bug fix facebook ads.

zem360:
# Tell us what you do here
- [x] fixing a bug 

# More PR info
The helper function **get_start_date()** is using the **parse_iso_like_datetime()** function to convert the ISO datetime string into pendulum.datetime object. The **parse_iso_like_datetime()** function can not handle ISO date string and returns an datetime.date object instead of a pendulum.datetime object.

I replace the **parse_iso_like_datetime()** with **ensure_pendulum_datetime** function inside the  **get_start_date()**. I also wrote some test to ensure that pendulum.datetime is returned.

### DIFF
--- a/sources/facebook_ads/__init__.py
+++ b/sources/facebook_ads/__init__.py
@@ -10,7 +10,7 @@ from facebook_business.api import FacebookResponse
 import dlt
 from dlt.common import pendulum, logger
 from dlt.common.typing import TDataItems, TDataItem, DictStrAny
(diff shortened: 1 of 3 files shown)


rudolfix (changes requested):
this looks good and thx for the test! just one thing: could you resubmit this PR to our master branch in verified sources repo? don't use fork because CI does not see the credentials

rudolfix:
@zem360 you can close this

PR #244, opened 2023-08-14
Facebook ads datetime object conversion fix.

zem360:
# Tell us what you do here

# More PR info
The helper function **get_start_date()** is using the **parse_iso_like_datetime()** function to convert the ISO datetime string into pendulum.datetime object. The **parse_iso_like_datetime()** function can not handle ISO date string and returns an datetime.date object instead of a pendulum.datetime object.

I replace the **parse_iso_like_datetime()** with **ensure_pendulum_datetime** function inside the  **get_start_date()**. I also wrote some test to ensure that pendulum.datetime is returned.

### DIFF
--- a/sources/facebook_ads/__init__.py
+++ b/sources/facebook_ads/__init__.py
@@ -186,9 +186,6 @@ def facebook_insights_source(
             query = {
                 "level": level,
                 "action_breakdowns": list(action_breakdowns),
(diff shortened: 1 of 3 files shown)


zem360:
I have removed the breakdown parameter.

rudolfix (approved):

Issue #251, opened 2023-08-28
Slack verified source


adrianbr:
# Quick source info
- Name of the source: Slack
- What is the data source: We prefer using the REST api to avoid adding more dependencies (as opposed to their python wrapper)

# Current Status
- to be done

# What source does/will do
- We want to get info about the conversations and engagement. To do so, we will read
- conversations in all public channels https://slack.com/api/conversations.list
- Reactions - Maybe Reactions get or reactions list https://api.slack.com/methods/reactions.get
- access logs https://api.slack.com/methods/team.accessLogs

# Test account / test data
- I'd like `dltHub` to create a test account before we start

# Additional context
Anything we should know about your source
- Make sure the endpoints selected work for free slack tier!
- we would like to load incrementally wherever the api allows incremental extraction

sehnem:
Hey @adrianbr there are a few things about the slack integration.

- I was able to implement the conversations.list endpoint, there is just no filter for incrementally getting the data.
- For the reactions.get endpoint there is an issue, probably the list of all messages will have to be implemented. I created the app as a bot, so it need to be added to the channels before getting the messages, probably it should be done as an admin account, right?
- The team.accessLogs has an issue that it is just available for the admin scope, waiting for one account with this permissions.

PR #253, opened 2023-08-30
add slack source

sehnem:
# Tell us what you do here

# Relevant issue

issue #251

# More PR info

Add slack source with access_logs and conversations.


sources/slack/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+from .helpers import SlackApi
+from .settings import MAX_PAGE_SIZE
+
+
+@dlt.source(name="slack")
+def slack_source(
+    access_token: str = dlt.secrets.value,
+    page_size: int = MAX_PAGE_SIZE,
+    created_before: Optional[TAnyDateTime] = None,
```
rudolfix, line 18: should we support same interface we have in zendesk and shopify? so you can specify both start and end range. and use incremental for incremental loading, not replace
btw. we have a new quick with fully working incremental snippet
https://dlthub.com/docs/getting-started

sources/slack/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+    def conversations() -> Iterable[TDataItem]:
+        """
+        The resource for conversations on your Slack channel, supports pagination.
+
+        Returns:
+            Iterable[TDataItem]: A generator of products.
+        """
+        conversations = SlackApi(
+            access_token=access_token,
+            resource="conversations.list",
```
rudolfix, line 49: this only gets list of channels, not messages

rudolfix (changes requested):
please add convestation.history.
- allow for a list of channel ids or if not provided use conversation.list to get channels
- implement incremental with standard behavior: see zendesk/shopify
- route differetn event types to different tables: see this example (https://deploy-preview-568--dlt-hub-docs.netlify.app/docs/getting-started/#dispatch-stream-of-events-to-tables-by-event-type)

sources/slack/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+        access_token=access_token,
+        resource="conversations.list",
+        response_path="channels",
+        datetime_fields=["updated", "created"],
+        page_size=page_size,
+    )
+
+    yield conversations
+
+    yield conversations | dlt.transformer(
```
rudolfix, line 90: OK this is really smart and could be used to create a DAG where we learn about channels at execution time and create tables for each channels. in case of slack this is more complicated though. I think we are better off if we follow the mongodb pattern:
- get a list of channels by evaluating list of conversations
- for each conversation (or if an explicit list of channel names/ids is present - filter only those on the list) create a dlt resource which will contain conversation history. give it meaningful name ie. the one you use below (but I think it is too long. maybe channel_{channel} is engough and please use channel name not id)
- and now the trick: channel history contains many different event types. route each event type to a separate tables! I sent you links to examples in previous review

so at the end:
- you have a list of resources for each conversation
- each resource routes data to many tables corresponding to event types

rudolfix (changes requested):
besides the review fixes:
- look at docs, docstrings and comments - you have many leftovers from shopify
- there's some code clearly added to test - pointed our in review. please make sure that code is lcear before pushing
- in general read your code again before pushing next time
thanks!

sources/slack/helpers.py
```diff
@@ -0,0 +1,163 @@
...
+            return item
+
+        for field in self.datetime_fields:
+            if field in item:
+                timestamp = item[field]
+                if isinstance(timestamp, str):
+                    timestamp = float(timestamp)
+                if timestamp > 1e10:
+                    timestamp = timestamp / 1000
+                item[field] = pendulum.from_timestamp(timestamp)
```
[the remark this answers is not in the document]
sehnem: Some of the timestamps are in milliseconds without decimal point and other in seconds and the `from_timestamp` method just supports seconds, this is the error I get when removing the conversion.
```
File "/home/josue/Documentos/dlthub/venv/lib/python3.10/site-packages/pendulum/__init__.py", line 272, in from_timestamp
dt = _datetime.datetime.utcfromtimestamp(timestamp)
ValueError: year 55286 is out of range
```

sources/slack/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+        name="conversations_history",
+        table_name=lambda i: f'conversations_history_{i["channel"]}',
+        write_disposition="merge",
+        primary_key="ts",
```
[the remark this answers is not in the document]
sehnem: No, if you want to fetch a specific message you need to use a combination of channel_id and ts.

sources/slack/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+            "ts",
+            initial_value=start_dt,
+            allow_external_schedulers=True,
+        ),
+    ) -> Iterable[TDataItem]:
+        start_date_ts = convert_dt(start_date.last_value, to_ts=True)
+        end_date_ts = convert_dt(end_date, to_ts=True)
+        params = {
+            "oldest": start_date_ts,
+            "latest": end_date_ts,
```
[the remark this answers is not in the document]
sehnem: Yes, that's why I am using this function, because I need to convert and check for Nones, so I do everything at it.

sources/slack/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+        start_date: dlt.sources.incremental[TAnyDateTime] = dlt.sources.incremental(
+            "ts",
+            initial_value=start_dt,
+            allow_external_schedulers=True,
+        ),
+    ) -> Iterable[TDataItem]:
+        start_date_ts = convert_dt(start_date.last_value, to_ts=True)
+        end_date_ts = convert_dt(end_date, to_ts=True)
+        params = {
+            "oldest": start_date_ts,
```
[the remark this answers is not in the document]
sehnem: They are actually datetimes, I convert them from the slack response to datetime and they are just converted back to timestamp for the request.

sources/slack/__init__.py
sehnem: The list of conversations is the list of channels, using the admin token, it could bring other conversations using a user token, from what I could find doing some tests.

For event types in the `conversation.history`, they are all messages, there is no separation of event, not sure how to break it. And this `client_msg_id` is not always present, endpoints that get reaction, for example get it based on ts and channel.

Sample history:

```
        {
            "client_msg_id": "afa46452-4716-4c40-938a-a1cf398ee503",
            "type": "message",
            "text": "hey, I've got a question regarding the new mongodb source, does it also work for documentdb?",
            "user": "U05A0SCG6KT",
            "ts": "1693837655.609649",
            "blocks": [
                {
                    "type": "rich_text",
                    "block_id": "ZaxQX",
                    "elements": [
                        {
                            "type": "rich_text_section",
                            "elements": [
                                {
                                    "type": "text",
                                    "text": "hey, I've got a question regarding the new mongodb source, does it also work for documentdb?"
                                }
                            ]
                        }
                    ]
                }
            ],
            "team": "T04DHMAF13Q",
            "thread_ts": "1693837655.609649",
            "reply_count": 5,
            "reply_users_count": 4,
            "latest_reply": "1693844421.520359",
            "reply_users": [
                "U054H6GT3BP",
                "U05LJG8UV2R",
                "U05A0SCG6KT",
                "U04DQ9VUN1Y"
            ],
            "is_locked": false,
            "subscribed": false
        },
```

sources/slack/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+            return None
+        out_dt = ensure_pendulum_datetime(dt)
+        if to_ts:
+            return out_dt.timestamp()
+        return out_dt
+
+    end_dt: Optional[DateTime] = convert_dt(end_date)
+    start_dt: Optional[DateTime] = convert_dt(start_date)
+
+    @dlt.transformer(
```
AstrakhantsevaAA, line 54: and i would move this function out of source function. it will make the code more readable

sources/slack/helpers.py
```diff
@@ -0,0 +1,157 @@
...
+
+import pendulum
+from dlt.common.time import ensure_pendulum_datetime
+from dlt.common.typing import Dict, TDataItems
+from dlt.sources.helpers import requests
+
+from .settings import MAX_PAGE_SIZE, SLACK_API_URL
+
+
+class SlackApi:
```
AstrakhantsevaAA, line 13: this is not the best way to work with classes. 
- please rename it to SlackAPI (just my whim) =)
- it makes sense to create an object of the class only once. when creating a source, not a resource. I would remove all these variables from the init, and I would make all auxiliary functions static.
        self.resource = resource
        self.response_path = response_path
        self.params = params or {}
        self.datetime_fields = datetime_fields or []
        self.page_size = page_size
- it would be more logical to pass these variables to the function get_pages()

sources/slack/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+        resource="conversations.list",
+        response_path="$.channels[*]",
+        datetime_fields=["updated", "created"],
+    ):
+        channels.extend(page_data)
+
+    @dlt.resource(name="channels", primary_key="id", write_disposition="replace")
+    def channels_resource() -> Iterable[TDataItem]:
+        """Yield all channels as a DLT resource."""
+        yield from channels
```
AstrakhantsevaAA, line 56: why don't you put pagination code under the resource `channels_resource` function and don't use `add_filter` for selected_channels?

AstrakhantsevaAA: then you can use channels with transformer as earlier. why did you changed it?

rudolfix (changes requested):
OK this looks good! see the review
Please update docs - info on conversation history, selecting channels to load and how tables are named

rudolfix:
@sehnem please change the directory name where the test are. needs to be the same as source name so `slack`. otherwise tests are not selected!

sources/slack/helpers.py
```diff
@@ -0,0 +1,206 @@
...
+        next_cursor = None
+
+        # Iterate through all pages
+        while has_next_page:
+            # Make the request
+            response = requests.get(
+                url=self.url(resource),
+                headers=self.headers,
+                params=self.parameters(params or {}, next_cursor),
+                timeout=60,
```
rudolfix, line 180: please remove timeout. our requests have it and it can be also configured by the user

rudolfix (changes requested):
also please merge master into this branch to run tests.
this is really good now! thanks!

rudolfix (approved):
excellent! just CI must pass, we'll setup it now

AstrakhantsevaAA (approved):

Issue #216, opened 2023-07-17
[filesystem] verified source


willi-mueller:
# Quick source info
- Name of the source: **filesystem** in standard sources https://github.com/dlt-hub/verified-sources/issues/256
- What is the data source: `fsspec` supported file system: cloud object storage, S3, GCS, ADSL, local.

# Current Status
- [x] I plan to write it

# What source does/will do
This connector loads a series of files from an S3 or GCS bucket. Which files will be loaded depends on a configuration giving a file name pattern as well as the `write_disposition`.

## Why this is relevant
Many data providers deliver raw data or reports to bucket storage in S3 or GCS.

This connector can be the basis for additional readily configured source connectors that import from buckets. For example, [Adjust offers hourly free CSV export](https://help.adjust.com/en/article/csv-uploads) that can be configured by the Adjust customer. I plan to open a follow-up issue for an Adjust connector.

# Test account / test data
- [x] I'd like `dltHub` to create a test account before we start
- [ ] I have a test account and can share my test data or a test account with `dltHub` after we merge the source

# Additional context
This source, like other standard readers, will be used as a first step in data loading pipeline. In essence
1. it will read a specified folder and filter file names with specified regex or glob
2. it will yield FileItem(s) (file item data contract) - https://github.com/dlt-hub/verified-sources/issues/256
3. it will not read the files, it will not copy them locally!
4. it will provided a helper method that, when passed a file url, will return authenticated `fsspec` instance that behaves like file object - so next step in the pipeline can read the file - often piece by piece

All other common properties of standard source apply
- incremental load on `mtime` of a file (optional)
- primary key on file url

## `dlt` core code reuse
https://github.com/dlt-hub/dlt/pull/626 - introduces fsspec into common of dlt library. take things from there. also look at tests with usage examples

1. * [ ] we reuse filesystem configuration
2. * [ ] we reuse fsspec client factory
3. * [ ] we reuse mtime dispatch

## Requirements
1. * [ ] implement as `dlt.resource` which takes following arguments:

- `bucket_url` and `credentials` arguments as per ... spec
- `filename_filter` which can be compiled regex/glob expression or callback function
- a flag to enable/disable incremental load (disabled by default) + start/end date - like in slack/zendesk

3. * [ ] how it work internally

- it walks/globs the `bucket_url` and filters out files with regex/glob/callback
- yields file items in configurable pages

5. * [ ] implements following helper functions:

[...]

PR #259, opened 2023-09-04
init builtin

sehnem:
# Tell us what you do here

# Relevant issue

issue #225

# More PR info


rudolfix (changes requested):
looks good. this is a first time we build such thing so below are more like suggestions
- we should import all relevant sources and resources into the top level `__init__.py` and users should interact mostly with it
- let's write the top level documentation later. it should just explain how to pick one of the standard sources and setup credentials in general.
- what kind of credentials layout we prefer? my take would be 
```toml
[sources.inbox]
host="..."
```
(i think this is the default even in this structure)

sources/standard/inbox/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+                    else:
+                        attachments = chain(attachments, new_attachments)
+            email_info = extract_email_info(msg)
+
+            for attachment in attachments:
+                file_data = ImapSource(
+                    file_name=attachment["file_name"],
+                    storage_path=storage_path,
+                    file=attachment["payload"],
+                    mod_date=email_info["Date"],
```
rudolfix, line 181: we should never use `Date`. it may be fake, use internal timestamp as before.

sources/standard/inbox/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+                        attachments = new_attachments
+                    else:
+                        attachments = chain(attachments, new_attachments)
+            email_info = extract_email_info(msg)
+
+            for attachment in attachments:
+                file_data = ImapSource(
+                    file_name=attachment["file_name"],
+                    storage_path=storage_path,
+                    file=attachment["payload"],
```
rudolfix, line 180: IMO we should save files to a storage_path and pass the absolute url along. they user will decide what to do with the file later

rudolfix (changes requested):

sources/standard/inbox/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+        TDataItem: A dictionary containing the collected email information and attachment details.
+    """
+
+    with imaplib.IMAP4_SSL(host) as client:
+        client.login(email_account, password)
+        client.select()
+
+        for item in items:
+            message_uid = str(item["message_uid"])
+            msg = get_message(client, message_uid)
```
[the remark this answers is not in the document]
sehnem: I could not find any way, will give an other look at the imap protocol to see if I can find something.
But, as the body contains the whole email data, wouldn't it be better to get message by message and save the attachment storing just the metadata and so avoid using a lot of memory? Although email have a maximum of a few MBs so it would not hurt too much the memory.

sources/standard/inbox/__init__.py
sehnem: I am moving it back to the INTERNALDATE, I used it just to avoid doing an other request to the server just to get the date, but it looks like it is really a [different](https://www.metaspike.com/imap-internal-date-forensic-email-authentication/#:~:text=Internal%20Date%20is%20an%20IMAP,Let's%20start%20with%20an%20example.)

sources/standard/inbox/__init__.py
sehnem: Inside the class this is done, the class creates paths, generates the hash and save the file. But I will move that to the code.

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+from ..file_source import FileModel
+from .settings import DEFAULT_CHUNK_SIZE, DEFAULT_START_DATE, STORAGE_PATH
+
+from fsspec import AbstractFileSystem
+import pandas as pd
+
+
+@dlt.source
+def filesystem_source(
+    credentials: FilesystemConfiguration = dlt.secrets.value,
```
rudolfix, line 22: I'd like to have something more user friendly: just pass bucket_url and credentials like you see them `FilesystemConfiguration` as arguments and use `FilesystemConfiguration` as spec in `dlt.resource`. (you did that in Mondo).
in next step we'll subclass  `FilesystemConfiguration` and add `storage_path` `start_date` etc. to it

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    storage_path: str = STORAGE_PATH,
+    start_date: pendulum.DateTime = DEFAULT_START_DATE,
```
rudolfix, line 24: we can completely remove incremental loading from `filesystem`. users will be able to add those themselves via `filesystem.incremental` we can just document that.

we always read all matching files anyway - there's no way to filter by mtime so it is enough

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    filename_filter: Optional[str] = None,
+    chunksize: int = DEFAULT_CHUNK_SIZE,
+    extract_content: bool = False,
```
rudolfix, line 27: let's make it `load_content` and if set, they content of the file will be loaded into memory (`file_content` of FileModel) and yielded. we should load as bytes. opening string reader over bytes is trivial

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+
+    if not filename_filter:
+        filename_filter = "*"
+    files_url = os.path.join(bucket_url, filename_filter)
+    files = fs_client.glob(files_url, detail=True)
+
+    last_modified = modified_at.last_value
+
+    file_list = []
+    for file, md in files.items():
```
rudolfix, line 97: return full FileModel here!

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        yield file_md.dict()
+
+
+@dlt.transformer(
+    name="content",
+    write_disposition="merge",
+    merge_key=None,
+    primary_key=None,
+)
+def extract_files(
```
rudolfix, line 159: this is a part of the demo of the filesystem pipeline! and we have one problem here: there's no clear way to pass `fs_client` - you can't expect that instance will be automatically created.

my take is that we subclass Python built-in dict and add a special property that will hold `credentials`. we'll also implement a helper function that when passed such dict, will use `credentials` to get `AbstractFilesystem`. I wrote that in my ticket #216

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        ),
+    )
+
+    if extract_content:
+        return file_list | extract_files(fs_client=fs_client)
+    else:
+        return file_list | copy_files(fs_client=fs_client, storage_path=storage_path)
+
+
+@dlt.resource
```
rudolfix, line 64: name the resource `filesystem`

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+def list_remote_files(
+    fs_client: AbstractFileSystem,
+    bucket_url: str,
+    filename_filter: Optional[str] = None,
+    modified_at: dlt.sources.incremental[pendulum.DateTime] = dlt.sources.incremental(
```
rudolfix, line 69: and as mentioned, drop this but include in demo for the filesystem!

rudolfix (changes requested):
two biggest changes
1. just one `dlt.resource`, we do not need a source
2. see my comments how credentials are passed

overall there's progress! this will be really simple thing but with many demos and docs

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+    if not filename_filter:
+        filename_filter = "*"
+    files_url = os.path.join(credentials.bucket_url, filename_filter)
+
+    files_chunk = []
+
+    for file_model in get_files(fs_client, files_url):
+        file_dict: Dict[str, Any] = dict(**file_model)
+        if extract_content:
+            file_dict["content"] = fs_client.read_bytes(file_dict["file_url"])
```
rudolfix, line 49: make it part of the optional contract, name "file_content"

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        # Need to check if passing an open file is ok, probably we will need to create a class
+        # to manage the context, closing the file after the process.
+        else:
+            file_dict["file_instance"] = fs_client.open(file_dict["file_url"])
```
rudolfix, line 53: it is a bad idea. what we need to do (IMO) is to subclass `dict` and use custom dict in here dict(**file_model) ie
FilesystemDict(file_model, credentials) which behaves as normal dict but also has credentials from which you can recreate fsspec when needed, and then we yield this dict as below.

it should also implement:
1. property `filesystem` that returns `AbstractFilesystem`
2. method `open` with a signature of `open` of `AbstractFilesystem` that just passes arguments to it

and this is what you need to use in transformers to open and read files.

sources/standard/filesystem/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+from dlt.common.storages import filesystem_from_config
+from dlt.common.storages.configuration import FilesystemConfiguration
+from dlt.common.time import ensure_pendulum_datetime
+from fsspec import AbstractFileSystem  # type: ignore
+
+from ..file_source import FileModel
+
+
+@configspec
+class FilesystemConfigurationResource(FilesystemConfiguration):
```
rudolfix, line 17: good! and now use it!

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@dlt.resource(
+    name="filesystem",
+    merge_key="file_url",
+    primary_key="file_url",
+    spec=FilesystemConfiguration,
+)
+def filesystem_resource(
+    credentials: FilesystemConfiguration = dlt.secrets.value,
```
rudolfix, line 20: no, take a look at `FilesystemConfiguration`, what you need it:
bucket_url: str = dlt.config.value,
credentials: FileSystemCredentials = dlt.secrets.value

rudolfix (changes requested):
there's progress, let's talk tomorrow!

sources/standard_pipeline.py
```diff
@@ -0,0 +1,111 @@
...
+    """Reads files and copy them to local directory.
+
+    Args:
+        items (TDataItems): The list of files to copy.
+
+    Returns:
+        TDataItem: The file content
+    """
+    for file_obj in items:
+        df = pd.read_parquet(file_obj.open_fs())
```
rudolfix, line 58: heh what's this? OK it makes sense to show how to read parquet but I think you want to use pyarrow?

rudolfix (changes requested):
good set of tests and demos, but I think our csv stream reader is gone?

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+        opened_file: IOBase
+        # if the user has already extracted the content, we use it so there will be no need to
+        # download the file again.
+        if self["file_content"] in self:
+            opened_file = BytesIO(self["file_content"])
+        else:
+            opened_file = self.filesystem.open(self["file_url"])
+        return opened_file
+
+    def read(self) -> bytes:
```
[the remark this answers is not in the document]
sehnem: by default it opens as "rb", I am including the parameter in the function call to make it explicit.

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+    @property
+    def filesystem(self) -> AbstractFileSystem:
+        """The filesystem client based on the given credentials.
+
+        Returns:
+            AbstractFileSystem: The filesystem client.
+        """
+        return client_from_credentials(self["file_url"], self.credentials)
+
+    def open_fs(self) -> IOBase:
```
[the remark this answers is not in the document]
sehnem: fsspec open enforces the bytes mode, the other modes are just write

```
        if "b" not in mode:
            mode = mode.replace("t", "") + "b"

            text_kwargs = {
                k: kwargs.pop(k)
                for k in ["encoding", "errors", "newline"]
                if k in kwargs
            }
            return io.TextIOWrapper(
                self.open(
```

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    primary_key="file_url",
+    spec=FilesystemConfigurationResource,
+)
+def filesystem_resource(
+    bucket_url: str = dlt.secrets.value,
+    credentials: FileSystemCredentials = dlt.secrets.value,
+    filename_filter: Optional[str] = None,
+    chunksize: int = DEFAULT_CHUNK_SIZE,
+    extract_content: bool = False,
+) -> TDataItems:
```
rudolfix, line 94: let's use correct type here: `Iterator[List[FIleItem]]`

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    Returns:
+        TDataItems: The list of files.
+    """
+
+    fs_client = client_from_credentials(bucket_url, credentials)
+
+    # as it is a glob, we add a wildcard if no filter is given
+    if not filename_filter:
+        filename_filter = "*"
+    files_url = os.path.join(bucket_url, filename_filter)
```
rudolfix, line 114: use `posixpath`! os.path is will use `\` on windows

sources/standard/filesystem/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+
+    files_chunk = []
```
rudolfix, line 116: type that as List[FileItem]

sources/standard/filesystem/helpers.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Returns:
+        Iterable[FileItem]: The list of files.
+    """
+
+    protocol = urlparse(files_url).scheme or "file"
+    files = fs_client.glob(files_url, detail=True)
+
+    for file, md in files.items():
+        file_name = os.path.basename(file)
```
rudolfix, line 55: posixpath

sources/standard/inbox/__init__.py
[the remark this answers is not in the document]
rudolfix: OK let's keep what we have

sources/standard/inbox/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+                    new_attachments = extract_attachments(msg, mime_type)
+                    if not attachments:
+                        attachments = new_attachments
+                    else:
+                        attachments = chain(attachments, new_attachments)
+            email_info = extract_email_info(msg)
+            internal_date = get_internal_date(client, message_uid)
+
+            for attachment in attachments:
+                filename = f"{message_uid}_{attachment['file_name']}"
```
rudolfix, line 224: my take:
1. remove copying file to local
2. load the bytes into file_content
3. return FileSystemDict as in filesystem

sources/standard_pipeline.py
```diff
@@ -0,0 +1,119 @@
...
+    for file_obj in items:
+        file_dst = os.path.join(storage_path, file_obj["file_name"])
+        file_obj["path"] = file_dst
+        with open(file_dst, "wb") as f:
+            f.write(file_obj.read())
+        yield file_obj
+
+
+@dlt.transformer(  # type: ignore
+    table_name=lambda x: x["code"],
```
rudolfix, line 42: why we need that? I think it it too complicated

sources/standard_pipeline.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Args:
+        item (TDataItem): The list of files to copy.
+        incremental (dlt.sources.incremental[str], optional): The incremental source.
+
+    Returns:
+        TDataItem: The file content
+    """
+    for file_obj in items:
+        df = pd.read_csv(
```
rudolfix, line 66: please demonstrate a stream writer where we use csv reader to read page by page with given size

rudolfix (changes requested):
The next step is to add tests for all possible buckets: local filesyste, az, s3 and gs

tests/standard/test_file_source.py
```diff
@@ -0,0 +1,62 @@
+import dlt
+import pytest
+
+from sources.standard.filesystem import filesystem_resource
+from tests.utils import ALL_DESTINATIONS, assert_load_info, load_table_counts
+
+
+@pytest.mark.parametrize("destination_name", ALL_DESTINATIONS)
+def test_all_resources(destination_name: str) -> None:
```
rudolfix, line 9: we need a test that does not load data to database - you just instantiate `filesystem` with many different gobs and buckets and iterate it via list(). all resources and sources are just iterators. the tests I see:
- no glob - all files are returned (recursively)
- recursive and non recursive globs
- glob that is parsing files that contain dates (ie YYYY-MM-DD). - that one of the use cases
- you need to test all the details of the file item! (path, url, sizes, mtimes etc.) - I know this is some work so maybe you can test just one file (per bucket) for all the settings

tests/standard/test_file_source.py
```diff
@@ -0,0 +1,62 @@
...
+        dataset_name="file_source_data",
+        full_refresh=True,
+    )
+
+    # Load filter files
+    all_files = filesystem_resource(filename_filter="mlb*.csv")
+    load_info = pipeline.run(all_files)
+    assert_load_info(load_info)
+    table_counts = load_table_counts(pipeline, "filesystem")
+    assert table_counts["filesystem"] == 2
```
rudolfix, line 39: the filesystem_resource should be deselected by default and it should not load the file items (IMO) - but we need to say in docs how to enable it

please check the csv table as well

tests/standard/test_file_source.py
rudolfix: also test the FIleSystemDict operations
* load all the files
* open all the files (binary and text if applicable)
* verify the content of a single selected file

tests/standard/test_file_source.py
```diff
@@ -0,0 +1,62 @@
...
+        full_refresh=True,
+    )
+
+    @dlt.transformer
+    def ext_file(items) -> str:
+        content = items[0].read_bytes()
+        assert content == b"dlthub content"
+
+    # Load filter files
+    all_files = (
```
rudolfix, line 57: do that in the first test I mentiuoned

tests/standard/samples/freshman_kgs.csv
```diff
@@ -0,0 +1 @@
...
"M", 64, 68, 18.51, 19.83
"M", 66, 71, 21.40, 22.97
"F", 52, 57, 17.72, 19.42
"M", 71, 77, 22.26, 23.87
"F", 55, 60, 21.64, 23.81
"M", 65, 71, 22.51, 24.45
"M", 75, 82, 23.69, 25.80
"F", 42, 49, 15.08, 17.74
"M", 74, 82, 22.64, 25.33
"M", 94, 105, 36.57, 40.86
```
rudolfix, line 1: please use recursive folders for csvs to facilitate testing

rudolfix (changes requested):

sources/standard/inbox/README.md
```diff
@@ -0,0 +1,148 @@
...
+3. Customize the FILTER_EMAILS list in the `inbox/settings.py` file if you want to fetch emails
+   only from specific senders.
+4. Customize the GMAIL_GROUP in the `inbox/settings.py` file if you want to fetch emails for
+   specific Google Group.
+5. Set the DEFAULT_START_DATE in the `inbox/settings.py` file to the date you want to fetch emails
+   from.
+
+## Functionality
+
+### inbox_source source
```
rudolfix, line 59: we remove that

sources/standard/inbox/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+This dlt transformer resource takes an email message items from another resource (e.g.
+`messages_uids`) and extracts attachments from the email message using their UID.
+It connects to the IMAP server, fetches the email message by its UID, and return a file objects as
+filesystem. It yields a dictionary containing email metadata such as message UID, sender, date,
+content type, etc. It can be used with the same transformers used for filesystem.
+
+## Example
+
+Here's an example of how to use the `inbox_source` to fetch inbox emails and save attachments to a
```
rudolfix, line 87: give two examples
1. reading messages
2. reading attachments

sources/standard/inbox/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+from .settings import DEFAULT_CHUNK_SIZE, DEFAULT_START_DATE, FILTER_EMAILS, GMAIL_GROUP
+
+
+class ImapFileItem(FileItem):
+    """A Imap file item."""
+
+    file_hash: str
+
+
+@dlt.source
```
rudolfix, line 30: please remove that

sources/standard/inbox/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    if attachments:
+        return uids | get_attachments_by_uid(
+            filter_by_mime_type=filter_by_mime_type,
+        )
+    else:
+        return uids | read_messages
+
+
+@dlt.resource
```
rudolfix, line 68: make it selected=False

sources/standard_pipeline.py
```diff
@@ -0,0 +1,260 @@
+import json
```
rudolfix, line 1: maybe let's split into
filesystem_pipeline.py
inbox_pipeline.py

tests/standard/test_email_source.py
```diff
@@ -0,0 +1,38 @@
...
+        for item in items:
+            # Make sure just filtered emails are processed
+            assert "josue@sehnem.com" in item["From"]
+            if item["file_name"] == "dlthub.txt":
+                # Find the attachment with the file name and assert the loaded content
+                content = item.read_bytes()
+                assert item["file_name"] == "dlthub.txt"
+        assert content == b"dlthub content"
+
+    data_source = inbox_source(
```
rudolfix, line 28: we won't have source so please test messages and attachments

rudolfix (changes requested):

rudolfix (approved):

Issue #250, opened 2023-08-25
Add column schema to facebook_insights resource.


zem360:
**Source name**
Facebook_ads.

**Make sure that**
* [x] I checked the README in the source folder
* [x] If I have problems with the credentials I checked the [credentials docs](https://dlthub.com/docs/general-usage/credentials) and [troubleshooting](https://dlthub.com/docs/walkthroughs/run-a-pipeline#4-detect-and-handle-problems)

**Describe the bug**
The data loaded by the facebook_insights resource is being cast as *String*.

1.  dlt source: facebook_ads
2. dlt resource: facebook_insights 
3. destination: BigQuery
4. generated schema:
 `CREATE TABLE facebook_insights (
    campaign_id STRING NOT NULL,
    adset_id STRING NOT NULL,
    ad_id STRING NOT NULL,
    date_start STRING NOT NULL,
    cpc STRING,
    spend STRING,
    impressions STRING,
    reach STRING,
    cpm STRING,
    ctr STRING,
    unique_clicks STRING,
    date_stop STRING,
    cpp STRING,
    clicks STRING,
    unique_ctr STRING,
    frequency STRING,
    account_id STRING,
    _dlt_load_id STRING NOT NULL,
    _dlt_id STRING NOT NULL
);`

**To Reproduce**
- Run the facebook_insights resource with default parameters..

**Running environment**
 - OS: MacOS
 - `dlt`: 0.3.12

**FIx**
Add the columns schema via columns in facebook_insights resource.

rudolfix:
@zem360 any progress on it?

PR #266, opened 2023-09-20
Facebook Insight Resource Fixing Field Data Types

zem360:
# Tell us what you do here

# Relevant issue

issue #250 

# More PR info
facebook_insight resource cast all fields as strings. Fixing datatype for the default insight fields.


rudolfix (approved):
let's merge! ads token expired again but I will not refresh it until we have something more serious to test.
thx!

Issue #267, opened 2023-09-21
Applying user endpoint for Slack Source Integration


dat-a-man:
Which source is this for
Slack 

Describe the data you'd like, be specific, link the docs
I've been testing the Slack source, and it works quite well. However, I'd like to suggest an improvements that could make it even better:

The current pipeline fetches the user_id, which is helpful. However, for more insightful analytics, it would also be valuable to retrieve user_name and email address. This would aid in a more granular analysis and offer a more comprehensive user-centric view.

Additional context
Enriching the data with additional user endpoints can lead to better insights for end-users.

I hope these suggestions help in further refining the Slack source integration. Thanks for the great work so far!

rudolfix:
done in #269

PR #269, opened 2023-09-22
add users to slack source

sehnem:
# Tell us what you do here

# Relevant issue

issue #267

# More PR info
Add the list user endpoint and other small adjustments.


sources/slack/__init__.py
```diff
@@ -58,16 +63,28 @@ def channels_resource() -> Iterable[TDataItem]:
...
-        "latest_reply",
-        "blocks.thread_ts",
-        "blocks.latest_reply",
-        "attachment.thread_ts",
-        "attachment.latest_reply",
-        "edited.ts",
-    ]
+    @dlt.resource(name="users", primary_key="id", write_disposition="replace")
+    def users_resource(
+        datetime_fields: List[str] = DEFAULT_DATETIME_FIELDS,
```
rudolfix, line 68: It is a bad idea to pass lists as default parameters. I wonder why linter is not complaining. (maybe because this is inner function). Anyway - I do not think you need to pass it here. Just remove the argument and use `DEFAULT_DATETIME_FIELDS` like in other places

sources/slack_pipeline.py
```diff
@@ -45,8 +41,25 @@ def select_resource(selected_channels: List[str]) -> None:
...
+
+
+def get_users() -> None:
+    """Execute a pipeline that will load Slack users list."""
+
+    pipeline = dlt.pipeline(
+        pipeline_name="slack", destination="duckdb", dataset_name="slack_data"
     )
 
+    source = slack_source(
```
rudolfix, line 59: cool! but please remember to add it to tests!

rudolfix (changes requested):
please add a test for user endpoint. otherwise this looks fine

sources/slack_pipeline.py
sehnem: I am just waiting for the permissions in the main channel account to create the tests.

sources/slack/__init__.py
rudolfix: does not look like resolved - do not pass a list as default argument. remove argument altogether. you do not really need that

rudolfix (changes requested):
1. default argument is still there - see review
2. test where we have "-" in table names worries me. let's talk on slack!

rudolfix (approved):

PR #270, opened 2023-10-03
fix mongo conversion issue

sehnem:
# Tell us what you do here

# Relevant issue

# More PR info
There was a bug when converting Ids and decimals inside nested objects, this PR fixes that and also adds tests to these cases. A testing object was added to the db for that.

### DIFF
--- a/sources/mongodb/helpers.py
+++ b/sources/mongodb/helpers.py
@@ -87,7 +87,7 @@ def collection_documents(
         yield docs_slice
 
 
-def convert_mongo_objs(value):
+def convert_mongo_objs(value: Any) -> Any:
     if isinstance(value, (ObjectId, Decimal128)):
         return str(value)
     if isinstance(value, _datetime.datetime):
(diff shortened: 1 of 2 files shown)


sources/mongodb/helpers.py
```diff
@@ -59,19 +57,32 @@ def make_query(self) -> TCursor:
 
     def clean_document(self, document: Dict[str, Any]) -> Any:
         """Clean document to be json serializable"""
-        output = {}
+        # Convert lists to json serializable
```
rudolfix, line 60: this is a hack :) what you need to do is to recursively apply the simple mapping function recursively.
we have a util function: map_nested_in_place in dlt.common.util

the think that you need to write is a simple mapping that you have below and pass it:
```
            if isinstance(value, Decimal128):
                output[key] = value.to_decimal()
            elif isinstance(value, ObjectId):
            elif isinstance(value, (ObjectId, Decimal128)):
                output[key] = str(value)
else: return value

```

rudolfix (changes requested):

rudolfix (approved):
LGTM!

Issue #275, opened 2023-10-12
[Kinesis] verified source


sehnem:
# Quick source info
- Name of the source: Kinesis
- What is the data source: kinesis streams

# Current Status
- [x] I plan to write it
- [ ] I've written it and run it for fun
- [ ] I run it on my local laptop
- [ ] I run it in production
- I tried following destinations with it: [e.g. duckdb, bigquery]

# What source does/will do
Please provide descriptions up to 3 most important use cases that users of this source do. Those use cases will be:

The source will fetch the kinesis events for a defined period of time, it can be ran in a loop to keep fetching the events and with the storage of the incremental it will be possible to resume fetching the latest records.

use case description is not only targeted at the developers but also by people creating test account and doing demos.

# Test account / test data
- [ ] I'd like `dltHub` to create a test account before we start
- [ ] I have test account and can share my test data or a test account with `dltHub` after we merge the source

# Additional context
[Anything we should know about your source](https://aws.amazon.com/pt/pm/kinesis) is a aws tool built to stream data in real time.

PR #276, opened 2023-10-12
Amazon kinesis source

sehnem:
# Tell us what you do here

- [X] implementing verified source (please link a relevant issue labelled as `verified source`)

# Relevant issue

issue #275

# More PR info


tests/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+        | segment_stream
+    )
+
+    # Load entire database
+    assert_load_info(info)
+    table_names = [t["name"] for t in pipeline.default_schema.data_tables()]
+    table_counts = load_table_counts(pipeline, *table_names)
+    expected_tables = []
+    assert set(table_counts.keys()) >= set(expected_tables)
```
rudolfix, line 54: I do not get this :)

tests/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,55 @@
...
+
+import json
+import dlt
+import pytest
+from dlt.common import pendulum
+
+from sources.kinesis import read_kinesis_stream
+from tests.utils import ALL_DESTINATIONS, assert_load_info, load_table_counts
+
+@pytest.mark.parametrize("destination_name", ALL_DESTINATIONS)
```
rudolfix, line 12: duckdb test is IMO enough

tests/kinesis/test_kinesis.py
rudolfix: also please select form duckdb to make sure the message is there

tests/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,55 @@
...
+        | segment_stream
+    )
+
+    # Load entire database
+    assert_load_info(info)
+    table_names = [t["name"] for t in pipeline.default_schema.data_tables()]
+    table_counts = load_table_counts(pipeline, *table_names)
+    expected_tables = []
+    assert set(table_counts.keys()) >= set(expected_tables)
+    assert table_counts["sessions"] == 1
```
rudolfix, line 55: now run pipeline again to make sure you didn't receive message again (incremental!)

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        shard_iterators.append((shard_id, shard_iterator["ShardIterator"]))
+
+    records = []
+    for shard_id, shard_iterator in shard_iterators:
+        records_response = kinesis_client.get_records(
+            ShardIterator=shard_iterator, Limit=10000
+        )
+        for record in records_response["Records"]:
+            data = json.loads(record["Data"])
```
rudolfix, line 56: data should be base64 right? becaiuse you can send binaries into kinesis.

we need to test those cases

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+            if isinstance(data, str):
+                data = json.loads(data)
+            sequence_number = record["SequenceNumber"]
+            timestamp = record["ApproximateArrivalTimestamp"]
+            data["_shard_id"] = shard_id
+            data["_seq_no"] = sequence_number
+            data["_kinesis_ts"] = timestamp
```
rudolfix, line 63: maybe also add stream_name

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+        region_name=aws_region,
+    )
+    shards = kinesis_client.list_shards(StreamName=stream_name)["Shards"]
+    shard_iterators = []
+
+    for shard in shards:
+        shard_id = shard["ShardId"]
+        if timestamp := last_ts.last_value:
+            iterator_params = dict(
+                ShardIteratorType="AT_TIMESTAMP", Timestamp=timestamp
```
rudolfix, line 39: this is good for testing but not for production. let's build a full version with all the tests and then switch to iterator based on sequence. I had it in my code but it requires to store last value per shard which is possible but tricky with the incremental

rudolfix (changes requested):

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+            if isinstance(data, str):
+                data = json.loads(data)
+            sequence_number = record["SequenceNumber"]
+            timestamp = record["ApproximateArrivalTimestamp"]
+            data["_shard_id"] = shard_id
+            data["_seq_no"] = sequence_number
+            data["_kinesis_ts"] = timestamp
+
+            records.append(data)
+            if len(records) >= chunk_size:
```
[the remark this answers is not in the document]
sehnem: Yeah, I found that you can only fetch a certain number of records from a kinesis shard, it will just return those and other shards to fetch the data from, thats why I am setting the limit to max and then adding chunks, but I will explore more that.

sources/kinesis/__init__.py
sehnem: I checked the docs, it must be bytes and the library encodes it to base64 before sending.

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+                        "_kinesis_partition": partition,
+                        "stream_name": stream_name,
+                        "data": content,
+                    }
+                )
+                if len(records) >= chunk_size:
+                    yield records
+                    records = []
+            # If we are 0ms behind the latest record we can break out of the loop
+            if not records_response.get("MillisBehindLatest", 0):
```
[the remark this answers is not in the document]
rudolfix: also let's make the offset configurable. IMO 1 second should be the default

tests/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,96 @@
...
+    pipeline = dlt.pipeline(
+        pipeline_name="kinesis_test",
+        destination="duckdb",
+        dataset_name="kinesis_test_data",
+    )
+
+    stream_name = "dlt_ci_kinesis_source"
+    test_id = str(uuid.uuid4())
+
+    credentials = dlt.secrets._get_value("sources.kinesis_pipeline.credentials")[0]
```
rudolfix, line 23: you should really not need that! just the correct secrets toml format

rudolfix (changes requested):
please see the review. there are even more nuances when dealing with the shards but lets fix this

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        shard_iterator = kinesis_client.get_shard_iterator(
+            StreamName=stream_name, ShardId=shard_id, **iterator_params
+        )
+        shard_iterators.append((shard_id, shard_iterator["ShardIterator"]))
+
+    records = []
+    for shard_id, main_shard_iterator in shard_iterators:
+        shard_iterator = main_shard_iterator
+        while True:
+            records_response = kinesis_client.get_records(
```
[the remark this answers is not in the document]
sehnem: I am looping with `shard_iterator = records_response["NextShardIterator"]` so I get the other records from this shard, until we reach 0 `MillisBehindLatest`.

rudolfix (approved):
@sehnem I think all updates I need are there. your last algo to pick shards was OK, I refactored the code to make test easier

the last final test will pass once we release new dlt lib

sources/kinesis/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    stream_name: str = dlt.config.value,
+    credentials: AwsCredentials = dlt.secrets.value,
+    last_msg: Optional[dlt.sources.incremental[StrStr]] = dlt.sources.incremental(
+        "_kinesis", last_value_func=max_sequence_by_shard
+    ),
+    initial_at_timestamp: TAnyDateTime = 0.0,
+    max_number_of_messages: int = None,
+    milliseconds_behind_latest: int = 1000,
+    parse_json: bool = True,
+    chunk_size: int = 1000,
```
sehnem, line 29: It supports up to 10_000 records chunks, but I think that 1_000, as it is right now, is a more reasonable amount based on the kinesis limits. I am also not sure about the number of request per second that it will be doing, this way we can avoid some limit issues.

> Each data record can be up to 1 MiB in size, and each shard can read up to 2 MiB per second. You can ensure that your calls don’t exceed the maximum supported size or throughput by using the Limit parameter to specify the maximum number of records that GetRecords can return. Consider your average record size when determining this limit. The maximum number of records that can be returned per call is 10,000.

rudolfix (approved):

PR #277, opened 2023-10-12
adds readers to filesystem

rudolfix:
# Tell us what you do here

# Relevant issue
Adds a set of readers to filesystem source


sources/filesystem/__init__.py
```diff
@@ -11,9 +12,36 @@
     FilesystemConfigurationResource,
     fsspec_from_resource,
 )
+from .readers import ReadersSource, _read_csv, _read_jsonl, _read_parquet
 from .settings import DEFAULT_CHUNK_SIZE
 
 
+@dlt.source(_impl_cls=ReadersSource, spec=FilesystemConfigurationResource)
```
sehnem, line 19: I have thought about adding some readers, not just in the example but had no idea, this implementations looks really good!

PR #298, opened 2023-12-07
Fix: update_time is NoneType

AstrakhantsevaAA:
fix this issue https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1701947607888769

```
<class 'dlt.extract.exceptions.ResourceExtractionError'>
In processing pipe leads: extraction of resource leads in generator leads caused an exception: 'NoneType' object has no attribute 'last_value'
```

PR #302, opened 2023-12-14
fix(pipedrive): incremental data is None

IlyaFaer:
Fixing a bunch of Pipedrive things:

1. let's use `"1970-01-01 00:00:00"` as a default for `since_timestamp`. that should fix the problem
2. let's drop the `Optional` from all Incremental declarations. they are not optional
3. we unified how we deal with datetimes/timestamps in our sources. please look ie. at `def zendesk_talk` and how `TAnyDateTime` is used together with `ensure_pendulum_datetime` (yeah pipedrive is using some non ISO date format... so part of the conversion function should stay)
4. Steini left this comment
```
# TODO: This check can be replaced with `update_time.start_out_of_range` in dlt 0.3.5
```
let's implement it

See the original [issue](https://github.com/dlt-hub/verified-sources/pull/298#issuecomment-1854468600)

### DIFF
--- a/sources/pipedrive/__init__.py
+++ b/sources/pipedrive/__init__.py
@@ -187,7 +187,6 @@ def leads(
     )
     # Load leads pages sorted from newest to oldest and stop loading when
     # last incremental value is reached
-    last_value = update_time.last_value
     pages = get_pages(
         "leads",
         pipedrive_api_key,
--- a/sources/pipedrive/__init__.py
+++ b/sources/pipedrive/__init__.py
@@ -15,18 +15,19 @@ import dlt
 
 from .helpers.custom_fields_munger import update_fields_mapping, rename_fields
 from .helpers.pages import get_recent_items_incremental, get_pages
-from .helpers import parse_timestamp, group_deal_flows
+from .helpers import group_deal_flows
 from .typing import TDataPage
 from .settings import ENTITY_MAPPINGS, RECENTS_ENTITIES
-from dlt.sources import DltResource
 from dlt.common import pendulum
+from dlt.common.time import ensure_pendulum_datetime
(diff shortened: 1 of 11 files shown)


sources/pipedrive/__init__.py
```diff
@@ -184,17 +187,13 @@ def leads(
     )
     # Load leads pages sorted from newest to oldest and stop loading when
     # last incremental value is reached
-    last_value = update_time.last_value
```
IlyaFaer, line 187: This value is no longer needed, `start_out_of_range` should do the job.

sources/pipedrive/helpers/__init__.py
```diff
@@ -5,18 +5,6 @@
...
-def parse_timestamp(timestamp: Any) -> str:
-    if not timestamp:
-        return None
-    if isinstance(timestamp, str):
-        timestamp = pendulum.parse(timestamp)
-    assert isinstance(
-        timestamp, pendulum.DateTime
-    ), "since_timestamp must be a valid ISO datetime string or pendulum.DateTime object"
-    timestamp = timestamp.in_timezone("UTC")
-    return cast(str, timestamp.to_iso8601_string().replace("T", " ").replace("Z", ""))
```
IlyaFaer, line 17: It's used only for Pipedrive, so it's no longer needed

rudolfix: thanks for cleaning that up!

rudolfix (approved):
LGTM! thanks for cleaning that up!

rudolfix (changes requested):
just one more thing: please resubmit this PR directly so all tests can run.

rudolfix:
continued in #303

PR #303, opened 2023-12-14
fix(pipedrive): incremental date is None

IlyaFaer:
Fixing a bunch of Pipedrive things:

1. let's use `"1970-01-01 00:00:00"` as a default for `since_timestamp`. that should fix the problem
2. let's drop the `Optional` from all Incremental declarations. they are not optional
3. we unified how we deal with datetimes/timestamps in our sources. please look ie. at `def zendesk_talk` and how `TAnyDateTime` is used together with `ensure_pendulum_datetime` (yeah pipedrive is using some non ISO date format... so part of the conversion function should stay)
4. Steini left this comment
```
# TODO: This check can be replaced with `update_time.start_out_of_range` in dlt 0.3.5
```
let's implement it

See the original [issue](https://github.com/dlt-hub/verified-sources/pull/298#issuecomment-1854468600)


rudolfix (changes requested):
@IlyaFaer several tests are not passing
1. please do not remove `Optional` from other sources. They really have it optional so some tests are failing
2. Please skip all tests for `unstructured_source`. they are segfaulting randomly (langchain + chrome - this really does not work). we planned to do thst anyway
3. pipedrive tests are not passing! please figure out the problems locally. you may want to bump dlt to the newest version

tests/pipedrive/test_pipedrive_source.py
```diff
@@ -323,10 +323,12 @@ def test_resource_settings() -> None:
 
     assert source.resources["custom_fields_mapping"].write_disposition == "replace"
 
+    s = source.discover_schema()
+
     for rs_name in resource_names - {"custom_fields_mapping"}:
         rs = source.resources[rs_name]
         assert rs.write_disposition == "merge"
-        assert rs.table_schema()["columns"]["id"]["primary_key"] is True
```
IlyaFaer, line 329: It doesn't look like this method even exists

IlyaFaer:
@rudolfix, I fixed the failing test, it was about the test itself, not functionality.

There are still two failures:
```
FAILED tests/personio/test_personio_source.py::test_all_resources[postgres] - assert 2 >= 21
FAILED tests/personio/test_personio_source.py::test_all_resources[duckdb] - assert 2 >= 21
```
I saw them in other PRs. I'll need credentials for personio, if it needs to be fixed in this PR as well.

Also, there was a failure `postgres transient error` - I restarted the workflow, as it's a maintenance error, not related to functionality.

AstrakhantsevaAA:
@IlyaFaer rebase please on master, personio tests were fixed

IlyaFaer:
Okay, personio errors gone. Postgres still remains:
```
psycopg2.OperationalError: connection to server at "dlttests.cwz0jfxu0m7m.eu-central-1.rds.amazonaws.com" (18.196.0.3), port 5432 failed: FATAL:  the database system is in recovery mode
```
Seems some actions in the test database on Amazon are required.

tests/pipedrive/test_pipedrive_source.py
rudolfix: huh! how it was working before. anyway: thanks for fixing!

rudolfix (approved):
LGTM! IDK what AWS is doing with out postgres instance but all those recovery modes end by themselves

Issue #273, opened 2023-10-04
[kafka] verified source


rudolfix:
# Quick source info
Incrementally reads events from a specified kafka topics.

# Current Status
- [x] I plan to write it

# What source does/will do
The source reads events from a specified topics or set of topics and yields them in configurable chunks. It implements incremental loads so events are not duplicated.

Following use cases should be demonstrated:
- reading events from one topic and dispatching them to tables by event type
- reading events from many topics and dispatching them by topic name
- passing KafkaConsumer directly instead of credentials
- how to configure event and channel metadata that will be passed along with the event

Following authorization methods should be implemented:
- PLAIN, SSL, SASL_PLAIN + SSL, directly pass consumer

Optional: reading topic schemas and converting them into pydantic model that we pass as `columns` to resource. Please investigate

# Test account / test data
- [x] I'd like `dltHub` to create a test account before we start

# Additional context
Incremental loading is tricky and must be well tested. Where to start:
1. dont' use autocommit
2. store the offset information in resource state
3. I do not think we need incremental here. depends how offsets are represented
4. **use the state to cut offset on next run before getting messages, not in this run**

nicor88:
Something more to add to this is issue is:
* Being able to use custom decoders after message consumption.
e.g. in the past, I've been working with 3 type of data serialization with Kafka:
* pure JSON
* pure Avro - the schema is attached in the message
* Avro with schema registry - only the reference to the schema registry is attached in the first 5 bytes, then the avro bytes are in the remaining bytes

Said so, the concept of message Decoding is something that must be configurable.

PR #306, opened 2023-12-15
feat(sources.kafka): implement Kafka source

IlyaFaer:
Towards: #273 

Functional implementation of the Kafka source.

**Extraction**
When extraction just starts, the source reads the maximum offset for every given topic/partition and reads data only up to them. As soon as current offsets are equal to the maximum offsets, reading stops (instead of waiting for new messages in a loop).

**Consumer**
Only one consumer is used to read data. An initiated consumer object can be given to the source, as well as credentials. If no credentials given, the consumer will be initiated from secrets. The same consumer reads data from all the given topics and directs messages to particular tables (one table for one topic).

**Topics read**
As a user can have a lot of topics, it makes sense to read all the topics from the cluster with just one requests and then choose those we need to extract (Kafka allows to read all the topics from the cluster, or just one - and reading topics one-by-one can produce too many separate requests).

**Message processor**
A user can pass a function into the source, which'll be called for every read message in order to process it. As a Kafka message includes topic and partition, only one processor is needed - if a user wants to process messages from different topics in different way, they can make `if-else` in the same processor.

**Offsets tracker**
As we can have a lot of topics and partitions at the same time, it makes sense to track them in a separate object of a form:
```python
{
    "topic_name": {
        "partition_name": {
            "cur": 2,   # current offset in the partition
            "max": 10,   # maximum offset in the partition (length)
        }
    }
}
```
When a message read, tracker updates current offset. When `cur == max`, it means we read all the messages we had at the extraction start.

**State**
The pipeline state keeps offsets index and uses it in a form of an incremental, thus, on the next extractions, only new messages are read.


sources/kafka_pipeline.py
```diff
@@ -0,0 +1,26 @@
...
+from sources.kafka import kafka_consumer
+
+
+# def my_msg_proc(msg):
+#     return dict(
+#         topic=msg.topic(),
+#         key=msg.key().decode("utf-8"),
+#         value="no!",
+#         partition=msg.partition(),
+#     )
```
IlyaFaer, line 11: An example of a custom messages processor

rudolfix (changes requested):
this looks good overall. I had several comments where I'd like to follow our patterns more closely. pls take a look

sources/kafka/helpers.py
```diff
@@ -0,0 +1,145 @@
...
+    """Basic Kafka message processor.
+
+    Args:
+        msg (confluent_kafka.Message): A single Kafka message.
+
+    Returns:
+        dict: Processed Kafka message.
+    """
+    return dict(
+        key=msg.key().decode("utf-8"),
```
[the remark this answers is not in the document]
IlyaFaer: Done

sources/kafka/helpers.py
```diff
@@ -0,0 +1,145 @@
...
+
+            # designate current and maximum offsets for every partition
+            for i, part in enumerate(parts):
+                cur_offset = self._cur_offsets[t_name].get(
+                    str(part.partition), OFFSET_BEGINNING
+                )
+
+                self[t_name][str(part.partition)] = {
+                    "cur": cur_offset,
+                    "max": self._consumer.get_watermark_offsets(part)[1],
```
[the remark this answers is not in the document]
IlyaFaer: Yeah, I know, but I don't see how to do it in one request, even `AdminClient` does it per one partition. There are several feature requests in the Confluent Kafka repos for methods, which could give all the watermarks, however, they are all open. I think Kafka is just not intended for operations like this, that's why they don't rush to implement it.

sources/kafka/helpers.py
```diff
@@ -0,0 +1,145 @@
...
+        for t_name, topic in self._topics.items():
+            self[t_name] = {}
+
+            # init all the topic partitions from the partitions' metadata
+            parts = [TopicPartition(t_name, part) for part in topic.partitions]
+
+            # designate current and maximum offsets for every partition
+            for i, part in enumerate(parts):
+                cur_offset = self._cur_offsets[t_name].get(
+                    str(part.partition), OFFSET_BEGINNING
```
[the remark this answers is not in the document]
IlyaFaer: This is possible. I pushed a raw version - working fine, but the code seems a bit complex. Will try to write some facades tomorrow.

sources/kafka/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+    consumer.subscribe(topics)
+    tracker = OffsetTracker(consumer, topics, dlt.current.resource_state(), start_from)
+
+    # read the messages up to the maximum offsets,
+    # not waiting for new messages
+    while tracker.has_unread:
+        batch = []
+        for msg in consumer.consume(batch_size, timeout=1):
+            if msg.error():
+                print(f"ERROR: {msg.error()}")
```
rudolfix, line 84: any ideas what message error could be? we typically raise and abort on any errors but maybe it is something different? I expect the problem with the read is an exception

sources/kafka/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+
+    return {
+        "_kafka": {
+            "partition": partition,
+            "topic": topic,
+            "key": key,
+            "offset": msg.offset(),
+            "ts": {
+                "type": ts[0],
+                "value": ensure_pendulum_datetime(ts[1] / 1e3),
```
rudolfix, line 43: how does that work when `TIMESTAMP_NOT_AVAILABLE`? you should probably skip setting value in that case

sources/kafka/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+            },
+            "data": msg.value().decode("utf-8"),
+        },
+        "_kafka_msg_id": digest128(topic + str(partition) + str(key)),
```
rudolfix, line 47: nothing against that. is `key()` really not globally unique though?

sources/kafka/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+            else:
+                batch.append(msg_processor(msg))
+                tracker.renew(msg)
+
+        yield batch
+
+    consumer.close()
```
rudolfix, line 91: maybe use context manager instead, if not put that into `finally`

sources/kafka/helpers.py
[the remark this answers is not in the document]
rudolfix: OK! but here you can get offset for timestamps `ts_offsets = self._consumer.offsets_for_times(parts)` so you could do the same to get offset for `now`.
but maybe there is a reason not to do that :) anyway - we can live with the thing below

rudolfix (changes requested):
this is really good. the structure is there. just minor issues with the code itself.
now we need tests. like with kinesis, tests should be self contained (so we send and receive messages ie. to some random topics). 
from my POV the logic that initializes offset is quite complicated and we should test cases both with `start_from`and without.
also incremental load. but that should be easy

sources/kafka/__init__.py
IlyaFaer: Actually, yes, I'm periodically getting an error like this, when I start extraction:
```ERROR: KafkaError{code=UNKNOWN_TOPIC_OR_PART,val=3,str="Subscribed topic not available: books: Broker: Unknown topic or partition"}```

The thing is - several seconds pass, and it reads messages without problems (in the same loop). According to the docs, when you just start reading, it may take several seconds before consumer will be balanced to a broker - seems, we're trying to read before balancing finished, but it's not a reason to abort extraction.

The list of possible errors:
https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.KafkaError

There is a couple of methods, which can be used: `fatal()` and `retriable()`. If the error is not fatal or is retriable (it's not the same), then we continue. Otherwise, we'll abort extraction. Seems logical to me, will do it.

File: sources/kafka/helpers.py (hunk 7 above)
Comment (IlyaFaer) [reply sources/kafka/helpers.py:47]: It's not. Still, keys are used to calc hashes, which allows to decide into which partition the message must be saved.

Untitled

File: sources/kafka/__init__.py (hunk 8 above)
Comment (IlyaFaer) [reply sources/kafka/__init__.py:91]: Done

File: sources/kafka/helpers.py (hunk 6 above)
Comment (IlyaFaer) [reply sources/kafka/helpers.py:43]: According to the docs, the results is always `(int, int)`
I would keep this as is, not to create addition layer above Kafka's behavior.

File: sources/kafka/helpers.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (IlyaFaer) [reply sources/kafka/helpers.py:145]: Actually, I thought about the same, I even wrote you a message, but then deleted it :sweat_smile: 

Offset is a cursor at which the **user** is reading now. It's not the same as the **topic's** cursor for new messages insert. Messages are added to the end of the topic, but the user's offset may be somewhere in the middle. And we need both - to understand where to start reading (incremental), and when to stop (not to get into endless loop).

And when we're talking about getting offsets for timestamps, it's about user's offset, without the topic length. Also, there is quite strange trait of behavior in it:
```
If the provided timestamp exceeds that of the last message
in the partition, a value of -1 will be returned.
```
Seems to me we can't use `now` or any values close to it, because it'll give `-1` in most of the cases. `-1` means `no offset available for partition [...] happens at the very first read, or the group's offset expiration/deletion`.

I must say it's a very strange behavior by my opinion, but it's what it is. So, we can start reading from a particular timestamp, but we still don't know when to stop, and we'll need to somehow process situations, when our timestamp is after the last message.

File: sources/kafka/__init__.py (hunk 5 above)
Comment (rudolfix) [reply sources/kafka/__init__.py]: yeah! what I do not understand in the documentation is what `fatal` means. if `fatal` is a fatal error concerning particular message - you should rather skip the message and increase the offset so you never re-read it.
if the fatal error means that the consumer should disconnect because no further message consumption is possible then we need to raise. pls make sure what is going on

File: sources/kafka/helpers.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (rudolfix) [reply sources/kafka/helpers.py:145]: uh! this sucks. we must add checks against that in our code if indeed
`ts_offsets = self._consumer.offsets_for_times(parts)`
may create offset of -1 when our timestamp is too new, we must fallback to max time from watermark when creating curr_offset

also you must update the pipeline state at the end `_init_partition_offsets` with the current offsets. so in the case above we use the stored watermark as starting point when pipeline is run ie. an hour later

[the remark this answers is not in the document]
Comment (rudolfix) [reply sources/kafka/helpers.py:145]: this is also important test case
- send a message
- consume with start timestamp to now. you should get -1 as the offset
- the code should fallback to watermark and store in the state as current offset. it returns no messages
- then you send a message
- run the pipeline again and the second message should be returned

File: sources/kafka/__init__.py (hunk 5 above)
Comment (IlyaFaer) [reply sources/kafka/__init__.py]: @rudolfix, this is what the docs say about fatal errors:
```Idempotent producer will typically only raise fatal errors for unrecoverable cluster errors where the idempotency guarantees can’t be maintained```

So, it's about serios failures of the whole system.

As for errors on particular messages - I don't think it's possible. By default, Kafka doesn't store messages that have problems, but raises exceptions. However, the user may create a `dead queue` to save unprocessable messages in a separate topic. In this case they'll only need to add one more topic into the resource before extraction, and with a custom message processor they'll solve whatever is wrong there.

**Comment by IlyaFaer** [comment]
@rudolfix, okay, I pushed four test cases, quite complex. Locally works fine, but requires credentials in the GH Actions. I suppose it's you, who'll add them into Secrets or something?

image

**Review by rudolfix — CHANGES_REQUESTED** [review]
yeah I added the credentials. you have access to them in secret manager. thx for the tests!
1. the timestamp tests is not passing (on CI and locally)
2. there's a big comment thread that is not yet fixed

File: tests/kafka/test_kafka.py [hunk 9]
```diff
@@ -0,0 +1,232 @@
...
+    assert_load_info(load_info)
+
+    assert_query_data(
+        pipeline,
+        f"SELECT _kafka__key FROM {topic} ORDER BY _kafka__key",
+        ["1", "2"],
+    )
+
+
+def test_kafka_incremental_read(kafka_producer, kafka_topics):
```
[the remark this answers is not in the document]
IlyaFaer: Okay, I made the test more complex. The thing is - it can have flaky errors. For now, I didn't get them in these tests, but I when I was testing manually, it happened several times. For example, a topic is not yet replicated, by a consumer is already trying to read messages from it, using an unreplicated broker. Replication and balancing should be reduced on the cluster. We're not reading data eternally (as it was designed to be in Kafka), but extract a fixated amount of data - replication will only create problems in this case.

sources/kafka/helpers.py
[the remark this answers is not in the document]
IlyaFaer: Done

sources/kafka/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    table_name=lambda msg: msg["_kafka"]["topic"],
+    standalone=True,
+)
+def kafka_consumer(
+    topics: Union[str, List[str]],
+    credentials: Union[KafkaCredentials, Consumer] = dlt.secrets.value,
+    msg_processor: Optional[
+        Callable[[Message], Dict[str, Any]]
+    ] = default_msg_processor,
+    batch_size: Optional[int] = 3000,
```
sultaniman, line 34: should this also be configurable?

sources/kafka_pipeline.py
```diff
@@ -0,0 +1,75 @@
...
+            msg (confluent_kafka.Message): Message read from the cluster.
+
+        Returns:
+            dict: converted message.
+        """
+        return {
+            "_kafka": {
+                "topic": msg.topic(),  # required field
+                "key": msg.key().decode("utf-8"),
+                "partition": msg.partition(),
```
nicor88, line 45: there are some relevant information that must be stored:
* msg.offset - helpful for further downstream de-duplications
* msg.timestamp() - when the event was inserted into kafka- again really helpful for further downstream de-duplications
* event.headers() - could be relevant at time

sources/kafka_pipeline.py
```diff
@@ -0,0 +1,75 @@
...
+        Args:
+            msg (confluent_kafka.Message): Message read from the cluster.
+
+        Returns:
+            dict: converted message.
+        """
+        return {
+            "_kafka": {
+                "topic": msg.topic(),  # required field
+                "key": msg.key().decode("utf-8"),
```
nicor88, line 44: bare in mind that this doesn't work all the time, you might want to provide a way for the users to specific the type of encoding used.
e.g.
```
def message_key_decoder(*, decoder_type, message_key, deserializer):
    """
    Given the type of decoder it returns the decoded message as dict
    """
    if decoder_type == 'json' or decoder_type == 'json_schema':
        return message_key.decode('utf-8')
    elif decoder_type == 'avro_schema':
        return deserializer.__call__(message_key, None)
    else:
        raise ValueError('Decoder type not supported')
```

if the key is avro serialized this will fail.

sources/kafka_pipeline.py
```diff
@@ -0,0 +1,75 @@
...
+                "partition": msg.partition(),
+            },
+            "data": msg.value().decode("utf-8"),
```
nicor88, line 47: same as https://github.com/dlt-hub/verified-sources/pull/306/files#r1449034951

this works if the record is a JSON, it's pretty common to use avro with a schema registry, if it's the case you might want to consider to le the final user specify a custom serializer.

here my really rough implementation:

```
def message_value_decoder(*, decoder_type, message_value, deserializer):
    """
    Given the type of decoder it returns the decoded message as dict
    """
    if decoder_type == 'json':
        return json.loads(message_value.decode('utf-8'))
    elif decoder_type == 'json_schema':
        message = io.BytesIO(message_value)
        # exclude the first 5 bytes of the message
        # in case the schema registry is used the first byte is the magic byte and the rest is the schema_id
        # in case of JSON encoding we could use the schema_id just to extra validate the event
        magic, schema_id = unpack('>bI', message.read(5))
        logger.debug(f'Message: magic_byte: {magic}, schema_id: {schema_id}')
        return json.loads(message.read())
    elif decoder_type == 'avro_schema':
        return deserializer.__call__(message_value, None)
    else:
        raise ValueError('Decoder type not supported')
```

when working with avro and confluent schema registery a simple implementation of deserializer is the following:
```
from confluent_kafka.schema_registry.avro import AvroDeserializer
deserializer = AvroDeserializer(schema_registry_client=schema_registry_client)
```

sources/kafka/__init__.py
nicor88: good catch, it should be configurable, as the final user might want to have an initial batch size big enough to cover the all topic load.

nicor88:
I had a look and overall looks good to me, but there are some aspects missing:
- possibility to support decoding of keys and message that are not JSON, common used formats are pure Avro, pure ProtoBuf, Avro with schema registry (first 5 bytes of the message contain the schema id)...
- possibility to configure batch size

Also I need a clarification on this aspect:
- what happens if the processing goes wrong (e.g. the processor crash for whatever reason?)?
   will the messages that are already consumed be committed even if they are not written  to the destination? how do you handle such edge case in the current implementation?

sources/kafka/__init__.py
IlyaFaer: Well, that's a source argument. It's configurable, `3000` is just a default value.

IlyaFaer:
@nicor88, about messages processing... The solution here is to allow the user to pass their own message processor. Those test case and pipeline sample you commented are actually just examples.

Whatever you want, Avro, JSON or anything else - you just need to write your own function, which accepts a message, converts it as you wish, and returns a `dict` - analogic to the default one:
https://github.com/dlt-hub/verified-sources/blob/5d334bed5cdd149b1c5882410fd44b97850b0da1/sources/kafka/helpers.py#L14-L50

And then pass it into the source as an argument, like it's done in the test:
https://github.com/dlt-hub/verified-sources/blob/5d334bed5cdd149b1c5882410fd44b97850b0da1/sources/kafka_pipeline.py#L32-L50

Unlikely it's possible for us to cover all the possible cases. You want partition numbers to be extracted, but another user doesn't, the third one wishes to process the "dead queue", which may have broken or fluid format, and the fourth one wants to extract the same topic into several tables with different schemas.

Too many cases, how to satisfy all? We decided to just allow to use own pluggable message converters, which'll do whatever the user wants. The converter is called for every Kafka message read, right before saving it in the destination database.

IlyaFaer:
@nicor88, as DLT is supposed to be an ultimate interface between different sources and destinations, we're overriding (the unique) Kafka's offsets and commits in favor of DLT's `state` system (which is the same for any source-destination pair).

Thus, if you're just testing your approach, you can use `full_refresh=True`, so every new extraction will be started from the scratch and loaded into new dataset, despite failures or successes. Example:
https://github.com/dlt-hub/verified-sources/blob/5d334bed5cdd149b1c5882410fd44b97850b0da1/tests/kafka/test_kafka.py#L117-L125

If you're not using `full_refresh`, and something failed during the extraction, you can still use [pipeline state reset](https://dlthub.com/docs/general-usage/state#reset-the-pipeline-state-full-or-partial) to drop all the recorded offsets and start from the beginning.

nicor88:
@IlyaFaer thanks for the clarifications - I totally missed the fact that the custom processor could be passed to the pipeline, and that indeed gives enough flexibility.

> If you're not using full_refresh, and something failed during the extraction, you can still use [pipeline state reset](https://dlthub.com/docs/general-usage/state#reset-the-pipeline-state-full-or-partial) to drop all the recorded offsets and start from the beginning.

In case of failure, I don't want to start from the beginning, but from the last "successful" committed offset for the configured consumer user group - does your implementation allow to have such behavior?

sources/kafka/__init__.py
burnash: @sultaniman @nicor88 do you mean that `batch_size` should be configurable via a [Config value](https://dlthub.com/docs/general-usage/credentials/configuration)?

IlyaFaer:
@nicor88, yes, we update offsets after every successfully extracted message, so by default the pipeline will continue the extraction from the moment it failed (incremental extraction is working exactly the same way, and it's tested successfully).

nicor88:
> yes, we update offsets after every successfully extracted message, so by default the pipeline will continue the extraction from the moment it failed (incremental extraction is working exactly the same way, and it's tested successfully).

is then the offset committed only once is successfully delivered to the destination? or at the moment of the extraction?
If offset are committed -at extraction time, and not at delivery time, you can have edge cases:
* messages extracted and offset committed
* destination delivery fails - e.g. wrong destination permissions/configuration 
* next run will start from the "committed" offset, but the messages processed from the previous batch/run won't be in the destination.

It's not clear to me when the offset commitment happen, as it make a big difference in case of recovery.

sources/kafka/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+    tracker = OffsetTracker(consumer, topics, dlt.current.resource_state(), start_from)
+    consumer.subscribe(topics)
+
+    # read messages up to the maximum offsets,
+    # not waiting for new messages
+    with closing(consumer):
+        while tracker.has_unread:
+            batch = []
+            for msg in consumer.consume(batch_size, timeout=1):
```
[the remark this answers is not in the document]
IlyaFaer: Done, pushing soon

sources/kafka/__init__.py
nicor88: I realised that `batch_size ` is indeed a configurable - therefore this point is addressed already.

nicor88:
@IlyaFaer overall make sense - 

I use the EL approach generally as well, the T is done downstream generally.

Let's put that way, I want to have the less "amount" of manual work possible when a failure occur, and they do occur for many reasons (destination overload, change of permissions on the destination and many more..)

I believe that using `start_from` could be a good way to solve this, instead of picking the offset from then last failure.


Anyhow, I'm chiming in, as I'm very interested to use such source, and I want to be sure that this implementation allow me to have the minimal features necessary to use it in production - and easy recover from failure is one of my MUST requirements.

I will give a shot (once time allow me), to try out this feature.

rudolfix:
@nicor88 not sure if that was 100% clear. at the end of extraction dlt remembers offsets of all topics/partitions and keeps it in the state. those offsets will be used on a next run to get new messages and that will happen only if the data is fully loaded to the destination (state is stored together with your data).

in other words if you have any error during extract load, the best course of action is to repeat the pipeline run (ie. run the script again). if you do it in an environment that does not preserve the local state then it is even better - we'll take the last good state from the destination and acquire all the messages that got failed again

btw. the mechanism above was the biggest complication and the reason to implement this source, really. any dumb implementation (with auto commit) is very simple (but as you correctly pointed out - does not work in production)

@IlyaFaer if all is good and @nicor88 does not have more doubts then let's merge this :)

nicor88:
@rudolfix I think that your explanation clarify even more my main concern - and it's all good from my side 💯 

Thanks for the implementation @IlyaFaer

burnash (approved):

rudolfix (approved):

PR #206, opened 2023-07-02
Add hubspot properties

steinitzu:
# Tell us what you do here

- [x] improving, documenting, customizing existing source (please link an issue or describe below)

# Relevant issue

Implements https://github.com/dlt-hub/verified-sources/issues/195

# More PR info

PR #311, opened 2023-12-22
fix(hubspot): provide default lists of properties

### LINKED ISSUE #305: Error in HubSpot Pipeline Execution: HTTPError 414 Client Error: URI Too Long
**Source name**
Name of the affected source as displayed by `dlt init` or present in `sources` folder of this repo.
[hubspot](https://github.com/dlt-hub/verified-sources/tree/master/sources/hubspot)
**Make sure that**
* [X] I checked the README in the source folder
* [X] If I have problems with the credentials I checked the [credentials docs](https://dlthub.com/docs/general-usage/credentials) and [troubleshooting](https://dlthub.com/docs/walkthroughs/run-a-pipeline#4-detect-and-handle-problems)

**Describe the bug**
Please describe the bug. If possible provide us:
Previously functional and used for loading deals data, the pipeline is now consistently erroring despite updating to DLT in the system, using new virtual environments, and creating a new pipeline.

1. hubspot()
2. "deals"
3. BigQuery
4. any stack traces and logs you can share

**To Reproduce**
1. Initiate a Hubspot verified source.
2. Update credentials.
4. Run example pipeline `load_crm_data()` with source as `hubspot().with_resources("deals")`

**Expected behavior**
The pipeline was expected to load deal data from Hubspot to BigQuery. It was working fine last month, but is now encountering errors.

**Stack traces and other evidence**

```bash
(venv) (base) radheshyaam@Heenas-MacBook-Air_hubspot_recheck % python3 hubspot_pipeline.py
/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/google/cloud/bigquery/client.py:562: UserWarning: Cannot create BigQuery Storage client, the dependency google-cloud-bigquery-storage is not installed.
  warnings.warn(
Traceback (most recent call last):
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/dlt/extract/pipe.py", line 696, in _get_source_item_current
    item = next(gen)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/hubspot_recheck/hubspot/__init__.py", line 126, in deals
    yield from crm_objects("deal", api_key, include_history)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/hubspot_recheck/hubspot/__init__.py", line 91, in crm_objects
    yield from fetch_data(CRM_OBJECT_ENDPOINTS[object_type], api_key, params=params)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/hubspot_recheck/hubspot/helpers.py", line 122, in fetch_data
    r = requests.get(url, headers=headers, params=params)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/dlt/sources/helpers/requests/retry.py", line 192, in <lambda>
    self.get = lambda *a, **kw: self.session.get(*a, **kw)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/requests/sessions.py", line 602, in get
    return self.request("GET", url, **kwargs)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 289, in wrapped_f
    return self(f, *args, **kw)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 379, in __call__
    do = self.iter(retry_state=retry_state)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 314, in iter
    return fut.result()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 382, in __call__
    result = fn(*args, **kwargs)
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/dlt/sources/helpers/requests/session.py", line 48, in request
    resp.raise_for_status()
  File "/Users/radheshyaam/PycharmProjects/hubspot_v2/venv/lib/python3.10/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 414 Client Error: URI Too Long for url: [https://api.hubapi.com/crm/v3/objects/deals?properties=age_contact_person%2Camount_in_home_currency%2Cbudget%2Cbudget_available%2Cbusiness_budget%2Cch](https://api.hubapi.com/crm/v3/objects/deals?properties=age_contact_person%2Camount_in_home_currency%2Cbudget%2Cbudget_available%2Cbusiness_budget%2Cchampion%2Cchamppion%2Ccms%2Ccommitted_seats%2Ccompany_expected_user%2Ccompany_footprint%2Ccompany_growth_potential%2Ccompany_size%2Ccompany_type%2Ccompelling_event%2Ccompeting_options%2Ccompetition%2Ccompettition%2Ccoms%2Ccsm_owner%2Ccustomer_segment%2Cdays_to_close%2Cdeal_currency_code%2Cdeal_source%2Cdeal_source_web%2Cdeal_status%2Cdecision_criteria%2Cdecision_criteria_stakeholder%2Cdecision_kriteria%2Cdecision_maker_involved%2Cdecision_process%2Cdecision_prozess%2Cdecision_timeline%2Cdemos_scheduled%2Cdigital_afinity%2Cdiscovery_call_completed%2Cdiscovery_call_scheduled%2Cdt%2Ceconomic_buyer%2Ceconomic_decision_maker%2Cemployee_buisness_budget%2Cemployee_location%2Cemployees%2Cemployees_filled_in_web_form%2Centered_deal_stage_at%2Cestimated_go_live%2Cestimated_monthly_recurring_revenue%2Cexec_reach_out_done_%2Cgmv_goal%2Cgmv_potential%2Chr_it%2Chr_tool%2Chs_acv%2Chs_all_assigned_business_unit_ids%2Chs_all_collaborator_owner_ids%2Chs_all_deal_split_owner_ids%2Chs_analytics_latest_source%2Chs_analytics_latest_source_company%2Chs_analytics_latest_source_contact%2Chs_analytics_latest_source_data_1%2Chs_analytics_latest_source_data_1_company%2Chs_analytics_latest_source_data_1_contact%2Chs_analytics_latest_source_data_2%2Chs_analytics_latest_source_data_2_company%2Chs_analytics_latest_source_data_2_contact%2Chs_analytics_latest_source_timestamp%2Chs_analytics_latest_source_timestamp_company%2Chs_analytics_latest_source_timestamp_contact%2Chs_analytics_source%2Chs_analytics_source_data_1%2Chs_analytics_source_data_2%2Chs_arr%2Chs_campaign%2Chs_closed_amount%2Chs_closed_amount_in_home_currency%2)
```
Please paste any stack traces, logs and screenshots. Mind that the issue is public!

**Running environment**
 - OS: [ Linux,  MacOs]
 - `dlt` Version [0.3.25]
 - Virtualization / Runner [ local & Composer]
 - Orchestrator / Deployment platform [Airflow]

**Additional context**

There were some new custom columns added to Hubspot in the past month by the user. So total columns in the deals is 299, so that must be causing this error.

IlyaFaer:
Fixes #305 

The error `414 URI too long` happens, because we request the list of all the possible properties for an object here:
https://github.com/dlt-hub/verified-sources/blob/a1cf60a718389f63b1b73b3a6e1c92432ee7c269/sources/hubspot/helpers.py#L165-L184

In some cases, object has so many properties, that their list joined by `,` becomes longer than 8000 of symbols, which causes `414`.

**Proposed solution**:
Instead of reading all the possible properties (most of which are very low-level and unlikely needed, like number of the broker, which processed the record), we create a default list of properties for every endpoint. The default lists are copied from the Hubspot `search` (see https://developers.hubspot.com/docs/api/crm/search#crm-objects), so for Hubspot users it should be an expected behavior.

However, if user wants, they can initiate a resource with their own list of properties to read, e.g.:
```python
load_info = pipeline.run(
            contacts(api_key="fake_key", include_history=True, props=["prop1", "prop2"])
        )
```

In case the user still wants to try to read all the possible properties, they can send `props=None` - in this case, we'll work the old way. The length of the properties list depends on many factors, so it can still work fine in many cases.

### DIFF
--- a/sources/hubspot/__init__.py
+++ b/sources/hubspot/__init__.py
@@ -38,11 +38,17 @@ from .helpers import (
     fetch_property_history,
 )
 from .settings import (
+    CRM_OBJECT_ENDPOINTS,
+    DEFAULT_COMPANY_PROPS,
+    DEFAULT_CONTACT_PROPS,
+    DEFAULT_DEAL_PROPS,
+    DEFAULT_PRODUCT_PROPS,
+    DEFAULT_TICKET_PROPS,
+    DEFAULT_QUOTE_PROPS,
+    OBJECT_TYPE_SINGULAR,
+    OBJECT_TYPE_PLURAL,
     STARTDATE,
     WEB_ANALYTICS_EVENTS_ENDPOINT,
-    OBJECT_TYPE_SINGULAR,
-    CRM_OBJECT_ENDPOINTS,
-    OBJECT_TYPE_PLURAL,
 )
 
 THubspotObjectType = Literal["company", "contact", "deal", "ticket", "product", "quote"]
@@ -83,9 +89,25 @@ def crm_objects(
     object_type: str,
     api_key: str = dlt.secrets.value,
     include_history: bool = False,
+    props: Sequence[str] = None,
 ) -> Iterator[TDataItems]:
     """Building blocks for CRM resources."""
-    props = ",".join(_get_property_names(api_key, object_type))
+    if props is None:
+        props = ",".join(_get_property_names(api_key, object_type))
+    else:
+        props = ",".join(props)
+
+    if len(props) > 2000:
+        raise ValueError(
+            (
+                "Your request to Hubspot is too long to process. "
+                "Maximum allowed query length is 2000 symbols, while "
+                f"your list of properties `{props[:200]}`... is {len(props)} symbols long. "
+                "Use the `props` argument of the resource to set the list "
+                "of properties to extract from the endpoint."
+            )
+        )
+
     params = {"properties": props, "limit": 100}
 
     yield from fetch_data(CRM_OBJECT_ENDPOINTS[object_type], api_key, params=params)
@@ -104,50 +126,62 @@ def crm_objects(
 
 @dlt.resource(name="companies", write_disposition="replace")
 def companies(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    props: Sequence[str] = DEFAULT_COMPANY_PROPS,
 ) -> Iterator[TDataItems]:
     """Hubspot companies resource"""
-    yield from crm_objects("company", api_key, include_history=False)
+    yield from crm_objects("company", api_key, include_history=False, props=props)
 
 
 @dlt.resource(name="contacts", write_disposition="replace")
 def contacts(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    props: Sequence[str] = DEFAULT_CONTACT_PROPS,
 ) -> Iterator[TDataItems]:
     """Hubspot contacts resource"""
-    yield from crm_objects("contact", api_key, include_history)
+    yield from crm_objects("contact", api_key, include_history, props)
 
 
 @dlt.resource(name="deals", write_disposition="replace")
 def deals(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    props: Sequence[str] = DEFAULT_DEAL_PROPS,
 ) -> Iterator[TDataItems]:
     """Hubspot deals resource"""
-    yield from crm_objects("deal", api_key, include_history)
+    yield from crm_objects("deal", api_key, include_history, props)
 
 
 @dlt.resource(name="tickets", write_disposition="replace")
 def tickets(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    props: Sequence[str] = DEFAULT_TICKET_PROPS,
 ) -> Iterator[TDataItems]:
     """Hubspot tickets resource"""
-    yield from crm_objects("ticket", api_key, include_history)
+    yield from crm_objects("ticket", api_key, include_history, props)
 
 
 @dlt.resource(name="products", write_disposition="replace")
 def products(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    props: Sequence[str] = DEFAULT_PRODUCT_PROPS,
 ) -> Iterator[TDataItems]:
     """Hubspot products resource"""
-    yield from crm_objects("product", api_key, include_history)
+    yield from crm_objects("product", api_key, include_history, props)
 
 
 @dlt.resource(name="quotes", write_disposition="replace")
 def quotes(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    props: Sequence[str] = DEFAULT_QUOTE_PROPS,
 ) -> Iterator[TDataItems]:
     """Hubspot quotes resource"""
-    yield from crm_objects("quote", api_key, include_history)
+    yield from crm_objects("quote", api_key, include_history, props)
 
 
 @dlt.resource
--- a/sources/hubspot/__init__.py
+++ b/sources/hubspot/__init__.py
@@ -96,7 +96,7 @@ def hubspot(
         yield from crm_objects(
             "company",
             api_key,
-            include_history=False,
+            include_history=include_history,
             props=props,
             include_custom_props=include_custom_props,
         )
--- a/sources/hubspot/__init__.py
+++ b/sources/hubspot/__init__.py
@@ -56,33 +56,102 @@ THubspotObjectType = Literal["company", "contact", "deal", "ticket", "product", 
 
 @dlt.source(name="hubspot")
 def hubspot(
-    api_key: str = dlt.secrets.value, include_history: bool = False
+    api_key: str = dlt.secrets.value,
+    include_history: bool = False,
+    global_props: Sequence[str] = None,
 ) -> Sequence[DltResource]:
     """
-    A DLT source that retrieves data from the HubSpot API using the specified API key.
+    A DLT source that retrieves data from the HubSpot API using the
+    specified API key.
 
-    This function retrieves data for several HubSpot API endpoints, including companies, contacts, deals,
-    tickets, products and web analytics events. It returns a tuple of Dlt resources, one for each endpoint.
+    This function retrieves data for several HubSpot API endpoints,
+    including companies, contacts, deals, tickets, products and web
+    analytics events. It returns a tuple of Dlt resources, one for
+    each endpoint.
 
     Args:
-        api_key (str, optional): The API key used to authenticate with the HubSpot API. Defaults to dlt.secrets.value.
-        include_history: Whether to load history of property changes along with entities. The history entries are loaded to separate tables.
+        api_key (Optional[str]):
+            The API key used to authenticate with the HubSpot API. Defaults
+            to dlt.secrets.value.
+        include_history (Optional[bool]):
+            Whether to load history of property changes along with entities.
+            The history entries are loaded to separate tables.
+        global_props (Optional[Sequence[str]]):
+            List of properties to be requested in every resource included
+            into `hubspot`.
 
     Returns:
-        tuple: A tuple of Dlt resources, one for each HubSpot API endpoint.
+        Sequence[DltResource]: Dlt resources, one for each HubSpot API endpoint.
 
     Notes:
-        This function uses the `fetch_data` function to retrieve data from the HubSpot CRM API. The API key
-        is passed to `fetch_data` as the `api_key` argument.
+        This function uses the `fetch_data` function to retrieve data from the
+        HubSpot CRM API. The API key is passed to `fetch_data` as the
+        `api_key` argument.
     """
-    return [
-        companies(include_history=include_history),
-        contacts(include_history=include_history),
-        deals(include_history=include_history),
-        tickets(include_history=include_history),
-        products(include_history=include_history),
-        quotes(include_history=include_history),
-    ]
+
+    global_props = global_props or []
+
+    @dlt.resource(name="companies", write_disposition="replace")
+    def companies(
+        api_key: str = api_key,
+        include_history: bool = include_history,
+        props: Sequence[str] = DEFAULT_COMPANY_PROPS,
+    ) -> Iterator[TDataItems]:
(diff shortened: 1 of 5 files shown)


sources/hubspot/__init__.py
```diff
@@ -83,9 +89,25 @@ def crm_objects(
...
     api_key: str = dlt.secrets.value,
     include_history: bool = False,
+    props: Sequence[str] = None,
 ) -> Iterator[TDataItems]:
     """Building blocks for CRM resources."""
-    props = ",".join(_get_property_names(api_key, object_type))
+    if props is None:
+        props = ",".join(_get_property_names(api_key, object_type))
+    else:
+        props = ",".join(props)
```
IlyaFaer, line 98: If user set `props=None`, request all the possible properties

tests/hubspot/test_hubspot_source.py
```diff
@@ -139,27 +141,26 @@ def fake_get(url: str, *args, **kwargs) -> Any:  # type: ignore[no-untyped-def]
...
+    assert m.call_count == 2
 
     # Check that API is called with all properties listed
     m.assert_has_calls(
         [
-            call(
-                urljoin(BASE_URL, "/crm/v3/properties/contacts"),
-                headers=ANY,
-                params=None,
-            ),
```
IlyaFaer, line 153: This call is no longer used, as default list of properties is used

IlyaFaer:
@rudolfix, okay, I think I got it. Pushed the code.

There is one hubspot test failing - it's not related, I saw these failures before started the task. After I update the docs, I can take a look at it, if needed. It seems like a test data cleanup error, not functionality.

sources/hubspot/__init__.py
```diff
@@ -38,54 +38,169 @@
...
+        api_key: str = api_key,
+        include_history: bool = include_history,
+        props: Sequence[str] = DEFAULT_COMPANY_PROPS,
+    ) -> Iterator[TDataItems]:
+        """Hubspot companies resource"""
+        yield from crm_objects(
+            "company",
+            api_key,
+            include_history=False,
+            props=props if props is None else props + global_props,  # type: ignore
```
IlyaFaer, line 105: If the user set `props` to `None`, it means they want to read all the possible properties, so there is no need to consider `global_props` at all.

rudolfix: good idea. must be documented!

rudolfix (changes requested):
This looks good and stuff is in place! I have one more request? can we automatically add all the user-defined properties to all endpoints by default?

we could use a global flag to disable it. my take is that our users will love it.

here's how they are added for contacts: https://legacydocs.hubspot.com/docs/methods/contacts/v2/create_contacts_property

they will be listed in `_get_property_names` I think there's a flag `hubspotDefined` (but I'm just guessing here)

rudolfix (commented):
the test is failing because something got added to the test account. you'll need to update the asserts in the test.

tests/hubspot/test_hubspot_source.py
```diff
@@ -201,8 +252,8 @@ def test_all_resources(destination_name: str) -> None:
     # Check history tables
     # NOTE: this value is increasing... maybe we should start testing ranges
     assert load_table_counts(pipeline, *history_table_names) == {
-        "contacts_property_history": 17226,
-        "deals_property_history": 14349,
+        "contacts_property_history": 3616,
+        "deals_property_history": 3662,
```
IlyaFaer, line 256: As we no longer request all the fields, the amount of history records reduced

tests/hubspot/test_hubspot_source.py
```diff
@@ -211,14 +262,15 @@ def test_all_resources(destination_name: str) -> None:
             list(row)
             for row in client.execute_sql(
                 """
-                SELECT ch.property_name, ch.value, ch.source_type, ch.source_type, ch.timestamp
+                SELECT ch.property_name, ch.value__v_text, ch.source_type, ch.timestamp
```
IlyaFaer, line 265: Not sure what is the `value` field. I tried to extract data (not through tests, but with the pipeline examples, and the value is actually in `value__v_text`).

IlyaFaer:
@rudolfix, there is no easy way in Hubspot to read only custom properties, as in new versions of the API they dropped most of the properties filtering functionality. However, in the newest version, all the Hubspot driven properties have prefix `hs_` (while for custom properties this prefix is restricted), so we can read all the properties and filter them before sending a request.

Also, I added constants, because all this properties stuff becomes too complex.

tests/hubspot/test_hubspot_source.py
rudolfix: yeah this means that we are getting inconsistent types and `dlt` creates variant columns as value can be integer and then we get string value.

I expect we'll get a request to pivot history table. so we actually generate separate columns for all possible properties ie `email_value). I think we should ignore that for now

sources/hubspot/__init__.py
```diff
@@ -38,54 +38,174 @@
...
 
 def crm_objects(
     object_type: str,
     api_key: str = dlt.secrets.value,
     include_history: bool = False,
+    props: Sequence[str] = None,
 ) -> Iterator[TDataItems]:
     """Building blocks for CRM resources."""
-    props = ",".join(_get_property_names(api_key, object_type))
+    if props == ALL:
```
rudolfix, line 190: shouldn't all be a tuple? `("ALL",)`. props are a sequence? also inter should complain here

sources/hubspot/__init__.py
```diff
@@ -38,54 +38,174 @@
...
+        props = ",".join(_get_property_names(api_key, object_type))
+    elif props == CUSTOM_ONLY:
```
rudolfix, line 192: this is almost good. what I wanted is to get the custom props on top of `props` passed by the user. so defines the hubspot props but custom come automatically. let's do a boolean flag to include them

rudolfix (changes requested):
OK let's iterate once more :)

tests/hubspot/test_hubspot_source.py
IlyaFaer: Well, the tests will not pass with `value`. That's why I edited it.

sources/hubspot/__init__.py
IlyaFaer: Well, actually, I against it. That's a strange value `("ALL",)`, like there is a property `ALL`. Users don't have to think about constants values, that's the idea of a constant - its name says everything the user need to now. Anyway, it's a tiny thing. Fixed.

sources/hubspot/__init__.py
IlyaFaer: Not sure I understand. What are the use cases then?

`hubspot()` - return all custom properties
`hubspot(only_custom=False)` - return properties from the default `search` lists
`hubspot(only_custom=False, props=["prop1", "prop2"]` - return only properties set by the user
`hubspot(only_custom=False, props=ALL)` - return all the possible properties

Like that? It seems quite complex - two args to control actually the same thing, and the use cases are not very clear at the first glance.

With the current version the use cases are:

`hubspot()` - return properties from the default `search` lists
`hubspot(props=ALL)` - return all the possible properties
`hubspot(props=CUSTOM_ONLY)` - return all custom properties
`hubspot(props=["prop1", "prop2"])` - return only the requested properties

Actually, if we provide an ability to request all the custom properties, then default lists of properties don't make much sense anymore. So, as I see it, it should be like this then:

`hubspot()` - return all the custom properties
`hubpost(props=ALL)` - return all the possible properties
`hubspot(props=["prop1", "prop2"])` - return only the given properties

And those default lists of properties are not needed at all.

IlyaFaer: O-o-oh, or maybe you want to combine all custom props to whatever the user passes into the `props` arg? Like that:

`hubspot()` - return all the custom properties
`hubspot(props=[...])` - all custom props + `[...]`?
`hubspot(props=[...], add_custom=False)` - only `[...]` is read?
`hubspot(add_custom=False)` - what to return in this case then?

rudolfix: yeah! that was what I mean thx :)

sources/hubspot/__init__.py
rudolfix: Heh I was not clear here. I suggested exactly what you did: keep the constant but the constant was not a tuple in `settings` which you fixed.

rudolfix (dismissed):
@IlyaFaer  this is really good now!
@AstrakhantsevaAA and @burnash  please take a look. we do not try to fetch all possible fields from CRM - just certain defaults and all custom properties

we can merge that at any moment,

sources/hubspot/__init__.py
```diff
@@ -38,54 +38,181 @@
...
+        api_key: str = api_key,
+        include_history: bool = include_history,
+        props: Sequence[str] = DEFAULT_COMPANY_PROPS,
+        include_custom_props: bool = True,
+    ) -> Iterator[TDataItems]:
+        """Hubspot companies resource"""
+        yield from crm_objects(
+            "company",
+            api_key,
+            include_history=False,
```
AstrakhantsevaAA, line 99: include_history always is False for companies?

AstrakhantsevaAA (commented):
I like it! I have only one question (see `companies` resource).
Also I would replace all these function with dlt.resource() to create them [dynamically](https://dlthub.com/docs/general-usage/source#create-resources-dynamically), but we can do it later or keep it like this at all.

AstrakhantsevaAA:
And don't forget to update the documentation here: https://github.com/dlt-hub/dlt/pull/857

sources/hubspot/__init__.py
IlyaFaer: It was done without any explanations here: https://github.com/dlt-hub/verified-sources/pull/206
Looks like a testing change, which was not reverted before committing.

IlyaFaer:
I think it's better to merge it already. It fixes an error - errors should not stay in code for long time, especially considering it's a widespread error and it can't be walked around.

Other changes can be done later.

AstrakhantsevaAA (approved):

PR #313, opened 2023-12-29
Implement a basic generic API client

burnash:
# Tell us what you do here

This PR implements a basic generic API client. Mostly by trying to generilize pagination and auth logic which many of the sources in this repo reimplement. The goal is to speed up sources development by unifying some common functionality. At the same time I'm trying to design the client's API to be as close to Requests as possible.

Relevant work: https://github.com/dlt-hub/dlt-init-openapi/pull/2

### Original 
```python
import dlt
from dlt.sources.helpers import requests

@dlt.resource(
    table_name="issues",
    write_disposition="merge",
    primary_key="id",
)
def get_issues(
    updated_at=dlt.sources.incremental("updated_at", initial_value="1970-01-01T00:00:00Z")
):
    # NOTE: we read only open issues to minimize number of calls to the API. There's a limit of ~50 calls for not authenticated Github users
    url = f"https://api.github.com/repos/dlt-hub/dlt/issues?since={updated_at.last_value}&per_page=100&sort=updated&directions=desc&state=open"

    while True:
        response = requests.get(url)
        response.raise_for_status()
        yield response.json()

        # get next page
        if "next" not in response.links:
            break
        url = response.links["next"]["url"]

pipeline = dlt.pipeline(
    pipeline_name="github_issues_merge",
    destination="duckdb",
    dataset_name="github_data_merge",
)
load_info = pipeline.run(get_issues)
print(load_info)
```

### Using RESTClient
```python
import dlt
from rest_api import RESTClient

github_client = RESTClient(base_url="https://api.github.com")


@dlt.resource(
    table_name="issues",
    write_disposition="merge",
    primary_key="id",
)
def get_issues(
    updated_at=dlt.sources.incremental(
        "updated_at", initial_value="1970-01-01T00:00:00Z"
    )
):
    for page in github_client.paginate(
        "/repos/dlt-hub/dlt/issues",
        params={
            "since": updated_at.last_value,
            "per_page": 100,
            "sort": "updated",
            "direction": "desc",
            "state": "open",
        },
    ):
        yield page


# The rest of the pipeline remains the same
pipeline = dlt.pipeline(
    pipeline_name="github_issues_merge",
    destination="duckdb",
    dataset_name="github_data_merge",
)
load_info = pipeline.run(get_issues)
print(load_info)
```


rudolfix:
@burnash 

This is very helpful. At least from my side it makes way clearer how to design code here. The idea of request-like helper that has pagination, auth and other things built-in is good.

**api**
basically I see two components

1. paginated requests which has similar interface to requests but instead of returning Response it retruns a generator, yielding pages of data

on top of usual `requests` arguments it also accepts
- authenticator instance (we already have auth argument)
- paginator instance
- response unwrapper (extracts page of data from response)
- incremental instance if endpoint is incremental
- incremental param to connect incremental instance to

the returned generator may be passed directly to `dlt.resource` (I think it will be generator function, not generator so you can bind incremental to it)

ie.
```
paginated = paginated_request(url, pagination=json_paginator(), auth=bearer(api_secret), incremental=dlt.source.incremental("updated_at"), incremental_param="sice")

resource = dlt.resource(paginated.data, name="...", primary_key="...")
```

where paginated.data is a generator function ie.
```
def data(self, last_value: dlt.source.incremental[Any] = None):
	... # main loop happens here
```

I'd strive move as much logic into a common code and make paginators and authenticators as simple as possible. For example paginator could get a last response and a prepared request and modify it to get next page. we could even make it more abstract and pass header, paramas etc.


2. declarative part. it is similar to Alex to EndpointFactory but also adds definitions of the endpoints. why we need it:
- to declare most common behavior. we hope 90% of apis can be just declared using starndard auth and paginators
- to easily add new endpoints
- to further ease using of incremental

To declare an endpoint we basically provide kwargs to the paginated_request. the declaration can also take default authenticator and paginator

then in the `dlt.source` user can enumerate all paginated_requests and create dlt.resources out of them. there should be also a method to get a dynamic endpoint out of it (then it works like Alex's factory)

**Approach**
Currently you try to fit a tool into existing code. So you get those complicated conditional paginators serving many different endpoint types. All try something different: let's build a decent API and then fit our verified sources. starting with simple stuff like pokemon and then going for zendesk..

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+    @abstractmethod
+    def prepare_next_request_args(
+        self, url: str, params: Optional[Dict[str, Any]], json: Optional[Dict[str, Any]]
+    ) -> Tuple[Optional[str], Optional[Dict[str, Any]], Optional[Dict[str, Any]]]:
+        """
+        Prepare the arguments for the next API request based on the current state of pagination.
+
+        Subclasses must implement this method to update the request arguments appropriately.
+
+        Args:
```
sultaniman, line 52: I love this style of docstrings

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+            url (str): The original URL used in the current API request.
+            params (Optional[Dict[str, Any]]): The original query parameters used in the current API request.
+            json (Optional[Dict[str, Any]]): The original JSON body of the current API request.
+
+        Returns:
+            tuple: A tuple containing the updated URL, query parameters, and JSON body to be used
+                for the next API request. These values are used to progress through the paginated data.
+        """
+        ...
```
sultaniman, line 61: Should we raise exception or return not implemented error?

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        Args:
+            links_next_key (str, optional): The key (rel ) in the 'Link' header
+                that contains the next page URL. Defaults to 'next'.
+        """
+        super().__init__()
+        self.links_next_key = links_next_key
+
+    def update_state(self, response: Response) -> None:
+        self.next_reference = response.links.get(self.links_next_key, {}).get("url")
```
sultaniman, line 100: Wow I didn't know about `requests.links`.

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+    def extract_records(self, response: Response) -> Any:
+        return response.json()
+
+
+class JSONResponsePaginator(BaseNextUrlPaginator):
+    """A paginator that uses a specific key in the JSON response to find
+    the next page URL.
+    """
+
+    def __init__(self, next_key: str = "next", records_key: str = "results"):
```
sultaniman, line 111: I think we should give option of json path here because I believe someone will come asking how can they extract next link from nested structure maybe this [jmespath-py](https://github.com/jmespath/jmespath.py) is good?

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+class BearerTokenAuth(AuthBase):
+    def __init__(self, token: str):
+        self.token = token
+
+    def __call__(self, request):
+        request.headers["Authorization"] = f"Bearer {self.token}"
+        return request
+
+
+def join_url(base_url: str, path: str) -> str:
```
sultaniman, line 139: should we use [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) instead?

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+        headers: Optional[Dict[str, str]] = None,
+        auth: Optional[AuthBase] = None,
+        paginator: Optional[BasePaginator] = None,
+    ) -> None:
+        self.base_url = base_url
+        self.headers = headers
+        self.auth = auth
+        self.paginator = paginator if paginator else HeaderLinkPaginator()
+
+    def make_request(self, path="", method="get", params=None, json=None):
```
sultaniman, line 169: I think we need to start [requests.session](https://requests.readthedocs.io/en/latest/user/advanced/) here because they can be pre-configured with authentication strategy, custom headers etc.

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+
+        response = requests.request(
+            method=method,
+            url=url,
+            headers=self.headers,
+            params=params if method.lower() == "get" else None,
+            json=json if method.lower() in ["post", "put"] else None,
+            auth=self.auth,
+        )
+        response.raise_for_status()
```
sultaniman, line 183: should we also allow not to fail for example cases when users want to get successful results and process erred results later?

sultaniman: This will probably require some sort of monadic type `Result`

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+        return response
+
+    def get(self, path="", params=None):
+        return self.make_request(path, method="get", params=params)
+
+    def post(self, path="", json=None):
```
sultaniman, line 189: maybe we can also use typing override to define signature variances?

sources/api_client.py
```diff
@@ -0,0 +1,217 @@
...
+        return self.make_request(path, method="post", json=json)
+
+    def paginate(
+        self,
+        path: str = "",
+        method: str = "get",
```
sultaniman, line 195: should we define custom `typing.Literal` to denote methods?

sources/zendesk/helpers/paginators.py
```diff
@@ -0,0 +1,54 @@
...
+    def prepare_next_request_args(self, url, params, json):
+        params = params or {}
+        params["page[size]"] = settings.PAGE_SIZE
+
+        return self.next_reference, params, json
+
+
+class StreamPaginator(JSONResponsePaginator):
+    def get_next_page_url(self, response_json):
+        if not response_json["end_of_stream"]:
```
sultaniman, line 25: should we use `"end_of_stream" not in response_json`?

sultaniman (commented):
Looks cool, I also wonder if pagination strategy should also take params from `.dlt/config.yml`?

sources/notion/__init__.py
```diff
@@ -27,15 +30,24 @@ def notion_databases(
...
+        base_url=API_URL,
+        headers=DEFAULT_HEADERS,
+        auth=BearerTokenAuth(api_key),
+        paginator=NotionPaginator(),
+    )
 
     if database_ids is None:
-        search_results = notion_client.search(
-            filter_criteria={"value": "database", "property": "object"}
+        search_results = notion_client.paginate(
```
francescomucio, line 41: I guess also line 56 needs to be changed to use `paginate()` to get the databases

sources/rest_api/client.py
```diff
@@ -0,0 +1,96 @@
...
+        paginator (Optional[BasePaginator]): A paginator object for handling API pagination.
+            Note that this object will be deepcopied for each request to ensure that the
+            paginator state is not shared between requests.
+    """
+
+    def __init__(
+        self,
+        base_url: str,
+        headers: Optional[Dict[str, str]] = None,
+        auth: Optional[AuthBase] = None,
```
sultaniman, line 28: I think typing for this should be as `Optional[Type[AuthBase]]` or `AuthBase[T]` and probably same for pagination abstractions

sources/rest_api/detector.py
```diff
@@ -0,0 +1,82 @@
+from dlt.sources.helpers.requests import Response
+
+from .paginators import HeaderLinkPaginator, JSONResponsePaginator
+
+RECORD_KEY_PATTERNS = {'data', 'items', 'results', 'entries'}
+NEXT_PAGE_KEY_PATTERNS = {'next', 'nextpage', 'nexturl'}
+
+
+def find_records_key(dictionary, path=None):
```
sultaniman, line 9: Maybe we can restrict `dictionary` using typing?
image
https://replit.com/@sultaniman/restricted-dict-keys-typing?v=1

sources/rest_api/detector.py
```diff
@@ -0,0 +1,82 @@
...
+    if path is None:
+        path = []
+
+    for key, value in dictionary.items():
+        # Direct match
+        if key in RECORD_KEY_PATTERNS:
+            return path + [key]
```
sultaniman, line 16: should we copy list instead like `[*path, key]`

sources/rest_api/client.py
```diff
@@ -0,0 +1,105 @@
...
+    ) -> None:
+        self.base_url = base_url
+        self.headers = headers
+        self.auth = auth
+        self.paginator = paginator if paginator else UnspecifiedPaginator()
+
+    def make_request(self, path="", method="get", params=None, json=None):
+        if path.startswith("http"):
+            url = path
+        else:
```
sultaniman, line 41: I think we should do some sort of safety checks for example, what if the path still has some other scheme, can this lead to security risks, should we only allow `http/s` scheme and also validate it in else branch to prevent such things? It is hypothetical but some companies probably can be indirectly targeted for RCE.

sultaniman: I see we have `dlt.requests` session mount calls https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/helpers/requests/retry.py#L239-L240
```
        session.mount("http://", self._adapter)
        session.mount("https://", self._adapter)
```

sultaniman: Then imo this should be fine?

sources/rest_api/detector.py
```diff
@@ -0,0 +1,95 @@
...
+    return None
+
+
+def json_links_detector(response: Response):
+    dictionary = response.json()
+
+    records_key = find_records_key(dictionary)
+
+    if not records_key:
+        return None
```
sultaniman, line 66: should we continue to check the result for `find_next_page_key`?

sources/rest_api/client.py
```diff
@@ -0,0 +1,105 @@
...
+                if paginator is None:
+                    raise ValueError(
+                        "No suitable paginator found for the API response."
+                    )
+                else:
+                    logger.info(f"Detected paginator: {paginator.__class__.__name__}")
+
+            yield paginator.extract_records(response)
+
+            paginator.update_state(response)
```
sultaniman, line 101: Since paginator knows about Response object should we pass `make_request` into it or maybe make similar method in them like so incrementing and state update is hidden?

```py
with paginator.paginate(path=path, method=method, params=params, json=json) as pager:
    while pager.has_next_page:
         response = self.make_request(
             path=path, method=method, params=params, json=json
        ) # maybe let paginator make next request?
        yield paginator.extract_records()
```

sultaniman: Also exception handling logic will be localized inside paginator

sources/notion/helpers/paginator.py
```diff
@@ -0,0 +1,14 @@
+from dlt.sources.helpers.requests import Response
+from rest_api import JSONResponsePaginator
+
+class NotionPaginator(JSONResponsePaginator):
+    def __init__(self, cursor_key='next_cursor', records_key='results'):
+        super().__init__(next_key=cursor_key, records_key=records_key)
+
+    def prepare_next_request_args(self, url, params, json):
+        json = json or {}
```
sultaniman, line 9: if someone will debug and import json module I think this might be inconvenient.

sources/rest_api/utils.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import Any, Dict
+
+
+def join_url(base_url: str, path: str) -> str:
+    if not base_url.endswith("/"):
+        base_url += "/"
+    return base_url + path.lstrip("/")
+
+
+def create_nested_accessor(path):
```
rudolfix, line 12: should be a part of default record extractor

sources/rest_api/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    if isinstance(path, (list, tuple)):
+        return lambda d: reduce(getitem, path, d)
+    return lambda d: d.get(path)
+
+
+def remove_key(d, key):
+    return {k: v for k, v in d.items() if k != key}
+
+
+def deep_merge(a: Dict[str, Any], b: Dict[str, Any]) -> Dict:
```
rudolfix, line 22: we have tested deep merge in dlt.common.utils

sources/rest_api/client.py
```diff
@@ -0,0 +1,109 @@
...
+            >>> for page in client.paginate("/search", method="post", json={"query": "foo"}):
+            >>>     print(page)
+        """
+        paginator = copy.deepcopy(paginator if paginator else self.paginator)
+        while paginator.has_next_page:
+            response = self.make_request(
+                path=path, method=method, params=params, json=json
+            )
+
+            if isinstance(paginator, UnspecifiedPaginator):
```
rudolfix, line 91: autodetection code should be a part of `UnspecifiedPaginator`

sources/rest_api/client.py
rudolfix: IMO paginator should not touch making physical connections. for that we have this client class and our request session

sources/rest_api/client.py
rudolfix: I do not think `requests` should handle any other protocol?

sources/rest_api/client.py
```diff
@@ -0,0 +1,109 @@
...
+        paginator: Optional[BasePaginator] = None,
+        session: BaseSession = None
+    ) -> None:
+        self.base_url = base_url
+        self.headers = headers
+        self.auth = auth
+        self.session = session or client.session
+        self.paginator = paginator if paginator else UnspecifiedPaginator()
+
+    def make_request(self, path="", method="get", params=None, json=None):
```
rudolfix, line 42: we should IMO split this method into `_prepare_request` and `_make_request` and add a few "hook" methods for people to subclass easily.

in _prepare_request you create a new prepared request, pass it to authenticator and paginator for modification, IMO call an abstract method that people can implement for custom logic and return

in make request you do actual request

sources/rest_api/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+        response = self.session.request(
+            method=method,
+            url=url,
+            headers=self.headers,
+            params=params if method.lower() == "get" else None,
+            json=json if method.lower() in ["post", "put"] else None,
+            auth=self.auth,
+        )
+        response.raise_for_status()
```
rudolfix, line 61: instead call an abstract method that people can subclass. `response.raise_for_status()` should be default implementation

sources/rest_api/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+                # If no paginator is found, raise an error
+                if paginator is None:
+                    raise ValueError(
+                        "No suitable paginator found for the API response."
+                    )
+                else:
+                    logger.info(f"Detected paginator: {paginator.__class__.__name__}")
+
+            yield paginator.extract_records(response)
```
rudolfix, line 103: I like that Alex let's people to easily customize this:
```python
def parse_response(self, response: requests.Response) -> t.Iterable[dict]:
        """Parse the response and return an iterator of result records."""
        yield from extract_jsonpath(self.records_jsonpath, input=response.json())
```

rudolfix (changes requested):
this is my input for bulding blocks. TLDR;>
- copy the Authenticator code from Alex almost fully
- record extractor should be separated from paginator
- paginator is mostly fine. IMO we should use prepared requests in it the same way Authenticator does that
- detections should be part of paginators
- proposed a lot of abstract functions in client to easily change the behavior

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+    param_name: str
+    resolve_config: ResolveConfig
+
+
+class Endpoint(TypedDict, total=False):
+    path: str
+    method: str
+    params: Optional[Dict[str, Any]]
+    json: Optional[Dict[str, Any]]
+    paginator: Optional[PaginatorType]
```
rudolfix, line 92: Ideally we should also allow instance of paginator here (and maybe later a dict with paginator config)

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+
+
+class EndpointResource(TypedDict, total=False):
+    name: TTableHintTemplate[str]
```
rudolfix, line 96: we'll move that to core library. only Endpoint should stay here

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+    columns: TTableHintTemplate[TTableSchemaColumns]
+    primary_key: TTableHintTemplate[TColumnNames]
+    merge_key: TTableHintTemplate[TColumnNames]
+    incremental: Incremental[Any]
+    # schema_contract: TTableHintTemplate[TSchemaContract]
+    table_format: TTableHintTemplate[TTableFormat]
+
+
+class RESTAPIConfig(TypedDict):
+    client: ClientConfig
```
rudolfix, line 109: could we also configure type of the client?
```
client_class: Type[RESTClient]
```
by default

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+        "auth": create_auth(client_config.get("auth")),
+        "paginator": create_paginator(client_config.get("default_paginator")),
+    }
+
+
+def setup_incremental_object(request_params, incremental_config):
+    for key, value in request_params.items():
+        if isinstance(value, dlt.sources.incremental):
+            return value, key
+        if isinstance(value, dict) and value.get("type") == "incremental":
```
rudolfix, line 139: what is this `type` doing here?

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+class IncrementalConfig(TypedDict, total=False):
+    cursor_path: str
+    initial_value: str
+    param: str
+
+
+class EndpointConfig(TypedDict):
+    params: Dict[str, Any]
+    resource: ResourceConfig
+    incremental: Optional[IncrementalConfig]
```
rudolfix, line 67: should accept instance and config

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+    method: str
+    paginator: Optional[PaginatorType]
```
rudolfix, line 69: should accept instance and config

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+            )
+
+    return setup_incremental_object_from_config(incremental_config)
+
+
+def setup_incremental_object_from_config(config):
+    return (
+        (
+            dlt.sources.incremental(
+                config.get("cursor_path"), initial_value=config.get("initial_value")
```
rudolfix, line 154: there are more useful settings for incremental

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+        else (None, None)
+    )
+
+
+def make_parent_key_name(resource_name, field_name):
+    return f"_{resource_name}_{field_name}"
+
+
+@dlt.source
+def rest_api_source(config: RESTAPIConfig):
```
rudolfix, line 168: do not decorate. use `dlt.source` as function. allow for the custom name, section, schema, max table nesting and other params that decorator accepts

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+                    params[incremental_param] = incremental_object.last_value
+
+                yield from client.paginate(
+                    method=method,
+                    path=path,
+                    params=params,
+                    paginator=paginator,
+                )
+
+            resources[resource_name] = dlt.resource(
```
rudolfix, line 330: here you'll be able to pass client.paginate directly to resource

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+            def paginate_dependent_resource(
+                items,
+                method,
+                path,
+                params,
+                paginator,
+                param_name=param_name,
+                field_path=resolved_param.resolve_config.field_path,
+            ):
+                items = items or []
```
rudolfix, line 354: allow for items that are not lists

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+                paginator=paginator,
+            )
+
+        else:
+            predecessor = resources[resolved_param.resolve_config.resource_name]
+
+            param_name = resolved_param.param_name
+            request_params.pop(param_name, None)
+
+            def paginate_dependent_resource(
```
rudolfix, line 345: I wonder if this function could be top level and not inner? so maybe people can mock their own in worst case?

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+                paginator,
+                param_name=param_name,
+                field_path=resolved_param.resolve_config.field_path,
+            ):
+                items = items or []
+                for item in items:
+                    formatted_path = path.format(**{param_name: item[field_path]})
+                    parent_resource_name = resolved_param.resolve_config.resource_name
+
+                    parent_record = (
```
rudolfix, line 359: 👍

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,768 @@
...
+    - As a string: The string is interpreted as both the resource name
+        and its endpoint path.
+    - As a dictionary: The dictionary must include `name` and `endpoint`
+        keys. The `endpoint` can be a string representing the path,
+        or a dictionary for more complex configurations. If the `endpoint`
+        is missing the `path` key, the resource name is used as the `path`.
+    """
+    if isinstance(resource, str):
+        resource = {"name": resource, "endpoint": {"path": resource}}
+        return deep_merge(copy.deepcopy(default_config), resource)
```
rudolfix, line 410: better to use deep_merge on empty dict to get an initial copy. we do not want to clone objects. especially if for example someone wants to authenticate authenticator before passing

rudolfix (changes requested):
top level interface is almost good. we can cleanup the code and better integrate with the core lib

sources/rest_api/client.py
```diff
@@ -0,0 +1,252 @@
...
+    ) -> None:
+        self.base_url = base_url
+        self.headers = headers
+        self.auth = auth
+
+        if session:
+            self._validate_session_raise_for_status(session)
+            self.session = session
+        else:
+            self.session = Client(raise_for_status=False).session
```
steinitzu, line 73: I think we need to use `Client` obj as session cause session is not thread safe. The `session` property creates a thread-local session.   
Also a case for switching to httpx which afaik is thread safe and has the same interface as requests, that should let us do a direct subclass instead of a proxy

sources/rest_api/typing.py
```diff
@@ -0,0 +1,109 @@
...
+    TTableFormat,
+    TTableSchemaColumns,
+    TWriteDisposition,
+)
+
+PaginatorConfigDict = Dict[str, Any]
+PaginatorType = Union[Any, BasePaginator, str, PaginatorConfigDict]
+
+HTTPMethodBasic = Literal["get", "post"]
+HTTPMethodExtended = Literal["put", "patch", "delete"]
```
sultaniman, line 32: I think just for the sake of completeness and maybe to support weird APIs should we also add `options` and `head` methods?

sources/rest_api/auth.py
```diff
@@ -0,0 +1,174 @@
...
+
+        logger.debug(f"Obtaining token from {self.auth_endpoint}")
+
+        response = requests.post(self.auth_endpoint, headers=self.headers, data=data)
+        response.raise_for_status()
+
+        token_response = response.json()
+        self.token = token_response["access_token"]
+        self.token_expiry = pendulum.now().add(
+            seconds=token_response.get("expires_in", 3600)
```
sultaniman, line 154: should we make default expiry configurable and maybe <1h just in case if token leaks?

sources/rest_api/typing.py
```diff
@@ -0,0 +1,109 @@
...
+from dlt.common.schema.typing import (
+    TColumnNames,
+    # TSchemaContract,
+    TTableFormat,
+    TTableSchemaColumns,
+    TWriteDisposition,
+)
+
+PaginatorConfigDict = Dict[str, Any]
+PaginatorType = Union[Any, BasePaginator, str, PaginatorConfigDict]
```
sultaniman, line 29: Is `Any` intentional here?

sources/rest_api/client.py
rudolfix: OK I get it! @burnash so `Client` implements the same interface as `Session` but actually creates a Session per thread. so we should to drop `.session` here. please do a PR

sultaniman (dismissed):
🌟 🚀 🌟

rudolfix (changes requested):
TODO (for later)
1. "Endpoint" does not make it clear if single entity is returned or a list of entities. should we have "none" paginator to indicate that? or some boolean flag
2. What is the use of SinglePage paginator now? we have this `add_limit` which makes any paginator single page
3. I should be able to bind same incremental to start/since param (start_value) and to end/before param (end_value). 
4. I should be able to bind same transformer to many params when id is compound (this is less important)
5. paginators should be configspecs (dataclasses) so they can be configured and initialized like authentiators. we could use the same factory approach for both - I can do this one

6. `RESTAPIConfig` should accept an instance of DltResource here to allow completely code to be mixed with declared endpoint resources
```
resources: List[Union[str, EndpointResource]]
```
(this is less important)
7. several types in typing.py must go to dlt core ASAP. I can also handle that. we could also generate typed dicts automatically for all our configspecs.

sources/rest_api/__init__.py
```diff
@@ -0,0 +1,346 @@
...
+    Any,
+    Dict,
+    Tuple,
+    List,
+    Optional,
+    Generator,
+    Callable,
+    cast,
+)
+import graphlib  # type: ignore[import,unused-ignore]
```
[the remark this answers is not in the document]
burnash: It was added in 3.9. Let me know if there's a better way to silence mypy.
https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-type-ignore-comment-is-used-unused-ignore

sources/rest_api/paginators.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class OffsetPaginator(BasePaginator):
+    """A paginator that uses the 'offset' parameter for pagination."""
+
+    def __init__(
+        self,
+        initial_offset: int,
+        initial_limit: int,
+        offset_key: str = "offset",
```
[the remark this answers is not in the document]
burnash: Fixed in 3b7a0b611acf87c62efa92e6587a0a31e9d70360

sources/rest_api/paginators.py
```diff
@@ -0,0 +1,151 @@
...
+        return
+
+
+class OffsetPaginator(BasePaginator):
+    """A paginator that uses the 'offset' parameter for pagination."""
+
+    def __init__(
+        self,
+        initial_offset: int,
+        initial_limit: int,
```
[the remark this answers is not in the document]
burnash: Correct, I use this value as a limit increment. This implementation assumes total items count is available in the response. As it is in DRF's [LimitOffsetPagination](https://www.django-rest-framework.org/api-guide/pagination/#limitoffsetpagination). If the total is not available, one option is to keep incrementing offset until there are no records. Not sure if we really need to implement that in the first release.

sources/rest_api/paginators.py
```diff
@@ -0,0 +1,151 @@
...
+    def update_request(self, request: Request) -> None:
+        return
+
+
+class OffsetPaginator(BasePaginator):
+    """A paginator that uses the 'offset' parameter for pagination."""
+
+    def __init__(
+        self,
+        initial_offset: int,
```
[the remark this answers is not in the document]
burnash: Done

sources/rest_api/detector.py
```diff
@@ -0,0 +1,161 @@
...
+        "extras",
+        "headers",
+    ]
+)
+
+NEXT_PAGE_KEY_PATTERNS = frozenset(["next", "nextpage", "nexturl"])
+NEXT_PAGE_DICT_KEY_PATTERNS = frozenset(["href", "url"])
+
+
+def single_entity_path(path: str) -> bool:
```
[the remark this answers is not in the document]
burnash: Agreed. The proper way is to detect based on the shape of the data. I suggest to just remove this for now from the RESTClient code to upper level and review it later.

sources/rest_api/config_setup.py
```diff
@@ -0,0 +1,277 @@
...
+    return (
+        BearerTokenAuth(cast(TSecretStrValue, auth_config.get("token")))
+        if auth_config
+        else None
+    )
+
+
+def setup_incremental_object(
+    request_params: Dict[str, Any],
+    incremental_config: Optional[IncrementalConfig] = None,
```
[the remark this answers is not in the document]
burnash: Added in d63e30f6e4385c27f115c989669133d2fd328162

sources/rest_api/config_setup.py
```diff
@@ -0,0 +1,277 @@
...
+        paginator_class = get_paginator_class(paginator_type)
+        return paginator_class(**paginator_config)
+
+    return None
+
+
+def create_auth(
+    auth_config: Optional[Union[AuthConfig, AuthConfigBase]],
+) -> Optional[AuthConfigBase]:
+    if isinstance(auth_config, AuthConfigBase):
```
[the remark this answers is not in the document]
burnash: Added in 6dab45153c7d380c5946a76b94a77c795cc400fc

sources/rest_api/config_setup.py
```diff
@@ -83,15 +89,50 @@ def create_paginator(paginator_config: PaginatorType) -> Optional[BasePaginator]
...
         return auth_config
-    return (
-        BearerTokenAuth(cast(TSecretStrValue, auth_config.get("token")))
-        if auth_config
-        else None
-    )
+
+    if isinstance(auth_config, dict):
+        # Handle a shorthand auth configuration
+        if "token" in auth_config and len(auth_config) == 1:
```
burnash, line 99: @rudolfix I kept this shorthand for convenience but don't have a strong opinion how useful it could be. What do you think?

rudolfix: IMO it will be useful when we move to core library. then you'll do
```
for page in requests.paginate(auth="token"):
    ...
 
```
and actual token will come from secrets.toml
```
[sources.name.auth]
api_token="..."
``` 

we need aliases for all auths we support

burnash: The above works a bit different and is a shorthand for this version:

```py
"client": {
    "base_url": "https://api.github.com/repos/dlt-hub/dlt/",
    "auth": {
        "token": "...",
    },
},
```

For the other auth types I've used the OpenAPI-style definitions, based on what we have in configspecs. E.g.

```py
"client": {
    "base_url": "...",
    "auth": {
        "type": "http",
        "scheme": "bearer",        
    },
},
```

or 

```py
"client": {
    "base_url": "...",
    "auth": {
        "type": "apiKey",
        "name": "X-API-Key",        
         ...
    },
},
```

rudolfix (approved):
LGTM!

PR #315, opened 2024-01-02
test all sources every night

AstrakhantsevaAA:
# Tell us what you do here

# Relevant issue

issue #

# More PR info

We need to check all our verified sources sometimes. We test them, but only when create or change them.  They break silently all the time, and we find out about it from users or by accident.

Suggestion: we could have a nightly run, every night at 4am or so

### DIFF
--- a/.github/workflows/test_all_sources_nigthly.yml
+++ b/.github/workflows/test_all_sources_nigthly.yml
@@ -1,26 +1,21 @@
 
-name: test all sources (postgres and duckdb)
+name: test all sources (linux, duckdb)
 
 on:
   schedule:
   - cron: 0 0 * * *
 env:
-  ALL_DESTINATIONS: '["postgres", "duckdb"]'
+  ALL_DESTINATIONS: '["duckdb"]'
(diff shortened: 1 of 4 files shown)


burnash (approved):
Awesome work, @AstrakhantsevaAA

rudolfix:
@AstrakhantsevaAA run this only on duckdb on linux.

burnash:
@AstrakhantsevaAA one more thing: we need to exclude some tests from nightly tests. We can do that using custom pytest markers: https://docs.pytest.org/en/7.1.x/example/markers.html

VioletM:
@AstrakhantsevaAA maybe it makes sense to merge this even before all tests are up and running? This way we would see the stats on failing tests

AstrakhantsevaAA:
@burnash I skip tests without stable test account [here](https://github.com/dlt-hub/verified-sources/pull/379)
@VioletM now we're waiting for only Kafka test, so I think now it makes sense to merge it. 
@rudolfix OK

sultaniman (approved):

Issue #299, opened 2023-12-11
[csv + duckdb] reader


rudolfix:
# Quick source info
- Name of the source: [e.g. csv_reader]
- What is the data source: a csv files or a folder with csv files residing on any of fsspec filesystems (ie supported by `filesystem` source)

# Current Status
- [x] I plan to write it

# What source does/will do
The source is able to read a location with can be a single file or a folder with many csv files with the same columns. The user should be able to pass a list of such locations to the source and source should create a separate resource for each location that will extract desired csvs from that location. We may also need to provide resource (table names) for each location. Also see `sql_database` on how to modify resources after they are created (with `apply_hints`)

All locations must correspond to the same bucket. See `filesystem` source on how pass bucket and credentials.

We want to use duckdb as the parsing engine. There's a lot to figure out
1. Duckdb can read from fsspec locations now (but no idea how to pass credentials to it) https://duckdb.org/docs/guides/python/filesystems.html
2. Duckdb can read csvs in batches and yield arrow tables with partial data for given csv file. look in Python documentation and you'll find the correct method
3. Provide convenient interface to pass the config arguments from the source to csv reader in duckdb.
4. Allows for incremental loading via `filesystem` - based on last modified timestamp.

Demos:
We'll need to write a few demos:
1. getting csvs from a bucket from several locations, some of them files, some of them folders - using glob to specify
2. getting only the new files - with modification date
3. `apply_hints` on the resource to convert it into a merge resource with primary key

# Test account / test data
Please see the test data for `filesystem`, You will need access to all our supported buckets and filesystems in order to create additional test data (with cvs)
We want to run tests on all buckets - like we do for `filesystem`

# Additional context
You can implement first version as addon to `filesystem` source so you can use `filesystem` directly. There's already a demo of pandas based csv reader. It is a good starting point.

IlyaFaer:
The source is implemented, so the issue seems to be finished.

PR #319, opened 2024-01-12
feat(filesystem): implement a csv reader with duckdb engine

IlyaFaer:
Towards #299

A raw version of the source to discuss some details.

Test data in a form of CSV files and folders is required in different buckets. For now, I tested it on my local filesystem:

Untitled

Data from two CSV files is read and saved in the same table. As it works through the `filesystem` source, all the kinds of storages supported by `filesystem` are supposed to be working here as well. But as I don't have straight access to the buckets to create test CSV files, I had to use local file system.


sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        file_res = duckdb.sql(f"""SELECT * FROM read_csv_auto('{file["file_url"]}')""")
+        results += add_columns(file_res.columns, file_res.fetchall())
+
+        state["last_modified"] = max(file["modification_date"], state["last_modified"])
+
+    yield results
+
+
+@dlt.source
+def csv_reader(bucket: str, globs: List[str] = ("*",)) -> Iterable[TDataItem]:
```
IlyaFaer, line 45: The user can pass several `globs`. By default the most non-specific glob is used.

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        bucket (str): A bucket URL.
+        globs (Optional[List[str]]):
+            A list of glob patterns to match files.
+            Every glob will be extracted into a separate table.
+
+    Returns:
+        Iterable[TDataItem]:
+            Data items, read from the matched CSV files.
+    """
+    duckdb.register_filesystem(LocalFileSystem())
```
IlyaFaer, line 60: This item is very unpleasant - `duckdb` requires a filesystem registration. Looks like we need, looking at the bucket scheme, decide which filesystem to register.

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns:
+        Iterable[TDataItem]: Data items, read from the given CSV files.
+    """
+    state = dlt.current.resource_state()
+    start_from = state.setdefault("last_modified", pendulum.datetime(1970, 1, 1))
+
+    results = []
+    for file in files:
+        if file["modification_date"] <= start_from:
+            continue
```
IlyaFaer, line 34: Incremental
Files are not sorted by dates, and some storages doesn't even allow that, so we can't use the built-in incremental

sources/csv_reader/helpers.py
```diff
@@ -0,0 +1,16 @@
...
+
+    Returns:
+        List[Dict[str, Any]]:
+            A list of rows with column names.
+    """
+    result = []
+    for row in rows:
+        result.append(dict(zip(columns, row)))
+
+    return result
```
IlyaFaer, line 16: `duckdb` returns data from CSV in a form of tuples, without column names. It, of course, causes a very unpretty scheme of destination tables. I thought, it can be solved by turning tuples into dicts.

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    """A resource to extract data from the given CSV files.
+
+    Args:
+        files (List[FileItem]): A list of files to read.
+
+    Returns:
+        Iterable[TDataItem]: Data items, read from the given CSV files.
+    """
+    state = dlt.current.resource_state()
+    start_from = state.setdefault("last_modified", pendulum.datetime(1970, 1, 1))
```
sultaniman, line 34: can start_date be parameterized or will we actually have datastores with files from 70s?

sources/csv_reader/__init__.py
sultaniman: Should we pass filesystem via parameters and make it swappable, if not passed then maybe fallback to `LocalFileSystem` or otherwise just stop the pipeline?

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+    results = []
+    for file in files:
+        if file["modification_date"] <= start_from:
```
sultaniman, line 38: If some files are skipped then should we log the count of them or colect some sort of telemetry event?

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+            continue
+
+        file_res = duckdb.sql(f"""SELECT * FROM read_csv_auto('{file["file_url"]}')""")
```
rudolfix, line 42: please try `duckdb.read_csv()`. what I expect to happen there
1. you can possibly pass opened file as name parameter. we have a chance that any fsspec open file will work without registering filesystems in duck db
2. the method returns duckb relation object which you can read in batches. your code below will load all records in one go, this is not suitable for production

in other words: 
1. let's get reading and yielding batches straight
2. we should be able to yield arrow tables or arrow batches - this will also carry the schema to dlt and I think it is our preferred method
3. let's make sure we can use fsspec opened objects instead of providing names

if (3) is true we'll move this code to `filesystem` verified source (or merge the filesystem to the core library...) but let's do another iteration here

rudolfix (changes requested):
@IlyaFaer good start. check my comment in `read_csv`

sources/csv_reader/__init__.py
IlyaFaer: @rudolfix, I don't think the item 3 is that easy. Those objects returned from fsspec are just `dict`s. It's like meta description of a file, not the file itself. The usual way in fsspect to read a file is this:

```python
fs.open('/remote/data/myfile.csv')
```

So, we can, I guess: read a file meta info, then create a file-like object with `fs.open()` and pass it into the `read_csv()` function, but I'm not sure if it makes sense.

IlyaFaer: @rudolfix, ah, actually, no, it'll not work. `fsspec.open()` returns not a standard file-like object, so `duckdb` can't read it:

Untitled2

IlyaFaer: @rudolfix, a quite cumbersome piece of code may make it work:

```python
connection = duckdb.connect()

for file in files:
    if file["modification_date"] <= start_from:
        continue

    with fsspec.open(file["file_url"], mode="rb") as f:
        file_res = connection.read_csv(f)
        results += add_columns(file_res.columns, file_res.fetchall())
```

sources/csv_reader/__init__.py
sh-rp: i don't think so, this is regular incremental load behavior

IlyaFaer:
This PR requires another PR to be merged first: https://github.com/dlt-hub/dlt/pull/906

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+            file_data = connection.read_csv(f).to_arrow_table()
+
+            for batch in file_data.to_batches(max_chunksize=5000):
+                yield batch
+
+        state["last_modified"] = max(file["modification_date"], state["last_modified"])
+
+
+@dlt.source
+def csv_reader(bucket: str, globs: List[str] = ("*",)) -> Iterable[TDataItem]:
```
rudolfix, line 69: this will not be necessary after moving to `filesystem`

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    from sources.filesystem.helpers import (
+        AbstractFileSystem,
+        FilesystemConfigurationResource,
+        FileSystemCredentials,
+    )
+
+
+@dlt.resource(spec=FilesystemConfigurationResource)
+def read_location(
+    files: List[FileItemDict],
```
rudolfix, line 35: please allow for the following configuration (after converting to transformer, see other comments:)
- let people pick if they want to use arrow or json. item format. some people may not want arrow because it is a huge dependency (I know you had a json version, so we need to restore it... also allow to read in batches, there's support for that in duckdb)
- allow to pass chunk size
- allow to pass kwargs that will be forwarded to duckdb read_csv so people have access to all the options

rudolfix (changes requested):
it's so cool that passing fsspec file works with duck db! next steps:
- move it to `filesystem` source, see review
- you still do not read in batches really (see review)
- I think we need both json and arrow option when yielding items (see review :)

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+        if file["modification_date"] <= start_from:
+            continue
+
+        with fsspec.open(file["file_url"], mode="rb", **kwargs) as f:
+            file_data = connection.read_csv(f).to_arrow_table()
+
+            for batch in file_data.to_batches(max_chunksize=5000):
+                yield batch
+
+        state["last_modified"] = max(file["modification_date"], state["last_modified"])
```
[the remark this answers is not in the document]
IlyaFaer: But in the docs it says incremental takes the last value, not the biggest. If so, then, as files are not sorted, it's going to work incorrectly in this case.

sources/csv_reader/__init__.py
IlyaFaer: Okay, done

sources/csv_reader/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    start_from = state.setdefault("last_modified", pendulum.datetime(1970, 1, 1))
+
+    connection = duckdb.connect()
+
+    for file in files:
+        if file["modification_date"] <= start_from:
+            continue
+
+        with fsspec.open(file["file_url"], mode="rb", **kwargs) as f:
+            file_data = connection.read_csv(f).to_arrow_table()
```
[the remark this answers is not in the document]
IlyaFaer: A-ah, I though you only mean returning to DLT in batches. Okay, done

sources/filesystem/readers.py
```diff
@@ -74,6 +84,60 @@ def _read_parquet(
...
+            Additional keyword arguments to pass to the `read_csv()`.
+
+    Returns:
+        Iterable[TDataItem]: Data items, read from the given CSV files.
+    """
+    import duckdb
+    import fsspec  # type: ignore
+    import pendulum
+
+    config = FilesystemConfiguration(bucket, credentials)
```
rudolfix, line 119: you do not need that. FileItemDict has methods to (1) open referenced file (2) get pre-authorized filesystem, look what existing csv reader does:
```
    for file_obj in items:
        # Here we use pandas chunksize to read the file in chunks and avoid loading the whole file
        # in memory.
        with file_obj.open() as file:
            for df in pd.read_csv(file, **kwargs):
                yield df.to_dict(orient="records")
```
user `open()` below. it will be already authenticated. (and maybe we do not need this `dlt` extension)

tests/filesystem/test_filesystem.py
```diff
@@ -174,16 +177,23 @@ def _copy(item: FileItemDict):
...
+        "parquet_example",
+        "listing",
+        "csv_example",
+        "csv_duckdb_example",
     ) == {
         "jsonl_example": 1034,
         "parquet_example": 1034,
         "listing": 10,
         "csv_example": 1270,
+        "csv_duckdb_example": 1272,
```
rudolfix, line 196: why do we have two more records? maybe we read headers as data or vice versa? please check it out

rudolfix (changes requested):
OK this is not a WIP anymore - we are almost ready to merge :)
1. pls fix the review
2. pls add demo (can be super simple along `stream_and_merge_csv`)
3. pls document new reader in filesystem README

sources/csv_reader/__init__.py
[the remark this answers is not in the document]
rudolfix: yes, the docs are not clear. our incremental implementation uses min/max/user defined function to compare values and return the "bigger" one. in other words: user can define what last value means. implementation does not have any assumptions on the order.

the use case for incremental on last modification:
1. you have a folder with csv files and every day a new file is added (with an update / full refresh) but the folder is never cleared. dlt will process only the new file (from the last update)
2. you have a folder (ie. git repo with docs like ours) with md files and you vectorize them and load to weaviate. if you run pipeline everyday, dlt will return only the files that got modified from previous day

sources/filesystem/readers.py
IlyaFaer: A-ah, okay. When `fsspec` rejected a `FileItemDict` object with the comment "not a file-like" I didn't check that it actually openable, not just a dict. Done.

tests/filesystem/test_filesystem.py
IlyaFaer: It's because a header in one of the files includes commas - it's not parsed correctly. By the very basic CSV standard, such lines must be escaped with double quotes (otherwise, many parsers will not deal with it). Added them:

image

rudolfix:
@IlyaFaer could you also look at this: https://clickhouse.com/docs/en/getting-started/example-datasets/nyc-taxi it is a taxi dataset as zip csv. it is quite large. pls make sure you can ingest it with duckdb reader
1. make sure we ingest zip streams (I think duckdb detects compression)
2. you do not need to load it all
3. you may add this as a duckb example I asked above

sources/filesystem/readers.py
```diff
@@ -74,6 +77,53 @@ def _read_parquet(
...
+
+    Returns:
+        Iterable[TDataItem]: Data items, read from the given CSV files.
+    """
+    import duckdb
+    import pendulum
+
+    read_csv_kwargs = read_csv_kwargs or {}
+
+    state = dlt.current.resource_state()
```
rudolfix, line 108: you really do not need it here. you can enable this behavior on the `filesystem` resource from which this transformer takes data. so the filtering of already processed files happens before transformer is called. here's an example:
```
# here we modify filesystem resource so it will track only new csv files
    # such resource may be then combined with transformer doing further processing
    new_files = filesystem(bucket_url=TESTS_BUCKET_URL, file_glob="csv/*")
    # add incremental on modification time
    new_files.apply_hints(incremental=dlt.sources.incremental("modification_date"))
    load_info = pipeline.run((new_files | read_csv()).with_name("csv_files"))
```
btw. this behavior is disabled by default

rudolfix (changes requested):

tests/filesystem/test_filesystem.py
```diff
@@ -174,16 +177,23 @@ def _copy(item: FileItemDict):
...
+        "jsonl_example",
+        "parquet_example",
+        "listing",
+        "csv_example",
+        "csv_duckdb_example",
     ) == {
         "jsonl_example": 1034,
         "parquet_example": 1034,
-        "listing": 10,
+        "listing": 11,
```
IlyaFaer, line 194: A new file added - `.gz`

rudolfix (approved):
LGTM!

files are added, but the core PR must be fixed and merged first

Issue #262, opened 2023-09-13
[web scraper] verified source


rudolfix:
# Quick source info
- Name of the source: [e.g. web_scraper]
- What is the data source: any website

# Current Status
- [x] I plan to write it

# What source does/will do
The idea is to base the source on [scrapy](https://github.com/scrapy/scrapy). In theory, scrapy can be used with dlt directly because you can get the scraped data as a generator, on the other hand it is typically wrapped in opaque process from where there's no way to get data. scrapy has its own framework so we can fit `dlt` into `scrapy` (ie as export option)

We must investigate if to use scrapy or switch to beautiful soup and write own spider.

The requirements [for scrapy]
* implement `dlt.resource` that when provided a scrapy `Spider` will yield all the data that spider yields
* ideally data is yielded as it appears, not when scrapping is over and re-yielded from ie. csv file
* ideally data is scrapped concurrently (like I believe scrapy does)
* data should be paginated to speed up processing see see ie.
* return page metadata together with the items (ie page url, last modification time etc.)

# Test account / test data
Looks like we'll have plenty of websites to test against

# Additional context
Please provide one demo where we scrap PDFs and parse them in transformer as in

sultaniman:
I did a small research on this topic and I think we can use 
1. Custom [item exporter](https://docs.scrapy.org/en/latest/topics/exporters.html) or
2. Custom [item pipeline item](https://docs.scrapy.org/en/latest/topics/item-pipeline.html) which basically runs `dlt`
3. And feeds json values or uses `dlt` to create load packages and then finally starts exporting process to destination.

sultaniman:
@rudolfix @burnash wdyt?

burnash:
@sultaniman good points. Could you please do a proof of concept of this?

sultaniman:
I will do it once I check out this one https://github.com/dlt-hub/dlt/issues/811

sultaniman:
So I created a draft prototype which
1. Integrates `dlt` pipeline into scrapy spider,
2. Then sets up simple queue for communication between the spider and `dlt` pipeline,
3. `dlt` pipeline runs in a separate thread,
4. And I manually start crawler process in scrapy


```mermaid
flowchart LR
    queue[[queue]]
    pipeline[[dlt pipeline]]
    exit{{scraping done}}
    save([exit & save data])
    nodata{scraping done?}
    spider-- push results -->queue
    spider-- no more data -->exit
    queue-->pipeline
[...]

PR #320, opened 2024-01-12
[WIP] Implement Scrapy source

sultaniman:
Hey,

NOTE: This PR is still WIP

This PR implements [Scrapy source](https://github.com/dlt-hub/verified-sources/issues/262) which set's up communication between threads via queue, see the image below

scrapy

Also in `spider.py` we have our own base spider `class DLTSpiderBase` which provides very basic logic and as reference you can see our generic `DLTSpider` which is used only when callbacks are specified.

NOTE:
1. If custom spider is specified then it has a higher priority thus our source will ignore it,
2. Else if spider is not provided then it expects two callbacks `on_result`, `on_next_page`

`on_result` callback must be a generator because we want to progressively yield results to dlt pipeline

```py
def parse(response: Response) -> Generator[Dict, None, None]:
    for quote in response.css("div.quote"):
        yield {
            "quote": {
                "text": quote.css("span.text::text").get(),
                "author": quote.css("small.author::text").get(),
                "tags": quote.css("div.tags a.tag::text").getall(),
            },
        }

```

`on_next_page` should provide the next page or `None` value thus indicating the end of crawling process and will stop the pipeline.

```py
def next_page(response: Response) -> Optional[str]:
    return response.css("li.next a::attr(href)").get()
```

## TODO

* [x] Write readme and manual,
* [x] README
* [ ] Tests


sources/scrapy_scraper/README.md
```diff
@@ -0,0 +1,3 @@
+# 🕸️ Scrapy source
+
+
```
burnash, line 3: @sultaniman could you please add a quick "Run the pipeline" section here. For starters, how to run `scrapy_scraper_pipeline.py` with `python` or `scrapy`?

sources/scrapy_scraper/spider.py
```diff
@@ -0,0 +1,77 @@
...
+which you find in the official documentation
+https://docs.scrapy.org/en/latest/intro/tutorial.html#extracting-data-in-our-spider.
+
+To send the data from the spider to pipeline the spider needs a queue instance,
+once scrapy closes the session we also send a termination flag so
+DLT pipeline can stop waiting for more data.
+"""
+
+
+class QuotesSpider(DLTSpiderBase):
```
burnash, line 47: Since this is example it's best to move it out of the common code and into the example pipeline file `scrapy_scraper_pipeline.py`.

sources/scrapy_scraper_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+from scrapy_scraper import scrapy_source
+from scrapy_scraper.helpers import Scraper
+from scrapy_scraper.settings import SOURCE_SCRAPY_QUEUE_SIZE
+from scrapy_scraper.spider import QuotesSpider
+from scrapy_scraper.types import BaseQueue
+
+
+def pipeline_runner(pipeline: dlt.Pipeline, queue: BaseQueue) -> None:
+    load_info = pipeline.run(
+        scrapy_source(queue=queue),
```
burnash, line 14: Would it be possible to make `scrapy_source()` configurable via:
1. https://dlthub.com/docs/general-usage/credentials/configuration
2. function args
And also hide away the Queue from the end-user? (at least by default).

The idea of the verified sources is to make them reusable (but providing some sane configuration defaults). Let's try to think about the optimal user interface for `scrapy_source()`. 

1. First approach is to cover the case where the end user implements the spider. For example in this (https://quotes.toscrape.com) case I imagine we could pass the `start_urls` to `scrapy_source()` via config.toml (or an argument), and pass the Spider class as an agument. Like so:

      ```py
      from quotes_spider import QuotesSpider
      
      scrapy_source(
          start_urls=['https://quotes.toscrape.com'],
          spider=QuotesSpider
      )
      ```

2. Then, would it be possible to further simplify the configuration by implementing the spider implicitly? E.g. passing the `start_urls` as above, and also configuration that declaratively describes data extractors and pagination? (to cover the most common case).

burnash (commented):
Great progress, see my comments above.

sultaniman:
@burnash Thanks for comments and guidelines, I will address your concerns.

sources/scrapy_scraper/README.md
sultaniman: Yes I will definitely add more contents here, just have been trying to figure out testing of new sources.

burnash: Got it, no need for a big writeup: just how to run the example script since this is a special case. (+ prerequisites, if any)

sources/scraping/spider.py
```diff
@@ -0,0 +1,93 @@
...
+                    **parsed_result,
+                }
+            )
+
+        # If next page is available
+        # Then we create next request
+        # Else we stop spider because no pages left
+        next_page = self.on_next_page(response)
+        if next_page is not None:
+            next_page = response.urljoin(next_page)
```
rudolfix, line 83: hmmmm why? this is how they do it in their spiders:
```
for next_page in response.css('a.next'):
            yield response.follow(next_page, self.parse)
```

sources/scraping/spider.py
```diff
@@ -0,0 +1,93 @@
...
+            logger.debug(
+                "Got next page link",
+                extra={"current_url": response.url, "next_url": next_page},
+            )
+            yield scrapy.Request(next_page, callback=self.parse)
+        else:
+            # Important: once we are done we need to send signal
+            # to pipeline so it can finalize and shutdown
+            logger.info("No pages left, done.")
+            self.done()
```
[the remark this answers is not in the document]
rudolfix: hmmm how does it work now? if you have many threads reading pages in parallel and just one of the spiders gets here, it will send "done" to dlt and close the writer. what about other spiders? did you make the spider single threaded?

sources/scraping/types.py
```diff
@@ -0,0 +1,24 @@
...
+    class _Queue(Generic[T], Queue):
+        pass
+
+
+class BaseQueue(_Queue[T]):
+    pass
+
+
+OnNextPage = Callable[[Response], Optional[str]]
+OnResult = Callable[[Response], Generator[Any, None, None]]
```
rudolfix, line 24: Iterator[Any] == Generator[Any, None, None]]

sources/scraping/helpers.py
```diff
@@ -0,0 +1,51 @@
...
+        queue=queue,
+        name=f"{name}_spider",
+        start_urls=start_urls,
+        settings=settings,
+        **kwargs,
+    )
+
+    d = runner.join()
+    d.addBoth(lambda _: reactor.stop())  # type: ignore[attr-defined]
+    reactor.run()  # type: ignore[attr-defined]
```
rudolfix, line 37: here you should
1. wait for scrapy to finish (all spiders stopped yielding)
2. make sure that all messages got consumed from the queue
3. close the queue at the very end

sources/scraping/helpers.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def start_pipeline(
+    pipeline_runner: Callable[[], None],
+    scrapy_runner: Callable[[], None],
+) -> None:
+    """Convenience method which handles the order of starting of pipeline and scrapy"""
+    logger.info("Starting scraping pipeline")
+    pipeline_thread_runner = threading.Thread(target=pipeline_runner)
```
rudolfix, line 46: we need a better "host" for dlt. at the very least if should close the queue if exception occurred or if dlt exited and queue is not yet closed. if possible the dlt exception should be re-raised in main thread

rudolfix (changes requested):
@sultaniman good job! it was not easy to put those complicated things together :) please look at my comments
- better use resource not dlt source
- I have a few edge cases for the queue and I do not understand how spider really works - happy to discuss that
- IMO we need better thread runner for dlt and scrapy

My other idea to feed dlt via scrapy pipeline https://docs.scrapy.org/en/latest/topics/item-pipeline.html#topics-item-pipeline
what is good about it is that (IMO) you can use standard scrapy spiders and scrapy cli to run scraping
still (most probably) you'd use queue and a thread to send items to dlt but architecture should be way way easier so please take a look

sources/scraping/README.md
```diff
@@ -0,0 +1,124 @@
+# 🕸️ Scraping source
+
+Scraping source allows you to scrape content from web and uses [Scrapy](https://doc.scrapy.org/en/latest/)
+to enable this capability.
+
+## 🧠 How it works?
```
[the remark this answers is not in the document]
sultaniman: Fair, I will move this kind of information to the very bottom or a separate markdown file.

sources/scraping/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+from dlt.sources import DltResource, DltSource
+
+from .helpers import init_scrapy_runner
+from .settings import SOURCE_SCRAPY_SPIDER_SETTINGS, SOURCE_SCRAPY_QUEUE_SIZE
+from .spider import DLTSpider, DLTSpiderBase
+from .types import BaseQueue, OnNextPage, OnResult
+
+__all__ = ["build_scrapy_source"]
+
+logger = logging.getLogger(__file__)
```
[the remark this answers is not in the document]
sultaniman: Will change it.

sultaniman:
Closing this in favor of https://github.com/dlt-hub/verified-sources/pull/332/

Issue #321, opened 2024-01-14
File chunking settings ignored in filesystem source


deanja:
See partial test and proposed fix PR below

**Source name**
filesystem

**Describe the bug**
filesystem() method returns wrong sized chunks.  

**To Reproduce**
Debug the new test at link below.  The test needs more work, but debugging it will show the problem and fix.

**Expected behavior**
2 files per chunk

**Actual behaviour**
up to 5 files, since it's checking against length of file_dict, not files_chunk

** Potential impact
- Pipelines that access a lot of files using filesystem Resource may run slow, because they are getting chunks of 5 files even if they requested 100, 1,000 etc. 
- Conversely, once the bug is fixed some memory probs might appear for pipelines with large files because chunks size will go up from 5 to 100 (default is 100)

**Other context**
Test and fix   https://github.com/dlt-hub/verified-sources/pull/322 . The automated test does not work yet because filesystem() does not return a chunk of files.  Files were already unchunked by pipe.py.   I didn't have time yet to go up the stack  into that.

burnash:
Fixed in #322

PR #322, opened 2024-01-14
Fix filesystem chunk bug #321

deanja:
# Tell us what you do here

Show and fix a bug

# Relevant issue

[issue 321](https://github.com/dlt-hub/verified-sources/issues/321)

# More PR info
Note that the automated test doesn't work yet but bug and fix can be observed by debugging the test.   Test needs to take into account that pipe.py unchunks the output already.


rudolfix (approved):
wow! nice catch @deanja thanks for fixing that!

Issue #323, opened 2024-01-18
emit scale and precision in sql database


rudolfix:
**Which source is this for**
sql database

**Describe the data you'd like, be specific, link the docs**
we support scale and precision for several data types
https://dlthub.com/docs/general-usage/schema#tables-and-columns
we can take this information from reflected source schema and pass it to dlt resources. following data types should be handled
- decimal
- text, binary
- bigint (8, 4, 2 and 1 byte integers)

this behavior should be disabled by default as incompatible precision may be generated

**Additional context**
Please update documentation and tests

PR #324, opened 2024-01-18
sql-database: detect precision and scale of supported column types

steinitzu:
# Tell us what you do here

Detects precision and scale attributes for reflected columns in the source database.

Types supported:

`int/bigint/smallint` -> `bigint` with corresponding precision
`numeric` -> `decimal` with precision (sqlalchemy `Numeric` is base of all "decimal" like types)
`timestamp/time` -> dlt `timestamp/time` with precision
`string/binary` -> `text`/`binary` with length as precision

Detects types using the base type classes in `sqlalchemy.sql.sqltypes` so should work for all dialects.

# Relevant issue

resolves  https://github.com/dlt-hub/verified-sources/issues/323

### DIFF
--- a/sources/sql_database/schema_types.py
+++ b/sources/sql_database/schema_types.py
@@ -1,0 +1,61 @@
+from typing import Any, Optional
+
+from sqlalchemy.sql import sqltypes
+from sqlalchemy import Table, Column
+
+from dlt.common.schema.typing import TColumnSchema, TTableSchemaColumns
+
+
+def sqla_col_to_column_schema(sql_col: Column[Any]) -> Optional[TColumnSchema]:
+    """Infer dlt schema column type from an sqlalchemy type.
+
+    Precision and scale is inferred from that types that support it,
+    such as numeric, varchar, int, bigint
+    """
+    sql_t = sql_col.type
+    col = None
+
+    if isinstance(sql_t, sqltypes.BigInteger):
+        col = dict(name=sql_col.name, data_type="bigint", precision=64)
+    elif isinstance(sql_t, sqltypes.SmallInteger):
+        col = dict(name=sql_col.name, data_type="bigint", precision=16)
+    elif isinstance(sql_t, sqltypes.Integer):
+        col = dict(name=sql_col.name, data_type="bigint", precision=32)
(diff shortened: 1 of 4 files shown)


steinitzu:
The rfam db is down I think, why these tests are failing

```
FAILED tests/sql_database/test_sql_database_source.py::test_load_mysql_data_load[postgres] - sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1045, "Access denied for user 'rfamro'@'172.183.107.247' (using password: NO)")
(Background on this error at: https://sqlalche.me/e/20/e3q8)
FAILED tests/sql_database/test_sql_database_source.py::test_load_mysql_data_load[duckdb] - sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1045, "Access denied for user 'rfamro'@'172.183.107.247' (using password: NO)")
(Background on this error at: https://sqlalche.me/e/20/e3q8)
```

sources/sql_database/schema_types.py
```diff
@@ -0,0 +1,61 @@
...
+            name=sql_col.name,
+            data_type="decimal",
+            precision=sql_t.precision,
+            scale=sql_t.scale,
+        )
+    elif isinstance(sql_t, sqltypes.String):
+        col = dict(name=sql_col.name, data_type="text", precision=sql_t.length)
+    elif isinstance(sql_t, sqltypes._Binary):
+        col = dict(name=sql_col.name, data_type="binary", precision=sql_t.length)
+    elif isinstance(sql_t, sqltypes.DateTime):
```
rudolfix, line 35: I'm not sure we should emit timestamp precision. it may just create incompatibilities and AFAIK sql alchemy uses datetime (floats) so precision is always microseconds anyway.
TLDR: I'd leave datetime and time precision

rudolfix (changes requested):
looks good! pls check my comment about datetime
this mysql database sometimes is online sometimes not. so let's ignore it for now

sources/sql_database/schema_types.py
steinitzu: True, makes sense. I've removed time/datetime

rudolfix (approved):
LGTM!

PR #328, opened 2024-01-22
fix tests and documentation depending on rfam database

sh-rp:
The rfam database currently yields access denied for the rfamro user. I have migrated to another public genome database called ensembl.


rudolfix (approved):
amazing

PR #331, opened 2024-01-25
test(kafka): use mocks instead of reading from Kafka cluster

IlyaFaer:
To avoid flaky errors because of slow Kafka sync, it's proposed to use mocks. It'll allow to test the functionality, not dealing with a real Kafka service, thus, all the sync errors will become not actual.

### DIFF
--- a/tests/kafka/test_kafka.py
+++ b/tests/kafka/test_kafka.py
@@ -1,6 +1,7 @@
 import pytest
 import random
 import time
+from unittest import mock
 
 from confluent_kafka import Producer
 from confluent_kafka.admin import AdminClient, NewTopic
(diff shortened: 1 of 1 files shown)


tests/kafka/test_kafka.py
```diff
@@ -189,58 +190,29 @@ def test_kafka_read_with_timestamp(kafka_timed_messages):
...
+    offset_mock = mock.Mock()
+    offset_mock.offset = -1
+
+    consumer = mock.Mock()
+    consumer.list_topics = mock.Mock(return_value=topics_mock)
+    consumer.offsets_for_times = mock.Mock(return_value=[offset_mock])
+    consumer.get_watermark_offsets = mock.Mock(return_value=[0, 10])
+
+    tracker = OffsetTracker(consumer, ["topic1"], {}, start_from=pendulum.now(tz="UTC"))
+    assert tracker["topic1"]["0"] == {"cur": 9, "max": 10}
```
IlyaFaer, line 215: As offsets are not indexes, but the number of messages to omit, the current cursor is shifted at `-1` in the state. However, during an extraction the consumer increments it, so in this case no data will be read, as the given timestamp happened AFTER the last message in the topic.

rudolfix: OK, LGTM

rudolfix (approved):

PR #332, opened 2024-01-25
Scrapy source using scrapy

sultaniman:
Hey,

This PR implements [Scrapy source](https://github.com/dlt-hub/verified-sources/issues/262) which set's up communication between threads via queue, see the image below

scrapy

## TODO

* [x] Write readme and manual,
* [x] README,
* [x] Demo with real scraping,
* [x] Another demo with more complex scraping,
* [x] Demo interaction with generate pipeline resource like using `add_limit(...)`
* [x] Tests

With the feedback from @rudolfix we decided to do the following changes

- Hide the queue,
- Remove source and resource in favor of dynamic creation,
- Higher level run_scrapy_pipeline -> __init__ => simply run pipeline and exit,
- Higher level run_scrapy_pipeline to run parallel crawls,
- Settings should be a param w/ default settings,
- Remove pipeline item use signals,
- Close the queue in the pipeline runner.


sources/scraping/scrapy/pipeline_item.py
```diff
@@ -0,0 +1,19 @@
...
+
+class ScrapingPipelineItem:
+    def process_item(
+        self, item: Type[scrapy.Item], spider: Type[DltSpider]
+    ) -> Type[scrapy.Item]:
+        """Handles sending the result of spider to scraping resource of Dlt pipeline"""
+        if not spider.queue.is_closed:
+            spider.queue.put(item)  # type: ignore
+        else:
+            logger.warning("Queue is closed")
```
rudolfix, line 17: if a writer detects that queue is closed that means that reader died. so here we must stop crawling ie. by raising an exception or doing it some other recommended way (by scrapy)

sources/scraping/scrapy/spider.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Optional, TypeVar
+import scrapy  # type: ignore[import-untyped]
+
+from ..queue import BaseQueue
+
+T = TypeVar("T")
+
+
+class DltSpider(scrapy.Spider):
```
rudolfix, line 9: why would our Spider need a queue? We do not need `DltSpider` now

sources/scraping/helpers.py
```diff
@@ -0,0 +1,159 @@
...
+            queue=self.queue,
+            include_headers=self.include_headers,
+            name="scraping_spider",
+            start_urls=self.start_urls,
+            **kwargs,
+        )
+
+        d = runner.join()
+        d.addBoth(lambda _: reactor.stop())  # type: ignore[attr-defined]
+        reactor.run()  # type: ignore[attr-defined]
```
rudolfix, line 75: after run you must join the queue (to make sure it is empty) and then close it

sources/scraping/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+            batch = []
+
+        try:
+            result = queue.get(timeout=queue_result_timeout)
+            batch.append(result)
+
+            # Mark task as completed
+            queue.task_done()
+        except Empty:
+            logger.info(
```
rudolfix, line 52: empty queue is not a reason to stop. just loop

sources/scraping/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    batch: List[T] = []
+    num_batches = 0
+    while True:
+        if len(batch) >= batch_size:
+            num_batches += 1
+            yield batch
+            batch = []
+
+        try:
+            result = queue.get(timeout=queue_result_timeout)
```
rudolfix, line 46: before get check if queue is not closed. if it is closed (means that all spiders are completed or terminated) you exit the while loop

sources/scraping/helpers.py
```diff
@@ -0,0 +1,159 @@
...
+        include_headers: bool = False,
+    ) -> None:
+        self.queue = queue
+        self.spider = spider
+        self.settings = settings
+        self.start_urls = start_urls
+        self.include_headers = include_headers
+
+    def run(self, *args: Any, **kwargs: AnyDict) -> None:
+        runner = CrawlerRunner(settings=self.settings)
```
rudolfix, line 63: Question: why use this when you can use (the recommended way) this: https://docs.scrapy.org/en/latest/topics/practices.html#run-scrapy-from-a-script

sources/scraping/helpers.py
```diff
@@ -0,0 +1,159 @@
...
+    queue_size: int = dlt.config.value,
+    spider: t.Type[DltSpider] = None,
+    start_urls: t.List[str] = dlt.config.value,
+    include_headers: t.Optional[bool] = dlt.config.value,
+) -> t.Tuple[PipelineRunner, ScrapyRunner, t.Callable[[], None]]:
+    if queue is None:
+        queue = BaseQueue(maxsize=queue_size)
+
+    settings = {
+        "ITEM_PIPELINES": {ScrapyPipelineItemModule: 100},
```
rudolfix, line 140: what this `100` here means?

sources/scraping_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+from scraping.helpers import create_pipeline_runner
+
+
+class MySpider(DltSpider):
+    def parse(self, response: Response, **kwargs: Any) -> Any:
+        for next_page in response.css("li.next a::attr(href)"):
+            if next_page:
+                yield response.follow(next_page, self.parse)
+            else:
+                self.queue.close()
```
rudolfix, line 18: this won't be necessary when we implement queue closing in the scrapy runner

sources/scraping/helpers.py
```diff
@@ -0,0 +1,159 @@
...
+        spider=spider,
+        start_urls=start_urls,
+        settings=settings,
+        include_headers=include_headers,
+    )
+
+    pipeline_runner = PipelineRunner(pipeline=pipeline)
+
+    def wait_for_results() -> None:
+        queue.join()
```
rudolfix, line 155: scrapy runner should close the queue

sources/scraping/helpers.py
```diff
@@ -0,0 +1,159 @@
...
+        write_disposition: TWriteDisposition = None,
+        columns: TAnySchemaColumns = None,
+        primary_key: TColumnNames = None,
+        schema: Schema = None,
+        loader_file_format: TLoaderFileFormat = None
+        ```
+        """
+
+        def run() -> None:
+            self.pipeline.run(data, write_disposition=write_disposition, **kwargs)
```
rudolfix, line 106: here's another trick: you need to catch an exception here and close the queue in the handler then reraise. this will inform writer (scrapy) that reader died so it will raise and stop scraping.

rudolfix (changes requested):
yeah, now with the closing queue implemented, you can simplify the code even simpler. please read the review. when it is done you'll be able to get rid of DltSpider and people will be able to use a standard spider here.

sources/scraping/helpers.py
sultaniman: Got it, I will move it inside the scrapy runner then.

sources/scraping/helpers.py
sultaniman: There were some issues with twisted not properly handling teardown process. I will try it again :)

sources/scraping/helpers.py
sultaniman: It defines the order of execution (calls) to pipeline items
```
The integer values you assign to classes in this setting determine the order in which they run: items go through from lower valued to higher valued classes. It’s customary to define these numbers in the 0-1000 range.
```

sources/scraping/scrapy/spider.py
sultaniman: The reason for this is that we need to pass queue instance to Item Pipeline and at the moment there is no officially documented way to pass queue instance in Item pipeline we have defined

sultaniman: Other option would be defining ItemPipeline inside scrapy runner and passing to crawler process something like
```py
lass ScrapyRunner(Runnable):
    def run(self, *args: Any, **kwargs: AnyDict) -> None:
        queue = self.queue
        class ItemPipeline:
              def process_item(...):
                   queue.put(...)
       
        runner = CrawlerProcess(settings={
              "ITEM_PIPELINES": {
                  ItemPipeline: 100
              }
        })
```

sources/scraping/__init__.py
sultaniman: Updated.

sources/scraping/queue.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class BaseQueue(_Queue[T]):
+    def __init__(self, maxsize: int = 0) -> None:
+        super().__init__(maxsize)
+        self._state_lock = threading.Lock()
+        self._is_closed = False
+
+    def close(self) -> None:
+        with self._state_lock:
```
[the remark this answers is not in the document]
sultaniman: Done.

sources/scraping/scrapy/pipeline_item.py
sultaniman: So now it will raise `CloseSpider` exception which should stop crawling.

sources/scraping/scrapy/pipeline_item.py
```diff
@@ -0,0 +1,46 @@
+from abc import ABC
```
sultaniman, line 1: @rudolfix I opted to use a closure function to return scrapy item pipeline and give access to queue instance to internal class

sources/scraping/scrapy/spider.py
sultaniman: We no longer need out custom spider.

sources/scraping/helpers.py
sultaniman: Adjusted it. Now we are using `CrawlerProcess`

sources/scraping/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    batch: List[T] = []
+    num_batches = 0
+    while True:
+        if len(batch) >= batch_size:
+            num_batches += 1
+            yield batch
+            batch = []
+
+        try:
+            if queue.is_closed:
```
rudolfix, line 46: would it be hard to move it to Queue (override `get` method)

sources/scraping/scrapy/pipeline_item.py
```diff
@@ -0,0 +1,50 @@
...
+    class ScrapingPipelineItem(PipelineItem):
+        def process_item(
+            self, item: Type[scrapy.Item], spider: scrapy.Spider
+        ) -> Type[scrapy.Item]:
+            """Handles sending the result of spider to scraping resource of Dlt pipeline"""
+            if not queue.is_closed:
+                queue.put(item)  # type: ignore
+            else:
+                logger.error("Queue is closed")
+                raise CloseSpider("Queue is closed")
```
rudolfix, line 46: so it works after all?

sources/scraping/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .helpers import ScrapingConfig
+from .queue import BaseQueue, QueueClosedError
+
+
+T = TypeVar("T")
+
+
+@dlt.source
+def scrapy_source(queue: BaseQueue[T]) -> Iterator[Any]:
```
rudolfix, line 18: I do not think we need to expose any of those. user needs only your `run` function for interactions.

sources/scraping/helpers.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self.thread_runner = threading.Thread(target=run)
+        self.thread_runner.start()
+
+    def join(self) -> None:
+        self.thread_runner.join()
+
+
+@with_config(sections=("sources", "scraping"), spec=ScrapingConfig)
+def create_pipeline_runner(
```
rudolfix, line 116: let's sync on how to simplify this. I think user needs a function that will just run everything without exposing separate function and wait.

rudolfix (changes requested):
looks good, but still we need to simplify it. I'll ping you for the code walkthrough

sources/scraping/queue.py
```diff
@@ -0,0 +1,81 @@
...
+
+                item = super().get(block, timeout=timeout)
+                batch.append(item)
+
+                # Mark task as completed
+                self.task_done()
+            except Empty:
+                logger.info(f"Queue has been empty for {timeout}s...")
+                return batch
+            except QueueClosedError:
```
sultaniman, line 57: @rudolfix should we catch both exceptions at once, I mean to merge `except Empty, QueueClosedError`?

sources/scraping/runner.py
```diff
@@ -0,0 +1,133 @@
...
+        ```
+        """
+
+        def run() -> None:
+            try:
+                self.pipeline.run(data, **kwargs)
+            except Exception:
+                logger.error("Error during pipeline.run call, closing the queue")
+                self.queue.close()
+                raise
```
sultaniman, line 91: just propagate the exception, raising CloseSpider won't have any effect on scrapy because we are running the pipeline in a separate thread.

sources/scraping/runner.py
```diff
@@ -0,0 +1,131 @@
...
+    ) -> None:
+        """You can pass kwargs which are passed to `pipeline.run`"""
+        logger.info("Starting pipeline")
+        pipeline_worker = self.pipeline_runner.run(
+            # Queue get_batches is a generator so we can
+            # pass it to pipeline.run and dlt will handle the rest.
+            dlt.resource(
+                self.queue.get_batches,
+                name=f"{self.pipeline_runner.pipeline.pipeline_name}_results",
+            ),
```
sultaniman, line 122: @rudolfix should we do this, I wrapped it because wanted to generate proper table name

rudolfix (approved):
LGTM!

we can move it into docs&demo phase. docstrings could be a little improves, also it is not explained how to change scrapy settings

sources/scraping/runner.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def on_item_scraped(self, item: Item) -> None:
+        if not self.queue.is_closed:
+            self.queue.put(item)
+        else:
+            logger.info(
+                "Queue is closed ",
+                extra={"pipeline_name": self.pipeline_name},
+            )
+            raise CloseSpider("Queue is closed")
```
rudolfix, line 31: looks like you cannot raise this to close spider:
https://github.com/scrapy/scrapy/issues/3435

we'd need to have something else. we may just schedule to stop crawl/engine. also test hand often after stopping

rudolfix (changes requested):

sources/scraping/runner.py
```diff
@@ -0,0 +1,181 @@
...
+            self.stopping = True
+            self.crawler.stop()
+
+    def on_engine_stopped(self) -> None:
+        logger.info(f"Crawling engine stopped for pipeline={self.pipeline_name}")
+        self.stopping = True
+        self.queue.join()
+        self.queue.close()
+
+    def __call__(self, crawler: CrawlerProcess) -> Self:
```
sultaniman, line 41: @rudolfix I added __call__ method here so the context manager expression with call looks better.

sources/scraping/runner.py
sultaniman: I incorporated discussed changes and pushed the update.

sultaniman:
@rudolfix I will add more docstrings and details clarifications in readme and pipeline examples.

rudolfix (changes requested):
All looks good now but tests somehow do not end. Most probably a thread still executes that is not a daemon thread.
It all works each time when I run tests locally, also when I mix them other tests :/
My take: 
- try to stop reactor after the test
- if it does not work then we skip all reactor tests

this is really good job! just the way scrapy works is a little bit obnoxious

sultaniman:
@rudolfix I decided to mark to skip the test since we have other more pressing issues and left the todo item to do 10% research on this.

Tried to reproduce this locally by
1. running tests for scraping along with other tests,
2. running scraping test suite,
3. running particular test if it hangs,
4. swapping out a queue with custom test queue which terminates after N retries

sultaniman:
Created a bug ticket: https://github.com/dlt-hub/verified-sources/issues/394

rudolfix (approved):
LGTM!

Issue #310, opened 2023-12-21
[Bing Webmaster] verified source


willi-mueller:
# Quick source info
- Name of the source: Bing Webmaster
- What is the data source: Provides data about organic traffic coming via Bing search. You get clicks and impressions per date, page, and query. I just finished working on the [GetPageQueryStats](https://learn.microsoft.com/en-us/dotnet/api/microsoft.bing.webmaster.api.interfaces.iwebmasterapi.getquerytrafficstats?view=bing-webmaster-dotnet) endpoint.

# Current Status
- [ ] I plan to write it
- [ ] I've written it and run it for fun
- [ ] I run it on my local laptop
- [x] I run it in production
- I tried the following destinations with it: duckdb, bigquery

# What source does/will do
Please provide descriptions up to 3 most important use cases that users of this source do. Those use cases will be:
- implemented
- reviewed
- demonstrated in the source demo pipeline script
- documented
use case description is not only targeted at the developers but also by people creating test accounts and doing demos.

# Test account / test data
**still thinking about this question**
- [ ] I'd like `dltHub` to create a test account before we start
- [ ] I have test account and can share my test data or a test account with `dltHub` after we merge the source 

# Additional context
Anything we should know about your source

sultaniman:
PR has been merged now. Thank you very much @willi-mueller 🚀 !

PR #335, opened 2024-01-26
Source bing webmaster

willi-mueller:
# Tell us what you do here

- [x] implementing verified source (please link a relevant issue labeled as `verified source`)

# Relevant issue

issue #310 

# More PR info

## TODO before merge
- [x] copy unit tests into the test file
- [x] implement test cases for data loading
- [x] write README.md
- [x] @sultaniman obtain an API key for CI and verify a domain, e.g. dlthub.com


willi-mueller:
Hi, could you please give me feedback on:
1. how to parse the unix timestamp int into a date if the CI bans import of `datetime.date`. I didn't have time yet to check how to achieve that with `dlt.commons.pendulum`. So, if it's trivial for you I'd appreciate a pointer.
2. If you can obtain an API key from Bing Webmaster and share it with me so that I can implement the test suite that calls the API. If you could verify your dlthub.com domain with Bing that would be great! I don't think you would need to submit one or two pages for indexing because I could find your pages already via bing.com.

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        page = record.get("page")
+        params = {"siteUrl": site_url, "page": page, "apikey": bing_webmaster_api_key}
+        print(f"Fetching for site_url: {site_url}, page: {page}")
+        response = get_stats_with_retry(api_path, params)
+        if len(response) > 0:
+            yield parse_response(response, site_url, page)
+
+
+@dlt.source
+def chess_dlt_config_example(
```
sultaniman, line 96: Should we rename `chess` -> `bing | webmaster`?

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    Yields:
+        Iterator[Dict[str, Any]]: An iterator over list of organic traffic statistics.
+    """
+    api_path = "GetPageQueryStats"
+    for record in site_url_pages:
+        time.sleep(0.5)  # this avoids rate limit observed after dozens of requests
+        site_url = record.get("site_url")
+        page = record.get("page")
+        params = {"siteUrl": site_url, "page": page, "apikey": bing_webmaster_api_key}
+        print(f"Fetching for site_url: {site_url}, page: {page}")
```
sultaniman, line 89: As a sidenote, although it is optional, should we use `dlt.common.logging`?

sources/bing_webmaster/__init__.py
sultaniman: Also looks like this is not being used

sources/bing_webmaster/__init__.py
willi-mueller: Wonderful. I was wondering why every verified source uses `print()`. I'm glad that there is a recommended way of logging. I'll implement that!

sources/bing_webmaster/__init__.py
willi-mueller: Indeed, I'll delete it since it's still part of the generated boilerplate.

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    A dlt source for the Bing Webmaster api.
+    It groups resources for the APIs which return organic search traffic statistics
+    Args:
+        site_urls: List[str]: A list of site_urls, e.g, ["dlthub.com", "dlthub.de"]. Use this if you need the weekly traffic per site_url and page
+        site_url_pages: Iterable[DictStrStr]: A list of pairs of site_url and page. Use this if you need the weekly traffic per site_url, page, and query
+    Returns:
+        Sequence[DltResource]: A sequence of resources that can be selected from including page_stats and page_query_stats.
+    """
+    return (
```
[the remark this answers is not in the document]
willi-mueller: Thank you! Did you have the following in mind?
```python
yield (
        page_stats(site_urls),
        page_query_stats(site_url_pages),
    )
```

I'm not sure because both the boilerplate code as well as all verified sources I checked return a tuple or list of resources, I couldn't find a source yet which yields resources.

sources/bing_webmaster/helpers.py
```diff
@@ -0,0 +1,76 @@
+"""Chess source helpers"""
```
willi-mueller, line 1:
```suggestion
"""Bing Webmaster source helpers"""
```

sources/bing_webmaster/__init__.py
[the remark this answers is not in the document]
AstrakhantsevaAA: it's required that data from resources must be yielded, but it's not necessary to yield resources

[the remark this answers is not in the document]
AstrakhantsevaAA: you can yield resources if you need to return them with pipe for example, look at pipedrive: https://github.com/dlt-hub/verified-sources/blob/master/sources/pipedrive/__init__.py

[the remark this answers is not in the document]
willi-mueller: Good! But I think that the data is yielded from the resources. Or can a yield not call a helper function which returns? Please advise.

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+)
+def page_stats(
+    site_urls: List[str], bing_webmaster_api_key: str = dlt.secrets.value
+) -> Iterator[Iterator[DictStrAny]]:
+    """
+    Get detailed traffic statistics for top pages belonging to a site_url
+    API documentation:
+    https://learn.microsoft.com/en-us/dotnet/api/microsoft.bing.webmaster.api.interfaces.iwebmasterapi.getpagestats
+    Args:
+        site_urls (List[str]): List of player usernames to retrieve archives for.
```
AstrakhantsevaAA, line 49: sounds like something from chess

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+        Sequence[DltResource]: A sequence of resources that can be selected from including page_stats and page_query_stats.
+    """
+    return (
+        page_stats(site_urls),
+        page_query_stats(site_url_pages),
+    )
+
+
+@dlt.resource(
+    write_disposition="merge",
```
AstrakhantsevaAA, line 36: is it possible to make it incremental (does the response have a "created_at" or similar key?)? you use merge mode and without incremental loading it will work very inefficient

sources/bing_webmaster/__init__.py
sultaniman: I think we should add this to verified source development page in documentation and to CONTRIBUTORS.md guideline wdyt @burnash ?

sultaniman: I am however not sure where to fit this, should it be in How-tos or in verified source development part?

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    https://learn.microsoft.com/en-us/dotnet/api/microsoft.bing.webmaster.api.interfaces.iwebmasterapi.getpagequerystats
+
+    Args:
+        site_url_page (Iterable[DictStrStr]): Iterable of site_url and pages to retrieve statistics for. Can be result of a SQL query, a parsed sitemap, etc.
+    Yields:
+        Iterator[Dict[str, Any]]: An iterator over list of organic traffic statistics.
+    """
+    api_path = "GetPageQueryStats"
+    for record in site_url_pages:
+        time.sleep(0.5)  # this avoids rate limit observed after dozens of requests
```
AstrakhantsevaAA, line 85: don't you want to handle the rate limit error instead of sleeping on every iteration?
here is an example: https://github.com/dlt-hub/verified-sources/blob/master/sources/workable/workable_client.py#L41

also there is no pagination?

sources/bing_webmaster/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def get_url_with_retry(url: str, params: DictStrStr) -> DictStrAny:
+    r = requests.get(url, headers=HEADERS, params=params)
+    r.raise_for_status()
+    return r.json()  # type: ignore
+
+
+def get_stats_with_retry(api_path: str, params: DictStrStr) -> List[DictStrAny]:
+    url = f"{BASE_URL}{api_path}"
```
sultaniman, line 20: should this use `urljoin`?

sources/bing_webmaster/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@dlt.resource(
+    write_disposition="merge",
+    merge_key=("date", "page", "site_url"),
+    primary_key=("date", "page", "site_url"),
+    table_name="bing_page_stats",
+)
+def page_stats(
+    site_urls: List[str], bing_webmaster_api_key: str = dlt.secrets.value
```
sultaniman, line 42: Wrt `bing_webmaster_api_key` should we exit/raise exception if key is not set?

sources/bing_webmaster/__init__.py
sultaniman: It is some weird MSFT API which does this kind of rate limiting, but if @willi-mueller has some time could we research if they have better alternative to handle rate limited endpoints?

willi-mueller:
> Should we also add a README.md?

absolutely, I'll type that up together with the integration test suite!
Thank you so much @AstrakhantsevaAA and @sultaniman for your review, I'll work on your points.

sources/bing_webmaster/__init__.py
willi-mueller: Yes, putting this in the docs would be great. And in the same PR we could refactor the existing sources to the logger.

sources/bing_webmaster/__init__.py
willi-mueller: Unfortunately, they just give an unspecific HTTP 400 authentication error and the API access is blocked for minutes. Thus, I would keep the sleeping – otherwise, I observed that after around 10min of hammering the API with requests (e.g. while iterating over a sitemap) our pipeline would have to wait a couple of minutes until it's unblocked. I haven't done systematic experiments to determine what would be a backoff time after getting this error.

In this way, I've been able to consistently retrieve data for 2100 pages (i.e. 2100 requests).

Good idea to experimentally research the specific rate limit though, that would be cleaner than putting a sleep timer.

sources/bing_webmaster/__init__.py
willi-mueller: Good idea! Do you have an exception in mind that we could use? `ValueError`?

I couldn't find any example yet in the verified sources.

sources/bing_webmaster/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Initialize the pipeline
+
+```bash
+dlt init bing_webmaster duckdb
+```
+
+Here, we chose duckdb as the destination. Alternatively, you can also choose redshift, bigquery, or
+any of the other [destinations](https://dlthub.com/docs/dlt-ecosystem/destinations/).
+
+## Add credentials
```
sultaniman, line 30: Can we also add a notion of secret name which should be in secrets.toml?

sources/bing_webmaster/__init__.py
sultaniman: I think it is fine to not raise any exception since requests will fail with HTTP 400 right?

sources/bing_webmaster/__init__.py
AstrakhantsevaAA: > "merge" is a [type of incremental loading](https://dlthub.com/docs/general-usage/incremental-loading).

It is, I talked only about efficiency =))
Such a weird API, I think you can keep it like this, looks like there is no another solution. When you request data there is no params like "since" or "end_date" or similar?

sources/bing_webmaster/__init__.py
AstrakhantsevaAA: crazy API :D

sources/bing_webmaster/__init__.py
willi-mueller: Good, thank you for the confirmation!
No, there is neither any filter parameter nor any pagination. They give all the data for 26 weeks for all your thousands of pages in a single response.

sources/bing_webmaster/__init__.py
willi-mueller: Indeed, some interesting design decisions. Those were the times before they had access to ChatGPT 🙃 

Ballmer I Love This Company

sources/bing_webmaster/README.md
willi-mueller: Good point, I just did that. Running the latest commit eeb1a9b would require that you change in your secrets.toml from `bing_webmaster_api_key` to `api_key`.

sources/bing_webmaster/__init__.py
willi-mueller: Yes, I removed my check code because even without it we have the following behaviour:


### Without any key set
Exception:
```
dlt.common.configuration.exceptions.ConfigFieldMissingException: Following fields are missing: ['api_key'] in configuration with spec PageStatsConfiguration
	for field "api_key" config providers and keys were tried in following order:
		In Environment Variables key BING_WEBMASTER_PAGE_STATS__SOURCES__BING_WEBMASTER__PAGE_STATS__API_KEY was not found.
		In Environment Variables key BING_WEBMASTER_PAGE_STATS__SOURCES__BING_WEBMASTER__API_KEY was not found.
		In Environment Variables key BING_WEBMASTER_PAGE_STATS__SOURCES__API_KEY was not found.
		In Environment Variables key BING_WEBMASTER_PAGE_STATS__API_KEY was not found.
		In secrets.toml key bing_webmaster_page_stats.sources.bing_webmaster.page_stats.api_key was not found.
		In secrets.toml key bing_webmaster_page_stats.sources.bing_webmaster.api_key was not found.
		In secrets.toml key bing_webmaster_page_stats.sources.api_key was not found.
		In secrets.toml key bing_webmaster_page_stats.api_key was not found.
		In Environment Variables key SOURCES__BING_WEBMASTER__PAGE_STATS__API_KEY was not found.
		In Environment Variables key SOURCES__BING_WEBMASTER__API_KEY was not found.
		In Environment Variables key SOURCES__API_KEY was not found.
		In Environment Variables key API_KEY was not found.
		In secrets.toml key sources.bing_webmaster.page_stats.api_key was not found.
		In secrets.toml key sources.bing_webmaster.api_key was not found.
		In secrets.toml key sources.api_key was not found.
		In secrets.toml key api_key was not found.
Please refer to https://dlthub.com/docs/general-usage/credentials for more information
```

### With a wrong key
Exception:
```
requests.exceptions.HTTPError: 400 Client Error: Error for url: https://ssl.bing.com/webmaster/api.svc/json/GetPageStats?siteUrl=satellite.me&apikey=my-wrong-key
```

sultaniman (approved):

Issue #308, opened 2023-12-17
support `gdrive` filesystem in `filesystem` source


rudolfix:
**Which source is this for**
adds google drive filesystem support to `filesystem` source

**Describe the data you'd like, be specific, link the docs**
We want to support gdrive in `filesystem` source the same way we support buckets. We already started an effort to improve existing gdrive spec and merged a few improvements but not all of them succeeded:
https://github.com/fsspec/gdrivefs/pull/36
https://github.com/dlt-hub/dlt/pull/738
https://github.com/dlt-hub/verified-sources/pull/288

To quickly get started:
1. * [ ] initially you can add all the code (including custom fsspec filesystem) to the `filesystem` verified source or to branch of `dlt`
2. * [ ] if you need to modify library core, do it a fork from which we can install it for testing. after a review we'll merge it
3. * [ ] if we make gdrive working we can do a PR to original repo - that would be preferred.
4. * [ ] please reproduce all the test cases by creating a repo with the test files used by other filesystems in `filesystem` source

We had interesting problem with passing `root_file_id` which is required to read files from other users' google drive (will be always the case when using service credentials). since we want to make it compatible with other file systems where we pass the location of the files via `url` we decided to add it to query string and then apply in the fsspec filesystem. there's actually a way to get query from url when implementing fsspec filesystem so maybe we can use that

PR #348, opened 2024-02-05
test(filesystem): enable testing

IlyaFaer:
Towards https://github.com/dlt-hub/verified-sources/issues/308

The core PR is supposed to be merged first:
https://github.com/dlt-hub/dlt/pull/932

Enabling testing with the `dlt` testing bucket.


IlyaFaer:
@rudolfix, yes, this one includes Google Drive tests on the common tests case. Now, when all the Google Drive things are merged, it can be merged too.

rudolfix (changes requested):
@IlyaFaer we have dlt with gdrive support finally merged! if you could take the newest code form master here, bump the dlt version and push it again it would be great! then we can close it

rudolfix (approved):
@IlyaFaer so actually the url to gdrive was wrong. that's why you got 404. possibly you were testing on old version of dlt locally... but I have no idea how it could work ;)

IlyaFaer:
@rudolfix, well, I don't know, locally it worked

image

PR #349, opened 2024-02-06
Squash messages to one table for slack

VioletM:
# Tell us what you do here

- [ ] implementing verified source (please link a relevant issue labeled as `verified source`)
    * Now all messages are extracted into one table instead of a table per channel. 
    * Add replies resource. Now replies are also extracted by slack source

### DIFF
--- a/sources/slack/__init__.py
+++ b/sources/slack/__init__.py
@@ -1,7 +1,6 @@
 """Fetches Slack Conversations, History and logs."""
 
-from functools import partial
-from typing import Any, Dict, Iterable, List, Optional
+from typing import Iterable, List, Optional, Tuple
 
 import dlt
 from dlt.common.typing import TAnyDateTime, TDataItem
@@ -24,6 +23,7 @@ def slack_source(
     start_date: Optional[TAnyDateTime] = DEFAULT_START_DATE,
     end_date: Optional[TAnyDateTime] = None,
     selected_channels: Optional[List[str]] = dlt.config.value,
+    replies: bool = False,
 ) -> Iterable[DltResource]:
     """
     The source for the Slack pipeline. Available resources are conversations, conversations_history
@@ -35,6 +35,7 @@ def slack_source(
         start_date: The start time of the range for which to load. Defaults to January 1st 2000.
         end_date: The end time of the range for which to load data.
         selected_channels: The list of channels to load. If None, all channels will be loaded.
+        replies: Boolean flag indicating if you want a replies table to be present as well. False by default.
 
     Returns:
         Iterable[DltResource]: A list of DltResource objects representing the data resources.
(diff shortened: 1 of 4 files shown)


sources/slack/__init__.py
```diff
@@ -48,24 +49,48 @@ def slack_source(
...
         """Yield all channels as a DLT resource."""
         yield from channels
 
-    yield channels_resource
-
     @dlt.resource(name="users", primary_key="id", write_disposition="replace")
     def users_resource() -> Iterable[TDataItem]:
-        """Yield all channels as a DLT resource.
+        """
+        Yield all channels as a DLT resource.
```
AstrakhantsevaAA, line 93: typo in the docstring
"Yield all users as a dlt resource"

sources/slack/__init__.py
```diff
@@ -79,78 +104,56 @@ def users_resource() -> Iterable[TDataItem]:
...
+            }
+
+            for page_data in api.get_pages(
+                resource="conversations.history",
+                response_path="$.messages[*]",
+                params=params,
+                datetime_fields=MSG_DATETIME_FIELDS,
+                context={"channel": channel_data["id"]},
+            ):
+                yield from page_data
```
AstrakhantsevaAA, line 146: yield whole pages, it's more efficient. it's better to have a for-loop in the "replies_resource" transformer

AstrakhantsevaAA: https://dlthub.com/docs/reference/performance#yield-pages-instead-of-rows

AstrakhantsevaAA (approved):

AstrakhantsevaAA (commented):
Very good work!

sources/slack/__init__.py
VioletM: O, thanks, I was planning to check if this approach is less efficient, I'll rewrite this. Thanks for the link!

sources/slack/__init__.py
```diff
@@ -161,4 +234,36 @@ def logs_resource() -> Iterable[TDataItem]:
...
+                write_disposition="append",
+                columns={"blocks": {"data_type": "complex"}},
+            )(channel)
+
+            yield messages_channel
+            if replies:
+                yield messages_channel | dlt.transformer(
+                    get_thread_replies,
+                    name=channel_name + "_replies",
+                    table_name=partial(table_name_func, channel_name + "_replies"),
```
rudolfix, line 257: `table_name_func` needs `type` and/or `subtype` to be present. maybe you should add them to the `context` of get_pages?

rudolfix (approved):
this is really high quality code! also thanks for all the tests!

my comments:
1. current code to get replies will work with append write disposition. but in practice you'll probably won't see most of the replies - replies can be added to messages after we acquire them. those replies we'll never see. 
our solution is to always get last 7 or 14 days of messages by setting the start and end dates, in that case incremental loading is disables and we get the full range.

our code however sets write disposition to `append`. maybe it is a good default but imo we should change it to merge when end date is set.

you could write a test for that. try to run the pipeline with an explicit range. current code will duplicate the messages


2. this code will have a problem when running on Airflow - similar to the problem we had on sql_database. in the `@dlt.source` function we fetch channels list from slack. on Airflow this will happen every 30 seconds when running the DAG. maybe this is not a big problem? in sql_database it was solved for a case where table names are provided by the user (we are fetching table properties in the resource function - which happens at run time)

rudolfix (changes requested):
see my comments above

sources/slack/__init__.py
VioletM: These properties are already present in `page_data`.  

For example, so far I've seen only `thread_broadcast` subtype in replies and current source generates separate table `channel_name_replies_thread_broadcast`.

I've added `get("type", "")` though, in case there is some unexpected data.

Screenshot 2024-02-19 at 12 37 18

sources/slack/__init__.py
```diff
@@ -161,4 +234,36 @@ def logs_resource() -> Iterable[TDataItem]:
         ):
             yield page_data
 
-    yield logs_resource
+    yield from (channels_resource, users_resource, logs_resource)
+
+    if table_per_channel:
```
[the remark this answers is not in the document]
VioletM: As for p.2 I'm not sure. 
On the one hand in the current code, the channels will be fetched for each source init. On the other hand, if we move the fetching channels properties to resources, we would need to do this separately for messages and for replies. 
So I guess it depends on the use case. Since the current approach is what was done before, I was conservative with implementation. But if you have a strong opinion on this, I can move channels fetching to resources

rudolfix (approved):
LGTM! thanks for all the tests!
we should deploy this with our slack to see how our moving merge window works. ping me with the deployment script (maybe we add it to the demo pipeline)

PR #361, opened 2024-02-19
Rest API: Ends pagination if next page path is not in response.json()

willi-mueller:
# Tell us what you do here

References #313 

# More PR info

The Workable API does not include the json path to the next page for the last page of the response. Thus, this PR catches such cases.

### DIFF
--- a/sources/rest_api/paginators.py
+++ b/sources/rest_api/paginators.py
@@ -182,7 +182,10 @@ class JSONResponsePaginator(BaseNextUrlPaginator):
         self._records_accessor = create_nested_accessor(records_key)
 
     def update_state(self, response: Response):
-        self.next_reference = self._next_key_accessor(response.json())
+        try:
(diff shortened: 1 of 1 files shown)


sources/rest_api/paginators.py
```diff
@@ -182,7 +182,10 @@ def __init__(
         self._records_accessor = create_nested_accessor(records_key)
 
     def update_state(self, response: Response):
-        self.next_reference = self._next_key_accessor(response.json())
+        try:
+            self.next_reference = self._next_key_accessor(response.json())
+        except KeyError:
+           self.next_reference = None
```
burnash, line 188:
```suggestion
            self.next_reference = None
```

burnash (changes requested):
Thanks! Good catch. I was also wondering if this should be part of a accessor itself. But haven't made up my mind

PR #365, opened 2024-02-21
[REST Source] allow skipping http errors

willi-mueller:
# Tell us what you do here

# Relevant issue

issue #313 

# More PR info

APIs, such as Workable return HTTP error instead of empty lists of records in case the API call has no result. This PR proposes two ways of handling these cases on the level of the REST client, not the level of resources.

The alternatives are:

- a) set the text of the resource which then gets parsed to JSON as usual
- b) overwrite the `response.json()` method and return `None`

Both worked equally well in my tests, I'm just not sure if there is any advantage of either one.

### DIFF
--- a/sources/rest_api/__init__.py
+++ b/sources/rest_api/__init__.py
@@ -24,8 +24,9 @@ from dlt.common.schema.typing import (
     TWriteDisposition,
 )
 from dlt.extract.incremental import Incremental
-from dlt.extract.source import DltResource
+from dlt.extract.source import DltResource, DltSource
 from dlt.extract.typing import TTableHintTemplate
+from dlt.common import logger
 from dlt.sources.helpers.requests.retry import Client
 
 from .auth import BearerTokenAuth, AuthBase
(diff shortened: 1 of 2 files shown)


sources/rest_api/client.py
```diff
@@ -31,13 +29,16 @@ def __init__(
         headers: Optional[Dict[str, str]] = None,
         auth: Optional[AuthBase] = None,
         paginator: Optional[BasePaginator] = None,
-        session: BaseSession = None
+        # session: BaseSession = None
```
burnash, line 32:
```suggestion
        session: BaseSession = None
```

sources/rest_api/client.py
```diff
@@ -31,13 +29,16 @@ def __init__(
...
+        request_client: Client = None,
+        ignore_http_status_codes: Optional[Sequence[int]] = [],
     ) -> None:
         self.base_url = base_url
         self.headers = headers
         self.auth = auth
-        self.session = session or client.session
+        self.session = request_client.session
```
burnash, line 39:
```suggestion
        if session:
            self.session = session
        elif request_client:
            self.session = request_client.session
        else:
            self.session = Client().session
```

burnash (approved):
Thanks, @willi-mueller. I've added a couple of minor changes.

sources/rest_api/client.py
```diff
@@ -32,12 +30,20 @@ def __init__(
         auth: Optional[AuthBase] = None,
         paginator: Optional[BasePaginator] = None,
         session: BaseSession = None
+        request_client: Client = None,
+        ignore_http_status_codes: Optional[Sequence[int]] = [],
```
burnash, line 34: Totally missed that but it's a [dangerous default value](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/dangerous-default-value.html)

willi-mueller: Uh, thank you, I didn't know that! I can fix it so that we pacify the linter.

Issue #20, opened 2023-02-20
setup CI: run only tests from modified pipelines


rudolfix:
Improve all the workflows so only tests for pipelines modified in current PR are run.

**Motivation**
1. My biggest worry is that our API Credentials and accounts will get locked due to excessive CI traffic.
2. People should be able to merge their changes even if other pipelines are (temporarily) out of order.

**Tasks**
1. * [ ] find the affected files by comparing current branch to master
2. * [ ] identify pipelines and run tests only for them
3. * [ ] apply to all workflows

PR #366, opened 2024-02-22
fix(Workable): use append write disposition for candidate subresources

ethanve:
# Tell us what you do here
Given that there's no primary key set for candidate sub-resources, we should just use `append`

- [ ] implementing verified source (please link a relevant issue labeled as `verified source`)
- [ ] improving, documenting, or customizing an existing source (please link an issue or describe below)

# Relevant issue

issue #

# More PR info

### DIFF
--- a/sources/workable/__init__.py
+++ b/sources/workable/__init__.py
@@ -112,8 +112,8 @@ def workable_source(
         # A transformer functions that yield the activities and offers for each candidate.
         for sub_endpoint in DEFAULT_DETAILS["candidates"]:
             logging.info(
-                f"Loading additional data for 'candidates' from '{sub_endpoint}' in 'merge' mode."
+                f"Loading additional data for 'candidates' from '{sub_endpoint}' in 'append' mode."
             )
             yield candidates_resource | dlt.transformer(
-                name=f"candidates_{sub_endpoint}", write_disposition="merge"
+                name=f"candidates_{sub_endpoint}", write_disposition="append"
             )(_get_details)("candidates", sub_endpoint, "id")


sources/workable/__init__.py
```diff
@@ -112,8 +112,8 @@ def _get_details(
         # A transformer functions that yield the activities and offers for each candidate.
         for sub_endpoint in DEFAULT_DETAILS["candidates"]:
             logging.info(
-                f"Loading additional data for 'candidates' from '{sub_endpoint}' in 'merge' mode."
+                f"Loading additional data for 'candidates' from '{sub_endpoint}' in 'append' mode."
             )
             yield candidates_resource | dlt.transformer(
-                name=f"candidates_{sub_endpoint}", write_disposition="merge"
+                name=f"candidates_{sub_endpoint}", write_disposition="append"
```
AstrakhantsevaAA, line 118: hey! you are right, we missed primary_key (or merge_key) here, [candidates details response](https://workable.readme.io/reference/candidate-activities) has no id key to make it primary. I think it makes more sense to set write_disposition='replace', to avoid data duplication.

AstrakhantsevaAA (changes requested):

sources/workable/__init__.py
ethanve: @AstrakhantsevaAA given that candidates is incremental, it will replace all existing data with just updated ones. 
So for example,
1. Initial sync syncs 100 candidates and 10 offers
2. Candidate #20 now has an offer
3. Second pull will replace candidate offers with only a single offer for candidate #20

While merge may have duplicates, it's better than the alternative of replacing everything

AstrakhantsevaAA: right!

AstrakhantsevaAA (approved):
Thank you for your contribution!

AstrakhantsevaAA:
@ethanve can you rebase on master pls?

rudolfix (approved):
LGTM!

PR #371, opened 2024-02-23
[REST source] test case for dependent resource

willi-mueller:
# Tell us what you do here

# Relevant issue

issue #313 

# More PR info
Currently, failing test case

### DIFF
--- a/tests/rest_api/test_rest_api_source.py
+++ b/tests/rest_api/test_rest_api_source.py
@@ -1,7 +1,7 @@
 import dlt
 import pytest
 
-from sources.rest_api import rest_api_source
+from sources.rest_api import SinglePagePaginator, rest_api_source
 from tests.utils import ALL_DESTINATIONS, assert_load_info, load_table_counts
 
 
@@ -75,8 +75,9 @@ def test_dependent_resource(destination_name: str) -> None:
                 "name": "pokemon_list",
                 "endpoint": {
                     "path": "pokemon",
+                    "paginator": SinglePagePaginator(records_key="results"),
                     "params": {
-                        "limit": "2",
+                        "limit": 2,
                     },
                 },
             },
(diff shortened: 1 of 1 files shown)


tests/rest_api/test_rest_api_source.py
```diff
@@ -26,26 +28,82 @@ def test_rest_api_source(destination_name: str) -> None:
...
+                },
+            }
+        },
+        "resources": [
+            {
+                "name": "pokemon_list",
+                "endpoint": {
+                    "path": "pokemon",
+                    "params": {
+                        "limit": "2",
```
burnash, line 79: This will result in even more requests, since it's only 2 items per page. Limiting should be done with `add_limit` to the resource, but that'd may not work for transformer resource (the `pokemon`) below, similar to test connection function from another PR. Need to think how to make it work actually since it's a frequent use case: to test the full source load.

burnash (changes requested):

tests/rest_api/test_rest_api_source.py
willi-mueller: I agree that many APIs work like that. But the Pokemon API returns indeed only 2 records if you query it with `limit=2`. See my output:

```
curl 'https://pokeapi.co/api/v2/pokemon?limit=2' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   232    0   232    0     0    212      0 --:--:--  0:00:01 --:--:--   212
{
  "count": 1302,
  "next": "https://pokeapi.co/api/v2/pokemon?offset=2&limit=2",
  "previous": null,
  "results": [
    {
      "name": "bulbasaur",
      "url": "https://pokeapi.co/api/v2/pokemon/1/"
    },
    {
      "name": "ivysaur",
      "url": "https://pokeapi.co/api/v2/pokemon/2/"
    }
  ]
}
```

The challenge I see is that calling `add_limit(2)` on a paginated resource means to stop yielding after two __pages__ of records, not after two records. Thus, it is unpredictable how many records will be contained in two pages because it depends on the request and the database.

**Example 1:** Workable account A1 can have 3 jobs and account A2 has 300 jobs. When the page_size is 100 then `add_limit(2)` will yield 3 jobs for the first account and 200 for the second. Thus, `add_limit(n)` yields `n <= length(records)`

**Example 2:** Consider the same accounts A1 and A2 as above. When the page_size is 100 then `add_limit(1)` will yield 3 jobs for the first account and 100 for the second. Thus, `add_limit(n)` yields `n >= length(records)` for A1 and `n <= length(records)` for A2. I was confused when I discovered that.

Test code often serves as an example of proper usage. Thus, I'd consider it misleading to use `add_limit(n)` in our example code as long as `n != length(records)` because for non-paginated resources `add_limit(n)` yields `n == length(records)`.

What are your thoughts?

burnash:
Thank you, @willi-mueller

PR #384, opened 2024-02-29
D#/add pandas pyarrow verified source

sh-rp:
# Tell us what you do here

This PR adds a simple starting point for creating pipelines with arrow tables or panda dataframes


rudolfix (changes requested):
this is not how we want our users to use pandas or arrow. the main use case is to pass this as is to the run method. it is one liner. we do not need any verified source for that

sultaniman:
Think about templating in the dlt init.

PR #392, opened 2024-03-04
Postgres replication

jorritsandbrink:
# Tell us what you do here

- [X] implementing verified source (please link a relevant issue labeled as `verified source`)
- [ ] improving, documenting, or customizing an existing source (please link an issue or describe below)

# Relevant issue

https://github.com/dlt-hub/dlt/issues/933

# More PR info
Adds initial support for postgres replication. Some things are still missing, but this is a good time to get feedback.

- Implements replication functionality based on [logical decoding](https://www.postgresql.org/docs/current/logicaldecoding.html)
- Processes changes in batch (not streaming)
- Uses built-in `pgoutput` plugin
- Uses `psycopg2`'s support for logical [replication](https://www.psycopg.org/docs/advanced.html#replication-support)—this streams messages from `pgoutput` into Python in an endless loop
- Uses `pypgoutput` to decode `pgoutput`'s binary messages—the library's functionality to consume messages and transform them into "change events" (Pydantic models) is not used because it only works on [Linux](https://github.com/dgea005/pypgoutput/issues/16)
- Uses "text mode", meaning data is provided as string values. These string values are coerced into `dlt`-compatible Python objects, e.g. the string `"t"` becomes the boolean `True`. "Binary mode" would be faster, but less [robust](https://www.postgresql.org/docs/14/protocol-logical-replication.html#PROTOCOL-LOGICAL-REPLICATION-PARAMS).
- ~~Relies on a dedicated replication slot and publication for a table. I.e. two tables means two slots and two publications. This provides granular control and does not intruduce significant overhead if I'm not mistaken.~~ No longer the case, changed because of user [feedback](https://github.com/dlt-hub/verified-sources/pull/392#issuecomment-1992500024). It is now possible to replicate one table, multiple tables, or an entire schema using a single publication.
- Maintains resource-scoped state to keep track of progress. At beginning of run, last-seen LSN of previous run is retrieved from state and used to advance the replication slot (flush messages from it).
- ~~Adds two resource types: `table_snapshot` for initial load, and `table_changes` for CDC.~~
  - ~~`table_snapshot` persists the state of the table in the snapshot that gets exported when creating a replication slot into a physical table, and then uses `sql_table` resource to do the rest~~
  - ~~`table_changes` generates and yields "data items" (`TDataItem`) and "metadata items" (`DataItemWithMeta`) from decoded replication messages. Items are first stored in-memory in a list, before they are yielded from this list.~~
- Handles both initial load and change propagation.
  - Provides `init_replication` to setup a replication slot and publication. This function optionally persists snapshot tables representing the state of the exact moment the replication slot got created. It then returns `sql_table` resources to enable an initial load. Users do not need to use `init_replication`—they can create a slot and publication in any other way they see fit. 
  - Provides `replication_resource` to create a `DltResource` that consumes a slot/publication and generates data items with metadata (`DataItemWithMeta`). It dispatches data to [multiple](https://dlthub.com/docs/general-usage/resource#dispatch-data-to-many-tables) tables if the publication publishes changes for multiple tables.
- Uses `include_columns` argument to exclude any columns not provided as input (or includes all columns if not provided) 
- ~~Organizes code in subfolder under `sql_database`: `/sources/sql_database/pg_replication`~~ Moved to its own top-level folder.

**What's not (yet) included:**
- ~~Chunking mechanism to limit batch size in `table_changes`~~ _implemented_
- ~~`DltSource` to handle multiple tables / an entire database~~ _no longer applies—multiple tables are now handled at the resource level_
- Support for the `truncate` operation
- ~~Perhaps some more data type mapping~~ _done—common types are handled and exotic types default to `text`_
- ~~Deletion of snapshot table after it has been consumed~~ _not implemented—couldn't find a good way to do this_
- ~~Example pipeline~~ _done_
- ~~More tests~~ _done_
- User docs


sources/sql_database/pg_replication/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+          slot is consumed. If not provided, all messages in the slot are consumed,
+          ensuring all new changes in the source table are included.
+
+        Returns:
+            DltResource that yields data items for changes in the postgres table.
+    """
+    return dlt.resource(
+        table_replication_items,
+        name=table,
+        write_disposition="merge",
```
rudolfix, line 55: for tables with INSERT only (ie. logs) `append` will also work?

sources/sql_database/pg_replication/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        primary_key=primary_key,
```
rudolfix, line 56: isn't that coming in every message? what happens if someone changes the primary key on a table? is that even possible?

sources/sql_database/pg_replication/helpers.py
```diff
@@ -0,0 +1,528 @@
+from typing import Optional, Dict, Iterator, Union, List, Tuple, Sequence, Any
+from dlt.common.pendulum import pendulum
+from copy import deepcopy
+from dataclasses import dataclass, field
+
+import psycopg2
```
rudolfix, line 6: there will be people using `sql_database` that are not using postgres replication. we are not surfacing this into top level init? anyway... I think we may land it in separate verified source (if we do not share anything with sql alchemy source)

sources/sql_database/pg_replication/helpers.py
```diff
@@ -0,0 +1,528 @@
...
+    """
+    col_str = "*"
+    if include_columns is not None:
+        col_str = ", ".join(map(escape_postgres_identifier, include_columns))
+    snapshot_table_name = f"{table}_snapshot_{snapshot_name}"
+    snapshot_qual_name = _make_qualified_table_name(snapshot_table_name, schema)
+    qual_name = _make_qualified_table_name(table, schema)
+    cur.execute(
+        f"""
+        START TRANSACTION ISOLATION LEVEL REPEATABLE READ;
```
rudolfix, line 237: why do we need repeatable read?

sources/sql_database/pg_replication/helpers.py
```diff
@@ -0,0 +1,528 @@
...
+
+
+def persist_snapshot_table(
+    snapshot_name: str,
+    table: str,
+    schema: str,
+    cur: psycopg2.extensions.cursor,
+    include_columns: Optional[Sequence[str]] = None,
+) -> str:
+    """Persists exported snapshot table state.
```
rudolfix, line 225: this is a table that user should load first right? the thing that I do not understand is that we do not get `upto_lsn` here. how do we make sure we do not lose data?
1. we create a table with a current snapshot
2. we replicate initial table
3. we start replication from a slot
what about changed between 2 and 3? I think the initial table and slot must be created in such a way that any change after initial table was created is accumulating in the slot

sources/sql_database/pg_replication/helpers.py
```diff
@@ -0,0 +1,528 @@
...
+        self.include_columns = include_columns
+
+        # data_items attribute maintains all data items
+        self.data_items: List[Union[TDataItem, DataItemWithMeta]] = []
+        # other attributes only maintain last-seen values
+        self.last_table_schema: TTableSchema
+        self.last_commit_ts: pendulum.DateTime
+        self.last_commit_lsn = None
+
+    def __call__(self, msg: ReplicationMessage) -> None:
```
rudolfix, line 422: do I understand correctly that we get a message one by one? I hope that our python code will keep up. also we need some batching (as you say in the ticket)

sources/sql_database/pg_replication/helpers.py
```diff
@@ -0,0 +1,528 @@
...
+
+    def process_relation(self, decoded_msg: Relation) -> None:
+        """Processes a replication message of type Relation.
+
+        Stores table schema information from Relation message in object state
+        and adds meta data item to the relation's list of data items to update
+        the table schema.
+        """
+        # store table schema information
+        columns = {c.name: _to_dlt_column_schema(c) for c in decoded_msg.columns}
```
rudolfix, line 464: when do we get this message? is it coming with every change? it looked like it in jsonpgout

rudolfix (commented):
this looks super solid already. I had a few questions, mostly about the transition from syncing the snapshot table to when we start replication from a slot (do we lose data?)

do you think that going for binary pgout is a good idea after all? I think json one will be simpler and faster
- messages are created AFAIK on the server so you do not need to decode them
- the structure is almost good to be loaded
we can actually keep both but we add a lot of code that should probably be tested

sources/sql_database/pg_replication/__init__.py
jorritsandbrink: Yes, should work, didn't think of that. I can fetch details about the publication using `pg_publication` [catalog](https://www.postgresql.org/docs/current/catalog-pg-publication.html#CATALOG-PG-PUBLICATION) and use `append` if it only publishes `insert`. Will be more efficient than `merge`.

sources/sql_database/pg_replication/__init__.py
jorritsandbrink: Primary key info is included in each message of type `Relation`. Normally we get one of those messages in each `run` (preceding the DML messages), unless the table schema has changed, then we will receive an additional `Relation` message for each change. Our logic generates a metadata item (with `gen_meta_data_item`) for each `Relation` message. The `table_changes` resource yields items as follows:

`DataItemWithMeta` ← we get this always, also if the table schema hasn't changed—contains PK info as part of `columns`
`TDataItem` ← we get one for each DML operation (`insert`, `update`, or `delete`)
...
`TDataItem`
`DataItemWithMeta` ← table schema has changed—contains PK info as part of `columns`
`TDataItem`
...
`TDataItem`
etc.

I will include tests for schema changes.

sources/sql_database/pg_replication/helpers.py
jorritsandbrink: Separate source makes sense to me. Only thing is that the `table_snapshot` resource for initial loading uses the `sql_table` resource under the hood. How should we handle one verified source depending on another?

sources/sql_database/pg_replication/helpers.py
jorritsandbrink: Doesn't work with `READ COMMITTED`:
image

Also mentioned on SO: https://stackoverflow.com/a/76284206

sources/sql_database/pg_replication/helpers.py
jorritsandbrink: Changes start getting accumulated the moment the slot is created. Slot creation happens as part of `init_table_replication`, so any changes occuring after that function was called will be captured. The `start_replication` method merely starts a session used to stream changes into Python, it is **not** the trigger to start capturing changes in the slot. The test I included also demonstrates this: it first calls `init_table_replication` (which creates the slot and returns the snapshot) and only after that it makes changes to the table—we see that those changes are indeed included in `table_changes` which reads from the slot.

sources/sql_database/pg_replication/helpers.py
jorritsandbrink: Yeah, messages come in one by one. I will implement batching and then do some load tests to see if performance is acceptable for larger volumes.

sources/sql_database/pg_replication/helpers.py
jorritsandbrink: See my previous [comment](https://github.com/dlt-hub/verified-sources/pull/392#discussion_r1514523102).

jorritsandbrink:
@rudolfix see my replies on your comments.

Regarding `pgoutput` vs `wal2json`:
- my main consideration to go with `pypgoutput` first was convenience for the user (no additional setup needed)—for what it's worth, Debezium also [supports](https://debezium.io/documentation/reference/stable/postgres-plugins.html) `pypgoutput` but not `wal2json`
- another reason is that publications only work with `pgoutput`, meaning that additional client-side filtering is needed (selecting the right table, and the right DML operation if you want to only propagate inserts for example), offsetting some of the benefits of server-side decoding—see this [SO](https://stackoverflow.com/questions/70120968/logical-replication-with-psycopg2-and-publishers) on the topic
- `wal2json` will probably be a little simpler, but not a lot—the message objects generated by `pypgoutput` (e.g. `Insert`, `Update`, ...) are almost as easy to work with as `wal2json`'s output

That being said, `wal2json` might be better (necessary) when data volumns are large. I'd suggest we go with `pgoutput` first, then add `wal2json` support later if needed.

jorritsandbrink:
image
Thanks for your feedback Martin, let's continue the discussion here for visibility.

- I think using one slot + publication for multiple tables can be enabled with some minor modifications to the code.
- If you have multiple tables in a single publication, it would make sense (be most efficient) if you process all tables concurrently while iterating over the replication messages once.
- The alternative would be to iterate over the replication messages multiple times, once for each table, and skipping all messages that aren't relevant for that table. Less efficient, but resources/tables can be processed independently.

Can you process individual tables in fivetran / peardb / the other systems you're referring to, or do you have to process all of them simultaneously?

rudolfix:
@jorritsandbrink it should be easy. if there > 1 table we still have one LSN with which we track new messages right? if so the change is trivial: you can specify a table name which is resolved dynamically:
https://dlthub.com/docs/general-usage/resource#dispatch-data-to-many-tables

drawbacks:
1. lambda needs table name in the data items. but you can also remove that name in the same lambda if we do not need this in the final table.
2.  full refresh of a table requires a reset of all tables in publication, right? also initial load will generate many `sql_table` resources

sources/sql_database/pg_replication/helpers.py
rudolfix: this is amazing, thanks!

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+                data_type=schema["data_type"],
+                byte1=data.col_data_category,
+                for_delete=for_delete,
+            )
+            for (schema, data) in pairs
+        }
+        data_item["lsn"] = lsn
+        if for_delete:
+            data_item["deleted_ts"] = commit_ts
+        return dlt.mark.with_hints(
```
rudolfix, line 748: this is all correct but will be very slow. what happens currently is
- you yield items one by one
- each item has full hints with it.
this will generate a lot of unnecessary processing in the pipeline.

what you should do instead (I think it is not very complicated):
- buffer data items per table name to be able to yield them together
- keep schema changes per table and yield them once
you yield stuff when we reach batch size:
- for each table and schema change
`yield dlt.mark.with_hints([], dlt.mark.make_hints(...)` - you can yield empty list just to update hints
- for each table and data items
`yield dlt.mark.with_table_name(data_items[table_name], table_name)`
where `data_items[table_name]` is a list of data items collected for a particular table

the speedup will be significant. and this code structure also makes easier to yield arrow tables if we want to be even faster
tell me WDYT

rudolfix (changes requested):
this is so good! I have a few code improvements and one optimization that will make this way faster. I love that we have so many tests.
it also looks to me that https://github.com/dlt-hub/dlt/issues/1105 is crucial to support replication of many tables from a single resource

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+) -> Iterator[Union[TDataItem, DataItemWithMeta]]:
+    """Yields data items from generator.
+
+    Maintains LSN of last consumed message in state to track progress.
+    At start of the run, advances the slot upto last consumed message in previous run.
+    Processes in batches to limit memory usage.
+    """
+    # start where we left off in previous run
+    resource_name = _gen_replication_resource_name(slot_name, pub_name)
+    start_lsn = dlt.current.resource_state(resource_name).get("last_commit_lsn", 0)
```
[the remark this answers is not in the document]
jorritsandbrink: I'd say it's simpler without `Incremental`, didn't see a reason to use it.

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+                "Truncate replication messages are ignored."
+            )
+
+    def process_relation(self, decoded_msg: Relation) -> None:
+        """Processes a replication message of type Relation.
+
+        Stores table schema information from Relation message in object state.
+        """
+        # store table schema information
+        columns = {c.name: _to_dlt_column_schema(c) for c in decoded_msg.columns}
```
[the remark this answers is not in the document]
jorritsandbrink: `decoded_msg.columns` has all columns, modified or not.

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+    """Returns primary key column(s) for postgres table.
+
+    Returns None if no primary key columns exist.
+    """
+    qual_name = _make_qualified_table_name(table_name, schema_name)
+    cur = _get_conn(credentials).cursor()
+    # https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns
+    cur.execute(
+        f"""
+        SELECT a.attname
```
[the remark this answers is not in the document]
jorritsandbrink: The `primary_key` hint on the `sql_table` resource gives an empty list, so perhaps something's going wrong there:
image

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+
+try:
+    from ..sql_database import sql_table  # type: ignore[import-untyped]
+except Exception:
+    from sql_database import sql_table
+
+from .schema_types import _to_dlt_column_schema, _to_dlt_val
+
+
+@dlt.sources.config.with_config(sections=("sources", "pg_replication"))
```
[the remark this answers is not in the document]
jorritsandbrink: Yes, but it only returns resources on the first call (when it creates the slot) and only if `persist_snapshots` is `True`. So I'd say it's primarily a helper to set up replication, rather than a source.

sources/pg_replication/README.md
```diff
@@ -0,0 +1,60 @@
...
+Resources that can be loaded using this verified source are:
+
+| Name                 | Description                                     |
+|----------------------|-------------------------------------------------|
+| replication_resource | Load published messages from a replication slot |
+
+## Initialize the pipeline
+
+```bash
+dlt init pg_replication duckdb
```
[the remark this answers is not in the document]
jorritsandbrink: Extended the README: https://github.com/dlt-hub/verified-sources/pull/392/commits/57b5e1eedf896b32669421914ba090216a17cf6d

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+
+def replication_items(
+    credentials: ConnectionStringCredentials,
+    slot_name: str,
+    pub_name: str,
+    include_columns: Optional[Dict[str, Sequence[str]]] = None,
+    columns: Optional[Dict[str, TTableHintTemplate[TAnySchemaColumns]]] = None,
+    target_batch_size: int = 1000,
+    flush_slot: bool = True,
+    write_disposition: TWriteDisposition = "append",  # TODO: remove after https://github.com/dlt-hub/dlt/issues/1031 has been released
```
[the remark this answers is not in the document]
jorritsandbrink: I tested with that version but the issue still seems to be there.
image

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,757 @@
...
+            options=options,
+            upto_lsn=upto_lsn,
+            start_lsn=start_lsn,
+            target_batch_size=target_batch_size,
+            include_columns=include_columns,
+            columns=columns,
+            write_disposition=write_disposition,  # TODO: remove after https://github.com/dlt-hub/dlt/issues/1031 has been released
+        )
+        yield from gen
+        if gen.generated_all:
```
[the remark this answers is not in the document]
jorritsandbrink: I think in this case we need the cleanup code to only execute if the batch has been yielded completely. We can't split at any LSN, it needs to be the LSN of a `Commit` message. This is because transactions need to be processed atomically, and transactions can consist of multiple messages (e.g. 100 inserted records = 100 messages). The code has been set up such that batch boundaries are always `Commit` messages. Replication can also only start (`start_lsn`) at the beginning of a transaction. So if we have a transaction of 100 inserts, and `add_limit(50)` we don't want to persist resource state, because we haven't yielded the entire transaction.

sources/pg_replication/helpers.py
jorritsandbrink: Much better indeed! I started building this, but faced the "hint accumulation [issue](https://github.com/dlt-hub/dlt/issues/1105)" that's blocking progress. This is on hold until that issue has been completed.

sources/pg_replication/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+    TWriteDisposition,
+)
+from dlt.sources.credentials import ConnectionStringCredentials
+from dlt.extract.resource import DltResource
+from dlt.extract.typing import TTableHintTemplate
+
+from .helpers import _gen_replication_resource_name, get_pub_ops, replication_items
+
+
+@dlt.sources.config.with_config(sections=("sources", "pg_replication"))
```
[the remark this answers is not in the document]
jorritsandbrink: Done in https://github.com/dlt-hub/verified-sources/pull/392/commits/eec75f02f756e063c20cfee8ea816adbe0d90af7

sources/pg_replication/helpers.py
jorritsandbrink: This has now been implemented in https://github.com/dlt-hub/verified-sources/pull/392/commits/5636e078c9f4b0085549d13cdc29bb5e17224eae.

jorritsandbrink:
@rudolfix I've addressed all your comments, please review.

To be done: increase `dlt` version in `requirements.txt` to include PR [1127](https://github.com/dlt-hub/dlt/pull/1127).

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,752 @@
...
+        """Processes replication message of type Insert, Update, or Delete.
+
+        Adds data item for inserted/updated/deleted record to instance attribute.
+        """
+        if isinstance(decoded_msg, (Insert, Update)):
+            column_data = decoded_msg.new_tuple.column_data
+        elif isinstance(decoded_msg, Delete):
+            column_data = decoded_msg.old_tuple.column_data
+        table_name = self.last_table_schema[decoded_msg.relation_id]["name"]
+        data_item = self.gen_data_item(
```
rudolfix, line 717: this looks good.

sources/pg_replication/helpers.py
```diff
@@ -0,0 +1,752 @@
...
+        """Processes encoded replication message.
+
+        Identifies message type and decodes accordingly.
+        Message treatment is different for various message types.
+        Breaks out of stream with StopReplication exception when
+        - `upto_lsn` is reached
+        - `target_batch_size` is reached
+        - a table's schema has changed
+        """
+        op = (msg.payload[:1]).decode("utf-8")
```
rudolfix, line 606: you could compare binary strings without decoding to make it a little bit faster. also you could put the more frequent (I/U/D) operations to the top

rudolfix (approved):
LGTM! I have one question to the missing PK in SqlAlchemy. let's take a look. PK should be really there.

I expect that for large loads this will be slow. we are doing a lot of operations per row. also the pgout decoder is quite heavy.

sources/pg_replication/helpers.py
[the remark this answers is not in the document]
jorritsandbrink: Okay I get it now. `sql_table` creates a resource for the snapshot table, which doesn't have a primary key, because it is created with a CTAS statement:

```python
  cur.execute(
      f"""
      START TRANSACTION ISOLATION LEVEL REPEATABLE READ;
      SET TRANSACTION SNAPSHOT '{snapshot_name}';
      CREATE TABLE {snapshot_qual_name} AS SELECT {col_str} FROM {qual_name};
  """
  )
```

I point `_get_pk` at the original table (`table_name` instead of `snapshot_table_name`), which does have a primary key, hence the difference.

sources/pg_replication/helpers.py
jorritsandbrink: Implemented in https://github.com/dlt-hub/verified-sources/pull/392/commits/7bd211b7d70de759c0809a0010e3de3fe5b66d4d.

rudolfix:
@jorritsandbrink I finally enabled tests on our CI, and most of them are passing but:
- this one consistently fails
```
with src_pl.sql_client() as c:
            qual_name = src_pl.sql_client().make_qualified_table_name("items")
            c.execute_sql(f"UPDATE {qual_name} SET foo = 'baz' WHERE id = 2;")
            c.execute_sql(f"DELETE FROM {qual_name} WHERE id = 2;")
        extract_info = dest_pl.extract(changes)
>       assert extract_info.asdict()["job_metrics"] == []
E       AssertionError: assert [{'created': ...e': 501, ...}] == []
E         Left contains one more item: {'created': 1713123289.1342065, 'extract_idx': 1, 'file_path': '/home/rudolfix/src/pipelines/_storage/.dlt/pipelines/d...lize/a93aba1460a1d099/1713123287.6514962/new_jobs/_dlt_pipeline_state.bfcae69f3b.0.typed-jsonl', 'file_size': 501, ...}
E         Full diff:
E           [
E         -  ,
E         +  {'created': 1713123289.1342065,
E         +   'extract_idx': 1,
E         +   'file_path': '/home/rudolfix/src/pipelines/_storage/.dlt/pipelines/dest_pl/normalize/a93aba1460a1d099/1713123287.6514962/new_jobs/_dlt_pipeline_state.bfcae69f3b.0.typed-jsonl',...
E
E         ...Full output truncated (6 lines hidden), use '-vv' to show
```
- those require postgres 15 while we are on 13 on CI. maybe you could take a look? is there a way to use the old syntax?
```
try:
>           cur.execute(
                f"ALTER PUBLICATION {esc_pub_name} ADD TABLES IN SCHEMA {esc_schema_name};"
            )
E           psycopg2.errors.SyntaxError: syntax error at or near "TABLES"
E           LINE 1: ALTER PUBLICATION "test_pub6589482f" ADD TABLES IN SCHEMA "s...
```

jorritsandbrink:
> @jorritsandbrink I finally enabled tests on our CI, and most of them are passing but:
> 
> * this one consistently fails
> 
> ```
> with src_pl.sql_client() as c:
>             qual_name = src_pl.sql_client().make_qualified_table_name("items")
>             c.execute_sql(f"UPDATE {qual_name} SET foo = 'baz' WHERE id = 2;")
>             c.execute_sql(f"DELETE FROM {qual_name} WHERE id = 2;")
>         extract_info = dest_pl.extract(changes)
> >       assert extract_info.asdict()["job_metrics"] == []
> E       AssertionError: assert [{'created': ...e': 501, ...}] == []
> E         Left contains one more item: {'created': 1713123289.1342065, 'extract_idx': 1, 'file_path': '/home/rudolfix/src/pipelines/_storage/.dlt/pipelines/d...lize/a93aba1460a1d099/1713123287.6514962/new_jobs/_dlt_pipeline_state.bfcae69f3b.0.typed-jsonl', 'file_size': 501, ...}
> E         Full diff:
> E           [
> E         -  ,
> E         +  {'created': 1713123289.1342065,
> E         +   'extract_idx': 1,
> E         +   'file_path': '/home/rudolfix/src/pipelines/_storage/.dlt/pipelines/dest_pl/normalize/a93aba1460a1d099/1713123287.6514962/new_jobs/_dlt_pipeline_state.bfcae69f3b.0.typed-jsonl',...
> E
> E         ...Full output truncated (6 lines hidden), use '-vv' to show
> ```
> 
> * those require postgres 15 while we are on 13 on CI. maybe you could take a look? is there a way to use the old syntax?
> 
> ```
> try:
> >           cur.execute(
>                 f"ALTER PUBLICATION {esc_pub_name} ADD TABLES IN SCHEMA {esc_schema_name};"
>             )
> E           psycopg2.errors.SyntaxError: syntax error at or near "TABLES"
> E           LINE 1: ALTER PUBLICATION "test_pub6589482f" ADD TABLES IN SCHEMA "s...
> ```

@rudolfix The first issue is actually also version related. I was testing on Postgres 16 locally, but have been able to reproduce both issues on Postgres 13.

1) It seems Postgres 13 publishes "empty transactions" for updates/deletes when they are excluded from the `publish` publication parameter (e.g. when `publish = 'insert'`). Postgres 16 does not do this. As a result we do find messages to process (the "empty transactions") when we've done an update or delete, even though we told Postgres we're not interested in them. `last_commit_lsn` in resource state needs to be updated accordingly. An item gets extracted for `_dlt_pipeline_state` because of the state update, where our test asserted nothing is extracted. Solved by making the test more specific and asserting nothing gets extracted for the `items` table: https://github.com/dlt-hub/verified-sources/pull/392/commits/34610b634f29348d6362a2f26fa946ed3b93cf37

2) Not really feasible to make this work for older Postgres versions. I could fetch all tables from the schema and add them one by one, but that wouldn't accomodate the case where a table gets added to the schema later. To keep things clean, I introduced a Postgres version requirement for schema replication instead: https://github.com/dlt-hub/verified-sources/pull/392/commits/7a070453e93bed4c7863b5ee82f681dd12e909e5

rudolfix:
@jorritsandbrink should we spawn another postgres instance just to test replication? I do not want to make it too complicated and I'm totally fine with 15. version.

jorritsandbrink:
@rudolfix yes, using a separate instance for replication sounds good.

rudolfix (approved):
LGTM!

PR #393, opened 2024-03-05
Update the issue templates

VioletM:
### Tell us what you do here

- [x] Update the existing issue templates
- [x] Remove the new source template, cause it can be a pull request right away


burnash (approved):
Thanks @VioletM, looks good.

PR #397, opened 2024-03-06
makes openapi friendly auth

rudolfix:
1. keeps AuthBase compatibility
2. aligns naming with openAPI spec
3. makes auths also dlt configurations


burnash (approved):
LGTM

PR #398, opened 2024-03-08
feat(kafka): autostop on empty topics

IlyaFaer:


tests/kafka/test_kafka.py
```diff
@@ -261,6 +274,21 @@ def test_kafka_incremental_read(kafka_producer, kafka_topics):
...
+        destination="postgres",
+        dataset_name="kafka_test_data",
+        full_refresh=True,
+    )
+
+    resource = kafka_consumer(empty_topic, batch_timeout=0.01)
+    pipeline.run(resource)
+
+    with pytest.raises(DatabaseUndefinedRelation):
+        load_table_counts(pipeline, empty_topic)
```
IlyaFaer, line 289: First of all, we're checking that it doesn't hang forever

sources/kafka/__init__.py
```diff
@@ -82,10 +83,19 @@ def kafka_consumer(
 
     # read messages up to the maximum offsets,
     # not waiting for new messages
+    attempts = 0
     with closing(consumer):
         while tracker.has_unread:
+            if attempts == 30:
```
IlyaFaer, line 89: Better to use higher number of attempts, as sometimes, when we just start reading, it takes several seconds before the consumer is directed to a broker.

PR #399, opened 2024-03-08
[REST CLIENT] yields data pages with requests context

rudolfix:
### Tell us what you do here
1. guesses the single entity endpoints from the path when creating resources
2. NOTE: data_selector is JSONPath now, backward compatile with string names
3. yielded page data is an extended list with requests context
4. all tests are passing

### DIFF
--- a/sources/rest_api/__init__.py
+++ b/sources/rest_api/__init__.py
@@ -18,12 +18,13 @@ import dlt
 from dlt.common.validation import validate_dict
 from dlt.extract.incremental import Incremental
 from dlt.extract.source import DltResource, DltSource
-from dlt.common import logger
+from dlt.common import logger, jsonpath
 from dlt.common.utils import update_dict_nested
 from dlt.common.typing import TSecretStrValue
 
-from .auth import BearerTokenAuth, AuthBase
+from .auth import BearerTokenAuth, AuthConfigBase
 from .client import RESTClient
+from .detector import single_entity_path
 from .paginators import (
     BasePaginator,
     HeaderLinkPaginator,
@@ -41,7 +42,6 @@ from .typing import (
     EndpointResource,
     RESTAPIConfig,
 )
-from .utils import remove_key
 
 
 PAGINATOR_MAP: Dict[str, Type[BasePaginator]] = {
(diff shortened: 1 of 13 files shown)


tests/rest_api/test_rest_api_source.py
```diff
@@ -87,20 +88,20 @@ def test_dependent_resource(destination_name: str) -> None:
                             "field": "name",
                         },
                     },
-                    "paginator": "single_page",
+
```
burnash, line 91: Nice!

sources/rest_api/typing.py
```diff
@@ -81,13 +79,19 @@ class EndpointResource(TypedDict, total=False):
     incremental: Optional[Incremental[Any]]
     table_format: Optional[TTableHintTemplate[TTableFormat]]
     include_from_parent: Optional[List[str]]
+    selected: Optional[bool]
+
+
+# NOTE: redefining properties of TypedDict is not allowed
+class EndpointResource(EndpointResourceBase, total=False):
```
burnash, line 86: Nice, thanks!

sources/rest_api/client.py
```diff
@@ -94,39 +118,37 @@ def request(
...
         json: Optional[Dict[str, Any]] = None,
-        auth: Optional[AuthBase] = None,
+        auth: Optional[AuthConfigBase] = None,
         paginator: Optional[BasePaginator] = None,
-        data_selector: Optional[Union[str, List[str]]] = None,
+        data_selector: Optional[jsonpath.TJsonPath] = None,
         response_actions: Optional[List[Dict[str, Any]]] = None,
         hooks: Optional[Dict[str, Any]] = None,
-    ) -> Generator[Any, None, None]:
+    ) -> Iterator[PageData[Any]]:
```
burnash, line 141: I believe the correct type is Generator, not Iterator. Even though it's allowed to type it as an iterator:

```
>>> def infinite_stream(start: int):
...     while True:
...         yield start
...         start += 1
...
>>> type(infinite_stream(1))
<class 'generator'>
```

sources/rest_api/client.py
```diff
@@ -149,20 +171,30 @@ def paginate(
...
-                yield data[0] if len(data) == 1 else data
-            else:
-                yield find_records(response.json())
-
+            data = self.extract_response(response, data_selector)
             paginator.update_state(response)
             paginator.update_request(request)
 
+
+            # yield data with context
```
burnash, line 179:
```suggestion
```

sources/rest_api/detector.py
```diff
@@ -29,6 +31,11 @@
 NEXT_PAGE_DICT_KEY_PATTERNS = {"href", "url"}
 
 
+def single_entity_path(path: str) -> bool:
```
burnash, line 34: Should this handle trailing slashes (e.g. `"/users/{user_id}/"`) ? I'll add a couple of tests in a PR for this. Most likely we need some high level default treatment for those.

burnash (changes requested):
Thanks for the PR. Please see comments.

burnash (approved):

burnash:
@rudolfix I'm merging it, let's address the trailing slash in a follow-up request.

sources/rest_api/__init__.py
```diff
@@ -73,17 +73,17 @@ def create_paginator(paginator_config: PaginatorType) -> Optional[BasePaginator]
         return paginator_class()
 
     if isinstance(paginator_config, dict):
-        paginator_type = paginator_config.get("type", "auto")
+        paginator_type = paginator_config.pop("type", "auto")
         paginator_class = get_paginator_class(paginator_type)
-        return paginator_class(**remove_key(paginator_config, "type"))
+        return paginator_class(paginator_config)
```
burnash, line 78: **paginator_config

PR #400, opened 2024-03-09
Sthor/api helper updates

steinitzu:
### Tell us what you do here
Adding a few changes to the rest api source:

* Support json path in "next url" paginatior
* Add "cursor" paginator
* Use "parse_native_value" instead of requiring args in init for the standard auth classes, they were not reading from config before
* Uppercase HTTP method names as is standard

I'm currently testing it out with openapi gen, so might add some more

### DIFF
--- a/sources/rest_api/auth.py
+++ b/sources/rest_api/auth.py
@@ -1,6 +1,6 @@
 from base64 import b64encode
 import math
-from typing import Dict, Final, Literal, Optional, Union
+from typing import Dict, Final, Literal, Optional, Union, Any, cast, Iterable
 from dlt.sources.helpers import requests
 from requests.auth import AuthBase
 from requests import PreparedRequest  # noqa: I251
@@ -15,6 +15,7 @@ from dlt.common import logger
 
 from dlt.common.configuration.specs.base_configuration import configspec
 from dlt.common.configuration.specs import CredentialsConfiguration
+from dlt.common.configuration.specs.exceptions import NativeValueError
 from dlt.common.typing import TSecretStrValue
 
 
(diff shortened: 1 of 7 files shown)


sources/rest_api/paginators.py
```diff
@@ -133,19 +134,46 @@ class JSONResponsePaginator(BaseNextUrlPaginator):
...
         Args:
-            next_key (str, optional): The key in the JSON response that
-                contains the next page URL. Defaults to 'next'.
+            next_url_path: The JSON path to the key that contains the next page URL in the response.
+                Defaults to 'next'.
         """
         super().__init__()
-        self.next_key = next_key
-        self._next_key_accessor = create_nested_accessor(next_key)
+        self.next_url_path = jsonpath.compile_path(next_url_path)
```
burnash, line 145: @steinitzu thanks for the suggestion. Could you please elaborate on the real use case for jsonpath here? I can imagine that there might be a non-trivial path required to extract data records (e.g., nested dictionaries plus list indices). Is this the case for the data at the next url?

sources/rest_api/auth.py
```diff
@@ -33,8 +38,15 @@ class BearerTokenAuth(AuthConfigBase):
     scheme: Literal["bearer"] = "bearer"
     token: TSecretStrValue
 
-    def __init__(self, token: TSecretStrValue) -> None:
```
rudolfix, line 36: OK I see we have to do that because we have this weird constructor in `CredentialsConfiguration` which takes native value as input. I want to get rid of that... So we'll add those again soon. Maybe you could leave them commented?

sources/rest_api/paginators.py
```diff
@@ -133,19 +134,46 @@ class JSONResponsePaginator(BaseNextUrlPaginator):
...
     def update_state(self, response: Response) -> None:
-        try:
-            self.next_reference = self._next_key_accessor(response.json())
-        except KeyError:
-            self.next_reference = None
+        values = jsonpath.find_values(self.next_url_path, response.json())
+        self._next_reference = values[0] if values else None
+
+
+class JSONResponseCursorPaginator(BasePaginator):
```
rudolfix, line 152: cool. but this one we won't be able to detect from data :/ we need to implement some simple heuristics in openAPI repo that use request and response entities to do that.
@steinitzu I'll ping you there

rudolfix (changes requested):

sources/rest_api/paginators.py
steinitzu: True, we can't detect the query param. But we can detect it from openapi for sure, or configured manually.

sources/rest_api/paginators.py
steinitzu: I'm thinking mostly about consistency, since we use json path in similar places, like the data path, incremental cursor, etc.  
I don't think otherwise having this buried in lists and such is common, but can be useful for re-using pagination config between endpoints when the API is not consistent. E.g. if the key is sometimes top level, sometimes wrapped in an envelope. Jsonpath `"$|*.next"` would work for all endpointds in that case.

burnash (approved):

PR #402, opened 2024-03-12
feat(kafka): add hanging break

IlyaFaer:


sources/kafka/__init__.py
```diff
@@ -80,12 +81,24 @@ def kafka_consumer(
     tracker = OffsetTracker(consumer, topics, dlt.current.resource_state(), start_from)
     consumer.subscribe(topics)
 
+    hanged_at = None
```
rudolfix, line 84: Let's do it differently. We set the default batch timeout to 5 seconds. If the consume goes back empty we exit no matter if we are at the end of stream or not. So we do not need any additional code. That fixed the problem with offsets becoming invalid (and never completing).

the problem with tests we can solve like you do now. with some kind of tenacity retry.

please run tests locally and if all works we merge thus

rudolfix (changes requested):
OK - let's use batch timeout to solve it

sources/kafka/__init__.py
IlyaFaer: > If the consume goes back empty we exit no matter if we are at the end of stream or not

It'll be breaking the read at the very beginning quite often. That's why I was against counting the number of attempts - the tests started to fail much more often.

When you just create a consumer, it takes some time to manage it to a cloud broker. We actually start to read before it happens, and up to the first 40 sec (it was the maximum pause I saw during development), `consume()` may be returning empty results. But then it can suddenly start to read - just because there was no broker to process requests.

Number of attempts in this case would not work fine, so I proposed to use 5 mins as the maximum amount we're ready to tolerate.

IlyaFaer: Let me show you. Here's what I do, I run a test with logging:

image

Here are the results. This one read all the messages from the first attempt:

image

Here it took 3 seconds and 2 attempts, before the messages showed up:

image

Here messages showed up on the first attempt, but not all of them:

image

So, every time it's different, while our code is always doing the same. Today I can't catch a really long delay, like 20 secs, but I'd say it still happens often enough not to consider it as an anomaly.

rudolfix: OK. very good. IMO we should just inform the user that they can expect this behavior if they create new topics: @IlyaFaer let's do the following
- set the batch timeout to 3 seconds (the unit is milliseconds so our current setting does not make sense... but make sure I;m right)
- remove your custom wait logic
- if no messages got returned exit immediately
- do a PR to main repo to update docs where you explain the above: so for new topics you may get a few empty reads before brokers sync

and let's close it, thanks :)

IlyaFaer: Okay, here I did all of it. Making a PR for docs now...

rudolfix (approved):
LGTM!

Issue #330, opened 2024-01-24
Exceeding Maximum Query Length in HubSpot Pipeline


dat-a-man:
**Source name**
Name of the affected source as displayed by `dlt init` or present in `sources` folder of this repo.

**Make sure that**
* [X] I checked the README in the source folder
* [X] If I have problems with the credentials I checked the [credentials docs](https://dlthub.com/docs/general-usage/credentials) and [troubleshooting](https://dlthub.com/docs/walkthroughs/run-a-pipeline#4-detect-and-handle-problems)

**Describe the bug**
Please describe the bug. If possible provide us:
1. hubspot()

**To Reproduce**
I'm encountering an issue when running a pipeline for HubSpot data extraction where the query length exceeds the maximum allowed limit of 2000 symbols. This happens even when specifying a small number of properties.

Steps to Reproduce:
1. Define a list of properties for the pipeline, even with a single item:

```python
DEFAULT_PIPELINES_PROPS_Entered_1 = [
    "hs_date_entered_100452487",
]
```
2. Set up and run the pipeline as follows:
```python
def load_pipelines_1_entered() -> None:
    p = dlt.pipeline(
        pipeline_name="hubspot_pipeline_entered",
        dataset_name="hubspot",
        destination='bigquery',
    )
    load_data = hubspot()
    load_data.deals.bind(props=DEFAULT_PIPELINES_PROPS_Entered_1, include_custom_props=True)
    info = p.run(load_data.with_resources("deals"), table_name="pipelines_entered")
    print(info)
```

**Expected behavior**
The pipeline should execute successfully, extracting the specified properties without exceeding the query length limit.

**Stack traces and other evidence**
The pipeline execution fails with an error indicating that the query length is too long, listing properties that were not explicitly specified:

```
<class 'dlt.extract.exceptions.ResourceExtractionError'> In processing pipe deals: extraction of resource deals in generator deals caused an exception: Your request to Hubspot is too long to process. Maximum allowed query length is 2000 symbols, while your list of properties `age_contact_person,amount,amount_in_home_currency,...` is 2111 symbols long. Use the `props` argument of the resource to set the list of properties to extract from the endpoint.
```

**Running environment**
 - OS: MacOs
 - `dlt` Version [0.4.2]
 - Virtualization / Runner [local]

**Additional context**
This issue occurs even with a minimal list of properties specified.
Include any other relevant details that might help in diagnosing and resolving the issue.

IlyaFaer:
[...]

PR #404, opened 2024-03-14
[enhancement] add include_custom_props to hubspot source

cmpadden:
### Tell us what you do here

- Adds top-level `include_custom_props` parameter to `hubspot` source

### Related Issues

Was experiencing the same issue as #330.

```
<class 'dlt.extract.exceptions.ResourceExtractionError'>
In processing pipe companies: extraction of resource companies in generator companies caused an exception: Your request to Hubspot is too long to process. Maximum allowed query length is 2000 symbols, while your list of properties `about_us,address,address2,admin__c,annualrevenue,arr__c,auto_renew__c,channel_partner__c,cicd_provider__c,city,closedate,closedate_timestamp_earliest_value_a2a17e6e,cloud_credits__c,cloud_provider__c,`... is 2854 symbols long. Use the `props` argument of the resource to set t
```

This introduced a top-level parameter to disable the custom props on all resources.


rudolfix (approved):
LGTM, also tests are passing

PR #408, opened 2024-03-20
fix: add optional type annotation to github source for pyright adherence

cmpadden:
### Tell us what you do here

The following changes were required for the Github source to have proper adherence to our company pyright / ruff policies. Thought I'd push upstream in case others require similar tweaks when bringing this source into their project.

- Adds `Optional` types to nullable function parameters
- Adds period to end of docstring sentences

Loving these verified sources! 🙏


rudolfix (approved):
LGTM! thanks for the improvement!

PR #423, opened 2024-04-11
test(filesystem): add a test for Windows UNC paths

IlyaFaer:
Towards https://github.com/dlt-hub/dlt/pull/1209


IlyaFaer:
@rudolfix, I tried to add UNC paths into the existing tests, but it doesn't work. UNC requires to keep the whole path of the file, without any attempts to make it relative, so it doesn't work with the existing tests.

PR #428, opened 2024-04-18
Source/add google ads

dat-a-man:
1. Created a branch from add_google_ads.
2. Merged master into it
3. Commented out tests


adrianbr:
Needs the readme

rudolfix (changes requested):
@IlyaFaer thx for fixing this!. looks like pytest is exiting with non 0 exit code when there are no tests so maybe we should uncomment this one test we have and skip it. so we have all checks green

rudolfix (approved):
LGTM!

PR #435, opened 2024-04-25
rest_api: use RESTClient class from the dlt core

### LINKED ISSUE #422: Use RESTClient from dlt core in the rest_api source
## Background

Currently rest_api source uses it's own implementation of RESTClient. 

## Task

- [x] Port RESTClient from the dlt core
- [x] Update authenticator metadata attributes 
- [x] Update tests

burnash:
- Resolves #422


sources/rest_api/auth.py
```diff
@@ -1,258 +1,37 @@
...
-            )
-
-    def __call__(self, request: PreparedRequest) -> PreparedRequest:
-        request.headers["Authorization"] = f"Bearer {self.token}"
-        return request
 
 
 @configspec
-class APIKeyAuth(AuthConfigBase):
+class APIKeyAuth(auth.APIKeyAuth):
```
burnash, line 18: @rudolfix is this a proper way to extend the configspec with an additional `type` field?

rudolfix: yes! but why are you subclassing? we just import the Auth implementations from the core... we got rid of the openAPI literals below and we should live with this :)

tldr;> just import core Auth into the `auth` module here

sources/rest_api/config_setup.py
```diff
@@ -19,21 +19,22 @@
...
+from dlt.sources.helpers.rest_client.paginators import (
     BasePaginator,
     HeaderLinkPaginator,
     JSONResponsePaginator,
     SinglePagePaginator,
     JSONResponseCursorPaginator,
 )
+from dlt.sources.helpers.rest_client.exceptions import IgnoreResponseException
+from dlt.sources.helpers.rest_client.auth import AuthConfigBase
+from .auth import (
```
rudolfix, line 31: ok then maybe we do not need auth module at all? we just import form core directly?

rudolfix (approved):
A question: we have the same mock REST API server for testing both in core and here?

sources/rest_api/auth.py
burnash: > yes! but why are you subclassing? we just import the Auth implementations from the core... we got rid of the openAPI literals below and we should live with this :)
I see, I thought we wanted to keep the OpenAPI fields in the verified source versions. Will update.

burnash:
> > A question: we have the same mock REST API server for testing both in core and here?
> 
> @rudolfix yes, not sure how to handle this, I use it in both places.

I think the easiest is to remove the mock rest API from the core. It's not as important there as here, where we test parent-child endpoints

burnash:
@rudolfix, tracking the mocker here: https://github.com/dlt-hub/dlt/issues/1293

Issue #448, opened 2024-04-24
RESTClient: update the template for `dlt init` to use RESTClient


burnash:

sultaniman:
@burnash I will use pokemon api to showcase pagination for `--generic` template

PR #449, opened 2024-05-07
Use RESTClient in generic template

sultaniman:
This PR updates `pipeline_generic` to use `RESTClient` to make requests.

Resolves: https://github.com/dlt-hub/verified-sources/issues/448

### DIFF
--- a/init/pipeline.py
+++ b/init/pipeline.py
@@ -33,6 +33,7 @@ def my_repo_pulls(
     for page in paginate(
         url,
         auth=BearerTokenAuth(api_secret_key),
+        # Note: for more paginators please see: https://dlthub.com/devel/general-usage/http/rest-client#paginators
         paginator=HeaderLinkPaginator(),
     ):
         print(page)
--- a/init/pipeline.py
+++ b/init/pipeline.py
@@ -12,14 +12,18 @@ from dlt.sources.helpers.rest_client.paginators import HeaderLinkPaginator
 
 
 @dlt.source
-def source(api_url: str = dlt.config.value, api_secret_key: str = dlt.secrets.value):
-    return my_repo_pulls(api_url, api_secret_key)
+def source(
+    api_url: str = dlt.config.value,
+    api_secret_key: str = dlt.secrets.value,
+    repository: str = dlt.config.value,
+):
+    return my_repo_pulls(api_url, api_secret_key, repository)
 
 
 @dlt.resource(write_disposition="append")
 def my_repo_pulls(
     api_url: str = dlt.config.value,
-    api_secret_key=dlt.secrets.value,
+    api_secret_key: str = dlt.secrets.value,
     repository: str = dlt.config.value,
 ):
     # repository url should be in the format `OWNER/REPO`
--- a/init/pipeline.py
+++ b/init/pipeline.py
@@ -42,7 +42,7 @@ if __name__ == "__main__":
     # otherwise the defaults will be used that are derived from the current script name
     pipeline = dlt.pipeline(
         pipeline_name="pipeline",
-        destination="bigquery",
+        destination="duckdb",
         dataset_name="pipeline_data",
     )
 
--- a/init/pipeline.py
+++ b/init/pipeline.py
@@ -4,11 +4,9 @@ from dlt.sources.helpers.rest_client import paginate
 from dlt.sources.helpers.rest_client.auth import BearerTokenAuth
 from dlt.sources.helpers.rest_client.paginators import HeaderLinkPaginator
 
-# This pipeline demonstrates how to use the dlt REST client for extracting data from the GitHub API.
-# It showcases the use of authentication via bearer tokens and pagination for navigating through
-# GitHub issues within a repository.
-
-# Note: Ensure the API key (Bearer token) is set up correctly in secrets or environment variables.
+# This is a generic pipeline example and demonstrates
+# how to use the dlt REST client for extracting data from APIs.
+# It showcases the use of authentication via bearer tokens and pagination.
 
 
 @dlt.source
@@ -17,23 +15,22 @@ def source(
     api_secret_key: str = dlt.secrets.value,
     repository: str = dlt.config.value,
 ):
-    return my_repo_issues(api_url, api_secret_key, repository)
+    return resource(api_url, api_secret_key, repository)
 
 
 @dlt.resource(write_disposition="append")
-def my_repo_issues(
-    api_url: str = dlt.config.value,
+def resource(
     api_secret_key: str = dlt.secrets.value,
-    repository: str = dlt.config.value,
+    org: str = "dlt-hub",
+    repository: str = "dlt",
 ):
-    # repository url should be in the format `OWNER/REPO`
-
     # paginate issues and yield every page
-    url = f"{api_url}/repos/{repository}/issues"
+    api_url = f"https://api.github.com/repos/{org}/{repository}/issues"
     for page in paginate(
-        url,
+        api_url,
         auth=BearerTokenAuth(api_secret_key),
-        # Note: for more paginators please see: https://dlthub.com/devel/general-usage/http/rest-client#paginators
+        # Note: for more paginators please see:
+        # https://dlthub.com/devel/general-usage/http/rest-client#paginators
         paginator=HeaderLinkPaginator(),
     ):
         print(page)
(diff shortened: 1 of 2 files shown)


burnash (changes requested):
@sultaniman please also update https://github.com/dlt-hub/verified-sources/blob/master/init/pipeline.py
As discussed, let's use GitHub API as a placeholder since it has authentication.
Consider also a `paginate()` shortcut from rest_client module: https://dlthub.com/devel/general-usage/http/rest-client#shortcut-for-paginating-api-responses

sultaniman:
@burnash thanks for updates I will adapt these to use GitHub and will explicitly set up paginator ad auth scheme

burnash (changes requested):
Great update. Since it's only a single endpoint I suggest refining it even further and using paginate shortcut:

```py
from dlt.sources.helpers.rest_client import paginate

for page in paginate("https://api.example.com/posts"):
    print(page)
```

init/pipeline.py
```diff
@@ -1,43 +1,54 @@
...
-@dlt.source
-def source(api_secret_key=dlt.secrets.value):
-    return resource(api_secret_key)
-
-
-def _create_auth_headers(api_secret_key):
-    """Constructs Bearer type authorization header which is the most common authorization method"""
-    headers = {"Authorization": f"Bearer {api_secret_key}"}
-    return headers
+# This pipeline demonstrates how to build a simple REST client for interacting with GitHub's API.
```
burnash, line 7:
```suggestion
# This pipeline demonstrates how to use the dlt REST client for extracting data from the GitHub API.
```

init/pipeline_generic.py
```diff
@@ -1,76 +1,90 @@
+from typing import Any, Dict, Optional
+
 import dlt
-from dlt.sources.helpers import requests
 
+from dlt.sources.helpers.rest_client import paginate
+from dlt.sources.helpers.rest_client.auth import BearerTokenAuth
+from dlt.sources.helpers.rest_client.paginators import HeaderLinkPaginator
+
+# This pipeline demonstrates how to build a simple REST client for interacting with GitHub's API.
```
burnash, line 9: Need to update this as in the pipeline.py file

init/pipeline.py
```diff
@@ -1,43 +1,54 @@
...
+    repository: str = dlt.config.value,
+):
+    # repository url should be in the format `OWNER/REPO`
+
+    # paginate pull requests and yield every page
+    url = f"{api_url}/repos/{repository}/pulls"
+    for page in paginate(
+        url,
+        auth=BearerTokenAuth(api_secret_key),
+        paginator=HeaderLinkPaginator(),
```
burnash, line 36: Let's add a comment with a not that other paginators are available for different strategies and link https://dlthub.com/devel/general-usage/http/rest-client#paginators

burnash (changes requested):
Looks good. See my comments with minor changes to make it even better.

init/pipeline_generic.py
sultaniman: @burnash what do you exactly mean?

init/pipeline.py
```diff
@@ -1,43 +1,55 @@
...
-@dlt.source
-def source(api_secret_key=dlt.secrets.value):
-    return resource(api_secret_key)
-
-
-def _create_auth_headers(api_secret_key):
-    """Constructs Bearer type authorization header which is the most common authorization method"""
-    headers = {"Authorization": f"Bearer {api_secret_key}"}
-    return headers
+# This pipeline demonstrates how to use the dlt REST client for extracting data from the GitHub API.
```
rudolfix, line 7: think of those templates as generic. they are not examples. it is OK if we use Github just move this comment into the dlt.resource

init/pipeline.py
```diff
@@ -1,43 +1,55 @@
...
-    # make an api call here
-    # response = requests.get(url, headers=headers, params=params)
-    # response.raise_for_status()
-    # yield response.json()
 
-    # test data for loading validation, delete it once you yield actual data
-    test_data = [{"id": 0}, {"id": 1}]
-    yield test_data
+@dlt.resource(write_disposition="append")
+def my_repo_issues(
```
rudolfix, line 24: IMO name it `resource` - engine will rename it anyway. again: this is not an example

init/pipeline.py
```diff
@@ -1,43 +1,55 @@
...
+    api_url: str = dlt.config.value,
+    api_secret_key: str = dlt.secrets.value,
+    repository: str = dlt.config.value,
+):
+    # repository url should be in the format `OWNER/REPO`
```
rudolfix, line 29: like in the previous example
1. print out the credentials
2. yield test data ie. single issue with just a few attributes
3. comment out the actual code
why? because we want people to run this thing right away

init/pipeline_generic.py
```diff
@@ -1,76 +1,90 @@
...
+# Note: Ensure the API key (Bearer token) is set up correctly in secrets or environment variables.
 
 
 @dlt.source
 def source(
-    explicit_arg,
     api_url=dlt.config.value,
     api_secret_key=dlt.secrets.value,
-    default_arg="default",
+    issues_params: Optional[Dict[str, Any]] = None,
```
rudolfix, line 20: instead of those pass org and repo as default params. ie `dlt-hub` and `dlt`

init/pipeline.py
```diff
@@ -1,43 +1,55 @@
...
-def resource(api_secret_key=dlt.secrets.value):
-    headers = _create_auth_headers(api_secret_key)
 
-    # check if authentication headers look fine
-    print(headers)
+@dlt.source
+def source(
+    api_url: str = dlt.config.value,
+    api_secret_key: str = dlt.secrets.value,
+    repository: str = dlt.config.value,
```
rudolfix, line 18: remove it. hardcode dlt repo in the resource, make it super simple

init/pipeline_generic.py
rudolfix: again: this is generic template. it creates a source with a default parameters and two resources, you can explain that and move github related things into resources

init/pipeline_generic.py
```diff
@@ -1,76 +1,90 @@
...
+    return [
+        my_repo_issues(api_url, api_secret_key, params=issues_params),
+        my_repo_pulls(api_url, api_secret_key, params=pulls_params),
+    ]
 
 
 @dlt.resource
-def resource_1(
-    explicit_arg, api_url=dlt.config.value, api_secret_key=dlt.secrets.value
+def my_repo_issues(
```
rudolfix, line 33: I'd keep resource_1 and resource_2

init/pipeline_generic.py
```diff
@@ -1,76 +1,90 @@
...
+    api_url: str = dlt.config.value,
+    api_secret_key: str = dlt.secrets.value,
+    repository: str = dlt.config.value,
+    params: Optional[Dict[str, Any]] = None,
 ):
-    headers = _create_auth_headers(api_secret_key)
-
-    # uncomment line below to see if your headers are correct (ie. include valid api_key)
```
rudolfix, line 30: keep the same pattern, leave the debug prints commented out

rudolfix (changes requested):

init/pipeline.py
sultaniman: @rudolfix in `#3` did you mean to just copy-paste a sample of a single issue and comment out the real code which paginates?

sultaniman: I updated the code.

init/pipeline_generic.py
```diff
@@ -1,76 +1,69 @@
...
+        resource_2(api_url, api_secret_key),
+    ]
 
 
 @dlt.resource
-def resource_2(
-    api_url=dlt.config.value, api_secret_key=dlt.secrets.value, default_arg="default"
-):
-    headers = _create_auth_headers(api_secret_key)
+def resource_1(api_url: str, api_secret_key: str = dlt.secrets.value):
```
sultaniman, line 31: @rudolfix should we generate resources instead or should we keep it this way?

init/pipeline.py
```diff
@@ -1,50 +1,75 @@
...
-    # make an api call here
-    # response = requests.get(url, headers=headers, params=params)
-    # response.raise_for_status()
-    # yield response.json()
+def resource(
+    api_secret_key: str = dlt.secrets.value,
+    org: str = "dlt-hub",
+    repository: str = "dlt",
+):
+    yield [
```
burnash, line 24: If we actually yielding data, as an in the previous file version we should keep the comment that explains what's going on. 
e.g "test data for loading validation, delete it once you yield actual data"

burnash (changes requested):

init/pipeline.py
sultaniman: That's a good point I added it.

burnash (approved):
Looks good!

rudolfix (approved):
LGTM!

PR #451, opened 2024-05-08
kafka: don't subscribe after consumer has been assigned

### LINKED ISSUE #450: Kafka source should either subscribe or assign, not both?
### dlt version

0.4.7

### Source name

kafka

### Describe the problem

> [!WARNING]
> Kafka is still a new concept for me, so bear with me if I'm using incorrect terminology or have misunderstood some concepts.

When using the Kafka source, if for any reason the extraction is interupted, messages consumed in the initial run are not reprocessed and are lost forever.

My understanding is that this happens because the assignment made in the offset tracker is undone by the `Conumser.subscribe` call immeaditely after:

> **sources/kafka/helpers.py**
> https://github.com/dlt-hub/verified-sources/blob/9d7c77e927ddec051467ebd68eaf261c3a3c6465/sources/kafka/helpers.py#L159

> **sources/kafka/__init__.py**
> https://github.com/dlt-hub/verified-sources/blob/9d7c77e927ddec051467ebd68eaf261c3a3c6465/sources/kafka/__init__.py#L81-L82

Commenting out the line in question (Line 82 `consumer.subscribe(topics)`) seems to resolve the issue.

### Expected behavior

The source tracks it's own state and should not rely on offsets provided by the broker?

### Steps to reproduce

Run a Kafka pipeline and abort during extraction. Restart the pipeline and watch all the precious messages be gone forever.

### Operating system

Linux

### Runtime environment

Kubernetes

### Python version

3.12

### dlt destination

bigquery

### Additional information

_No response_

aksestok:
### Tell us what you do here

### Short description

See #450.

Don't subscribe after assigning partitions and offsets.

### Related Issues

- Fixes #450

### DIFF
--- a/sources/kafka/__init__.py
+++ b/sources/kafka/__init__.py
@@ -79,7 +79,6 @@ def kafka_consumer(
         start_from = ensure_pendulum_datetime(start_from)
 
     tracker = OffsetTracker(consumer, topics, dlt.current.resource_state(), start_from)
-    consumer.subscribe(topics)
 
     # read messages up to the maximum offsets,
     # not waiting for new messages


AstrakhantsevaAA:
@IlyaFaer take a look please

IlyaFaer:
After every message consumed, offsets are moved further (both by Kafka and by `dlt` in pipeline's state). If a load failed, we are supposed to return offsets to the start of the failed load, which doesn't happen. This is why messages are lost.

So, I don't think it'll really solve the problem. `assign()` will still set offsets, taken from the pipeline state. Even if we take away `assign()`, and allow Kafka to track offset by itself, messages will still be lost, as Kafka also moves offsets after every consumed message (we basically duplicate this mechanism in `dlt` to add some high-level functionality like reading from a timestamp). We need to save offsets in the pipeline state only in case a load was successful - that's what supposed to solve the problem.

Taking a closer look at it...

IlyaFaer:
@aksestok, @rudolfix, after all, you may be right. For all cases, Kafka returns offset `-1001`. That's a nasty constant, which is supposed to be an invalid offset.

Though it's marked invalid, Kafka actually accepts our offsets. But! In case `subscribe()` called, Kafka reverts offsets we just assigned. And even if I explicitly commit offsets, `subscribe()` still reverts to the previous offset (though `-1001` is supposed to use the last committed offset, and we just committed it).

Seems to me the problem is in these `-1001` (which comes from backend - in code we have normal digits). All the tests are passing without `subscribe()` (but they do not pass anymore with `subscribe()` 🫤 ), so I think we can merge it.

Untitled

IlyaFaer:
The craziest is that before and after `subscribe()`, Kafka has the same assignments and positions. Though they are not.

image

IlyaFaer (approved):

rudolfix:
@IlyaFaer this also shows that we miss an essential test. could you extract the messages, `drop` the pipeline (to simulate state rollback) and run it again making sure the same messages are back?

PR #462, opened 2024-05-14
rest_api: add comments to example pipeline

burnash:


adrianbr (approved):

sources/rest_api_pipeline.py
```diff
@@ -28,19 +24,31 @@ def load_github() -> None:
...
             {
                 "name": "issues",
                 "endpoint": {
                     "path": "issues",
+                    # Query parameters for the endpoint
                     "params": {
                         "sort": "updated",
                         "direction": "desc",
                         "state": "open",
+                        # Define `since` as a special parameter
```
sultaniman, line 47: Maybe add a not that the name can be any valid name not only `since` or is it not the case?

sultaniman (approved):

sources/rest_api_pipeline.py
burnash: It'a s `since` query string parameter from the API, probably need to clarify that

PR #466, opened 2024-05-15
Pipedrive: fix tests

AstrakhantsevaAA:
we have no data in our Pipedrive test account (only "leads"), skip some tests for now


rudolfix (approved):
LGTM!
no activity because pipedrive deletes it after 60 days or something

PR #468, opened 2024-05-16
sql_database: Add flag to include views with reflected tables

### LINKED ISSUE #459: Enable default inclusion of views in MetaData object (SQL database)
### Feature description

Please enable the `MetaData.reflect` function in dlt to include views by default when accessing database resources. Currently, views are not included by default, requiring additional steps to configure, leading to potential oversight.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

When extracting and loading views from a SQL database, users need to manually enable view inclusion by setting `MetaData.reflect(bind=engine, views=True)` method. This additional step can be overlooked and is not initially intuitive.

### Proposed solution

Modify the default behavior of the MetaData object to include views by default by setting views=True in the MetaData.reflect method. This would streamline processes for users who manage databases and want to load views. 

### Related issues

_No response_

steinitzu:
### Tell us what you do here

- Add flag `include_views` to `sql_database` source to enable reflecting views
- Test for the flag
- Test for `sql_table` works when the table name is a view
- Test for `sql_database` always includes views when they're passed in `table_names` regardless of the `include_views` flag

### Related Issues
- Resolves #459

### DIFF
--- a/sources/sql_database/__init__.py
+++ b/sources/sql_database/__init__.py
@@ -34,6 +34,7 @@ def sql_database(
     defer_table_reflect: Optional[bool] = dlt.config.value,
     table_adapter_callback: Callable[[Table], None] = None,
     backend_kwargs: Dict[str, Any] = None,
+    include_views: bool = False,
 ) -> Iterable[DltResource]:
     """
     A dlt source which loads data from an SQL database using SQLAlchemy.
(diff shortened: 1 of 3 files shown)


rudolfix (changes requested):
@steini so code looks good by why counts do not match in the tests? any duplicated keys that multiply rows on the JOINs

steinitzu:
> @steini so code looks good by why counts do not match in the tests? any duplicated keys that multiply rows on the JOINs

Ah I didn't run all the tests locally in sequence so I missed this. The test db is session scoped and there's another incremental test that runs before and inserts more rows but didn't update the expected row counts for the view.

Should run now.

rudolfix (approved):

PR #471, opened 2024-05-17
reset_api: allow JSONPath in field value in ResolveParamConfig

burnash:
This pull request adds ability to use JSONPath in [path parameter resolve configuration](https://dlthub.com/docs/dlt-ecosystem/verified-sources/rest_api#define-resource-relationships):

Example:

Suppose you have parent resource data item like:

```
{
    "number": 42,
    "nested_data": {
       "some_nested_identifier": 24,
       ...
    }
}
```

Before this PR you could only refer the `number` field, like so: 

```
{
    "resources": [
        ...
        {
            "name": "issue_comments",
            "endpoint": {
                "path": "issues/{issue_number}/comments",
                "params": {
                    "issue_number": {
                        "type": "resolve",
                        "resource": "issues",
                        "field": "number",
                    }
                },
            },
        },
    ],
}
```

With this PR you can also refer `nested_data.some_nested_identifier` field by using JSONPath:

```
{
    "resources": [
        ...
        {
            "name": "issue_comments",
            "endpoint": {
                "path": "issues/{issue_number}/comments",
                "params": {
                    "issue_number": {
                        "type": "resolve",
                        "resource": "issues",
                        "field": "nested_data.some_nested_identifier",
                    }
                },
            },
        },
    ],
}
```

### DIFF
--- a/sources/rest_api/config_setup.py
+++ b/sources/rest_api/config_setup.py
@@ -19,6 +19,7 @@ from dlt.common import logger
 from dlt.common.configuration import resolve_configuration
 from dlt.common.schema.utils import merge_columns
 from dlt.common.utils import update_dict_nested
+from dlt.common import jsonpath
 
 from dlt.extract.incremental import Incremental
 from dlt.extract.utils import ensure_table_schema_columns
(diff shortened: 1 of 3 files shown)


rudolfix (approved):
@bursnash IMO would be cool to accept both compiled path and a string in "resolve" param type in RESTAPIConfig. we accept it ie. in paginators
your choice though.

burnash:
Thanks @rudolfix, I think passing compiled jsonpath should be covered already per se, it's just the typing needs to be updated, isn't it?

Issue #438, opened 2024-04-26
mongodb: support end_date, row_order, limit and arrow backend


rudolfix:
### Source name

mongo

### Describe the data you'd like to see

I'd like a `mongodb` features on par with `sql_database` features:
1. backfill with incremental (start / end ranges) - we need to support end_value
2. explicit ordering of results (support `row_order`
3. allow to set a `limit` on a source/resource level to load just N first results. will help with microbatching
4. `pymongo` got arrow support: https://github.com/mongodb-labs/mongo-arrow and we should use it

### Are you a dlt user?

Yes, I'm already a dlt user.

### Do you ready to contribute this extension?

Yes, I'm ready.

### dlt destination

any

### Additional information

ad `limit`: limit works well with ordered results. if results are not ordered (lack of incremental field and declared order) - issue a warning!

Please split ticket into two PRs
1. requirements 1-3
2. pyarrow backend - a separate one

PR #474, opened 2024-05-22
feat(mongodb): support start, end values and sorting

IlyaFaer:
Towards #438 

Implements items 1 and 2

### DIFF
--- a/sources/mongodb/helpers.py
+++ b/sources/mongodb/helpers.py
@@ -37,40 +37,62 @@ class CollectionLoader:
         self.client = client
         self.collection = collection
         self.incremental = incremental
+
         if incremental:
             self.cursor_field = incremental.cursor_path
-            self.last_value = incremental.last_value
         else:
-            self.cursor_column = None
-            self.last_value = None
+            self.cursor_field = None
 
     @property
     def _filter_op(self) -> Dict[str, Any]:
-        if not self.incremental or not self.last_value:
+        """Build a filter operator.
+
+        Describes what logical statement to apply to which field.
+
+        Returns:
+            Dict[str, Any]: The filter operator.
+        """
+        if not self.incremental:
             return {}
-        if self.incremental.last_value_func is max:
-            return {self.cursor_field: {"$gte": self.last_value}}
-        elif self.incremental.last_value_func is min:
-            return {self.cursor_field: {"$lt": self.last_value}}
+
+        if self.incremental.start_value:
+            return {self.cursor_field: {"$gte": self.incremental.start_value}}
+
+        if self.incremental.end_value:
+            return {self.cursor_field: {"$lte": self.incremental.end_value}}
+
         return {}
+
+    @property
+    def _sort_op(self) -> List[Optional[Tuple[str, int]]]:
+        """Build a sorting operator.
+
+        Describes what sorting to apply to which field.
+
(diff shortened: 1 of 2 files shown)


sources/mongodb/helpers.py
```diff
@@ -37,40 +37,62 @@ def __init__(
         self.client = client
         self.collection = collection
         self.incremental = incremental
+
         if incremental:
             self.cursor_field = incremental.cursor_path
-            self.last_value = incremental.last_value
         else:
-            self.cursor_column = None
```
IlyaFaer, line 44: That seems like a mistake: `cursor_field` and `cursor_column`, while the second one is never used

sources/mongodb/helpers.py
```diff
@@ -37,40 +37,62 @@ def __init__(
...
-            yield map_nested_in_place(convert_mongo_objs, docs_slice)
+        if self.incremental.start_value:
+            return {self.cursor_field: {"$gte": self.incremental.start_value}}
 
+        if self.incremental.end_value:
+            return {self.cursor_field: {"$lte": self.incremental.end_value}}
+
+        return {}
 
-class CollectionLoaderParallell(CollectionLoader):
```
IlyaFaer, line 63: Strange that `_sort_op()` was implemented in `CollectionLoaderParallell`, but no in its ancestor `CollectionLoader`. I moved it upper.

sources/mongodb/helpers.py
```diff
@@ -37,40 +37,62 @@ def __init__(
...
-            return {self.cursor_field: {"$gte": self.last_value}}
-        elif self.incremental.last_value_func is min:
-            return {self.cursor_field: {"$lt": self.last_value}}
-        return {}
 
-    def load_documents(self) -> Iterator[TDataItem]:
-        cursor = self.collection.find(self._filter_op)
-        while docs_slice := list(islice(cursor, CHUNK_SIZE)):
-            yield map_nested_in_place(convert_mongo_objs, docs_slice)
+        if self.incremental.start_value:
```
rudolfix, line 58: are tests really passing? because you removed `min` and `max` functions which reverse the range criteria! so if tests are passing please add them

sources/mongodb/helpers.py
```diff
@@ -37,40 +37,62 @@ def __init__(
...
+        Describes what sorting to apply to which field.
+
+        Returns:
+            List[Optional[Tuple[str, int]]]: The sorting operator.
+        """
+        if not (self.incremental and self.incremental.row_order):
             return []
-        if self.incremental.last_value_func is max:
+
+        if self.incremental.row_order == "asc":
```
rudolfix, line 78: same here, min and max should reverse the order. `max` is the normal order btw. (ascending)

rudolfix: we also need test for `min` and `max`. also check if incremental loading works

rudolfix (changes requested):
A few smaller things
1. rename Parallell -> Parallel
2. we need more tests
- incremental loading: use start - end to load a range and then start value without end to read remaining elements. subsequent load should return no elements because we are at the end of range
- make sure it works exactly the same with all possible sort ops (it should
- swap last value func (max -> min) and write the same test
3. please make CHUNK SIZE configurable

PR #475, opened 2024-05-24
feat(mongodb): support incremental end_value

IlyaFaer:
Towards #438

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ types-psycopg2 = "^2.9.0"
 pytest-mock = "^3.12.0"
 twisted = "22.10.0"
 pytest-forked = "^1.6.0"
+pendulum = "^3.0.0"
 
 [tool.poetry.group.sql_database.dependencies]
 sqlalchemy = ">=1.4"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ packages = [{include = "sources"}]
 python = ">=3.8.1,<3.13"
 dlt = {version = "0.4.11", allow-prereleases = true, extras = ["redshift", "bigquery", "postgres", "duckdb"]}
 graphlib-backport = {version = "*", python = "<3.9"}
+pendulum = ">=3.0"
 
 [tool.poetry.group.dev.dependencies]
 mypy = "^1.10.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,6 @@ packages = [{include = "sources"}]
 python = ">=3.8.1,<3.13"
 dlt = {version = "0.4.11", allow-prereleases = true, extras = ["redshift", "bigquery", "postgres", "duckdb"]}
 graphlib-backport = {version = "*", python = "<3.9"}
-pendulum = ">=3.0"
 
 [tool.poetry.group.dev.dependencies]
 mypy = "^1.10.0"
--- a/sources/mongodb/__init__.py
+++ b/sources/mongodb/__init__.py
@@ -84,7 +84,7 @@ def mongodb_collection(
             E.g., `incremental=dlt.sources.incremental('updated_at', pendulum.parse('2022-01-01T00:00:00Z'))`
         write_disposition (str): Write disposition of the resource.
         parallel (Optional[bool]): Option to enable parallel loading for the collection. Default is False.
-        limit (Optional[int]): Limit the number of documents to be loaded.
+        limit (Optional[int]): The number of documents to be loaded.
 
     Returns:
         Iterable[DltResource]: A list of DLT resources for each collection to be loaded.
--- a/sources/mongodb/__init__.py
+++ b/sources/mongodb/__init__.py
@@ -72,6 +72,7 @@ def mongodb_collection(
     write_disposition: Optional[str] = dlt.config.value,
     parallel: Optional[bool] = False,
     limit: Optional[int] = None,
+    chunk_size: Optional[int] = 10000,
 ) -> Any:
     """
     A DLT source which loads a collection from a mongo database using PyMongo.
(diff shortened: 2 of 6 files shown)


sources/mongodb/helpers.py
```diff
@@ -45,32 +45,53 @@ def __init__(
...
     @property
-    def _filter_op(self) -> Dict[str, Any]:
+    def _sort_op(self) -> List[Optional[Tuple[str, int]]]:
         if not self.incremental or not self.last_value:
+            return []
+        if self.incremental.last_value_func is max:
+            return [(self.cursor_field, ASCENDING)]
+        elif self.incremental.last_value_func is min:
+            return [(self.cursor_field, DESCENDING)]
+        return []
```
IlyaFaer, line 55: The `_sort_op()` method moved without changes into the parent class to make sorting work in both parallel and ordinary cases

sources/mongodb/helpers.py
```diff
@@ -45,32 +45,53 @@ def __init__(
...
+
+            return filt
+
         elif self.incremental.last_value_func is min:
-            return {self.cursor_field: {"$lt": self.last_value}}
+            filt = {self.cursor_field: {"$lte": self.last_value}}
+            if self.incremental.end_value:
+                filt[self.cursor_field]["$gt"] = self.incremental.end_value
+
+            return filt
```
IlyaFaer, line 81: Added `end_value` into filter

IlyaFaer:
@rudolfix, the last `pipeline.run()` call (the one that is supposed to return no data) fails with an error deep inside `pendulum`. Upgrading the version solves the problem.

sources/mongodb/helpers.py
rudolfix: take both row order and last value func. look here for code and tests (sql database didn't do that correctly as well): https://github.com/dlt-hub/verified-sources/pull/476

this is due to python code that applies both `row_order` and `last_value_func` to find out if items are out of range

rudolfix (changes requested):

sources/mongodb/helpers.py
```diff
@@ -45,32 +45,53 @@ def __init__(
...
+        if not (self.incremental and self.last_value):
             return {}
+
         if self.incremental.last_value_func is max:
-            return {self.cursor_field: {"$gte": self.last_value}}
+            filt = {self.cursor_field: {"$gte": self.last_value}}
+            if self.incremental.end_value:
+                filt[self.cursor_field]["$lt"] = self.incremental.end_value
+
+            return filt
```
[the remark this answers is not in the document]
IlyaFaer: Done

pyproject.toml
```diff
@@ -14,6 +14,7 @@ packages = [{include = "sources"}]
 python = ">=3.8.1,<3.13"
 dlt = {version = "0.4.11", allow-prereleases = true, extras = ["redshift", "bigquery", "postgres", "duckdb"]}
 graphlib-backport = {version = "*", python = "<3.9"}
+pendulum = ">=3.0"
```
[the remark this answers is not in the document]
IlyaFaer: Done

sources/mongodb/__init__.py
```diff
@@ -71,6 +71,7 @@ def mongodb_collection(
     incremental: Optional[dlt.sources.incremental] = None,  # type: ignore[type-arg]
     write_disposition: Optional[str] = dlt.config.value,
     parallel: Optional[bool] = False,
+    limit: Optional[int] = None,
```
IlyaFaer, line 74: Only adding into the resource, because `limit` in the source seems very questionable - there will be several collections, how to apply limit to them? To every collection? Or to all the collections together?

sources/mongodb/helpers.py
```diff
@@ -45,32 +46,66 @@ def __init__(
...
+            limit (Optional[int]): The number of documents to load.
+
+        Yields:
+            Iterator[TDataItem]: An iterator of the loaded documents.
+        """
         cursor = self.collection.find(self._filter_op)
+        if self._sort_op:
+            cursor = cursor.sort(self._sort_op)
+
+        if limit != 0:
```
IlyaFaer, line 96: MongoDB allows negative limits. But if `0` is given, it'll use the default request size.

sources/mongodb/helpers.py
```diff
@@ -71,19 +56,93 @@ def _sort_op(self) -> List[Optional[Tuple[str, int]]]:
...
+        """Apply a limit to the cursor, if needed.
+
+        Args:
+            cursor (Cursor): The cursor to apply the limit.
+            limit (Optional[int]): The number of documents to load.
+
+        Returns:
+            Cursor: The cursor with the limit applied (if given).
+        """
+        if (limit or 0) != 0:
```
IlyaFaer, line 94: MongoDB allows negative limits. But if 0 is given, it'll use the default request size = no explicit limit.

sources/mongodb/helpers.py
```diff
@@ -71,19 +56,93 @@ def _sort_op(self) -> List[Optional[Tuple[str, int]]]:
...
-        ]
+        if limit:
+            doc_count = min(doc_count, limit)
+
+        batches = []
+        left_to_load = doc_count
+
+        for sk in range(0, doc_count, self.chunk_size):
+            batches.append(dict(skip=sk, limit=min(self.chunk_size, left_to_load)))
+            left_to_load -= self.chunk_size
```
IlyaFaer, line 137: As there may be situations when `limit % chink_size !=0`, we need to count how many records we're about to return, not to return more than `limit`

tests/mongodb/test_mongodb_source.py
```diff
@@ -64,7 +65,135 @@ def test_nested_documents():
...
+        collection="comments",
+        limit=15,
+        parallel=True,
+        chunk_size=2,
+        incremental=dlt.sources.incremental("date"),
+    )
+    pipeline.run(comments)
+
+    table_counts = load_table_counts(pipeline, "comments")
+    assert table_counts["comments"] == 15
```
IlyaFaer, line 199: This test checks `limit % chunk_size != 0` case

rudolfix (changes requested):
let's add limit to the source as well!

otherwise: LGTM

I assume `row_order` comes separately

rudolfix (approved):

Issue #484, opened 2024-05-08
Type is not JSON serializable: Timestamp


kiwamizamurai:
### dlt version

0.4.10

### Describe the problem

I cannot do elt from mongodb atlas to bigquery, especially ingestion from mongo

Actually, this problem resides within the dependent package `orjson`
and there is an opening issue. https://github.com/ijl/orjson/issues/442

Error message

```bash
<class 'TypeError'>
Type is not JSON serializable: Timestamp
```

```bash
 {'allUsers': [{'db': 'admin', 'user': 'hogehoge'}],
  'appName': 'MongoDB Compass',
  'client': 'xxxxxxxxxxx',
  'command': {'$clusterTime': {'clusterTime': Timestamp(1686813362, 110),
                               'signature': {'hash': b'xxxxxxxx',
                                             'keyId': xxxxxxxxxxxx}},
              '$db': 'test',
              'aggregate': 'fugafuga',
              'cursor': {},
```

```bash
TypeError: Timestamp(1686813362, 110) is not JSON serializable
```

https://mongodb.github.io/node-mongodb-native/4.0/interfaces/clustertime.html

### Expected behavior

no error occur and the elt finishes successfully

### Steps to reproduce

working on it

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

mongo

### dlt destination

Google BigQuery

### Other deployment details

Referred this article
- https://www.holistics.io/blog/modeling-unstructured-data-mongodb/

rudolfix:
@IlyaFaer my take is that we are not translating several basic bson types to Python types. could you take a look?

rudolfix:
what we need is 
1. a test where we put all possible datatypes in mongo into a collection and try to retrieve
2. see which types you can actually decode to python dict/list item format
3. see which types we can decode into arrow
4. and then fix our decoder and possibly type casting for arrow

rudolfix:
@kiwamizamurai this got fixed and merged in #486

kiwamizamurai:
Thank you

PR #486, opened 2024-06-05
feat(mongodb): support mongodb through Arrow

IlyaFaer:
Towards #438, #484


sources/mongodb_arrow/helpers.py
```diff
@@ -0,0 +1,105 @@
...
+    table = find_arrow_all(collection, query)
+    for doc in table.to_batches(1):
+        doc = doc.to_pydict()
+        for field in doc:
+            if field in ids:
+                doc[field] = str(doc[field][0])
+            else:
+                doc[field] = doc[field][0]
+
+        yield doc
```
IlyaFaer, line 105: Some manual type casting is required here, as passing a schema into `find_arrow_all` as well as casting the result table columns don't work (in the 1st case all the ids are becoming `null`, in the second - there are casting failures deep inside `pyarrow`).

sources/mongodb_arrow_pipeline.py
```diff
@@ -0,0 +1,46 @@
...
+        collection="comments",
+        incremental=dlt.sources.incremental(
+            "date",
+            initial_value=pendulum.DateTime(2005, 1, 1, tzinfo=timezone("UTC")),
+            end_value=pendulum.DateTime(2005, 6, 1, tzinfo=timezone("UTC")),
+        ),
+    )
+
+    info = pipeline.run(comments)
+    return info
```
IlyaFaer, line 42: This pipeline works fine, I tested it locally, so principally the source is working

sources/mongodb/helpers.py
```diff
@@ -217,11 +258,19 @@ def collection_documents(
...
     Returns:
         Iterable[DltResource]: A list of DLT resources for each collection to be loaded.
     """
-    LoaderClass = CollectionLoaderParallel if parallel else CollectionLoader
+    if data_processor == "arrow":
+        LoaderClass = CollectionArrowLoader
+    elif parallel:
+        LoaderClass = CollectionLoaderParallel
+    else:
+        LoaderClass = CollectionLoader
```
IlyaFaer, line 273: Looking if Arrow will work with parallels...

sources/mongodb/helpers.py
```diff
@@ -198,13 +202,50 @@ def load_documents(self, limit: Optional[int] = None) -> Iterator[TDataItem]:
...
+            cursor = cursor.sort(self._sort_op)  # type: ignore
+
+        cursor = self._limit(cursor, limit)  # type: ignore
+
+        for batch in cursor:
+            process_bson_stream(batch, context)
+
+            table = context.finish().to_pylist()
+
+            yield map_nested_in_place(convert_mongo_objs, table)
```
IlyaFaer, line 238: So far, I tried to use schema for the `find_arrow_all()`, schema in the context and a single column type cast - none work. Whether it fails with an error or turns all the values into `null` (the column type changes though)

sources/mongodb/helpers.py
```diff
@@ -198,13 +205,89 @@ def load_documents(self, limit: Optional[int] = None) -> Iterator[TDataItem]:
...
+        context = PyMongoArrowContext.from_schema(
+            None, codec_options=self.collection.codec_options
+        )
+
+        for chunk in cursor.skip(batch["skip"]).limit(batch["limit"]):
+            process_bson_stream(chunk, context)
+
+            table = context.finish().to_pylist()
+
+            yield map_nested_in_place(convert_mongo_objs, table)
```
rudolfix, line 280: why this is yielding objects while the regular arrow yields tables? we need arrow version of `convert_mongo_objs` (which you started writing above for the object id) and you should use it in both places

sources/mongodb/helpers.py
```diff
@@ -12,10 +13,16 @@
 from dlt.common.typing import TDataItem
 from dlt.common.utils import map_nested_in_place
 from pendulum import _datetime
+import pyarrow
 from pymongo import ASCENDING, DESCENDING, MongoClient
 from pymongo.collection import Collection
 from pymongo.cursor import Cursor
 
+import pymongoarrow
```
rudolfix, line 21: pyarrow and pymongoarrow should be imported in ` CollectionArrowLoader` only when they are used.
please import arrow from `dlt.common.libs.pyarrow`. this handles missing dependencies nicely

rudolfix (commented):
OK! code structure is way better.
1. parallel arrow looks unfinished
2. you can update README and describe new functionality
3. I'll try to somehow convert this objectid to string in arrow

sources/mongodb/helpers.py
```diff
@@ -217,11 +289,23 @@ def collection_documents(
         parallel (bool): Option to enable parallel loading for the collection. Default is False.
         limit (Optional[int]): The maximum number of documents to load.
         chunk_size (Optional[int]): The number of documents to load in each batch.
+        data_processor (Literal["arrow"]): The data processor to use for loading.
```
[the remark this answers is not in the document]
IlyaFaer: Done

sources/mongodb/helpers.py
IlyaFaer: Done

sources/mongodb/helpers.py
IlyaFaer: Yeah, it's because I pushed changes to show you an approach. It actually wasn't finished.

IlyaFaer:
@rudolfix, seems to me a small change in the core is needed. There is an error happening:

```python
dlt.extract.exceptions.InvalidStepFunctionArguments: In processing pipe comments: Unable to call IncrementalResourceWrapper: Array type doesn't match type of values set: string vs null. The mapping/filtering function IncrementalResourceWrapper requires first argument to take data item and optional second argument named 'meta', but the signature is (item: Union[Any, List[Any]], meta: Any = None) -> Union[Any, List[Any], NoneType]
```

I figured out that such a change fixes it:

image

So, if there is nothing in `remove_ids`, then we don't even do filtering.

sources/mongodb/helpers.py
```diff
@@ -238,6 +325,29 @@ def convert_mongo_objs(value: Any) -> Any:
...
+
+    Returns:
+        pyarrow.lib.Table: The table with the columns converted.
+    """
+    import pymongoarrow  # type: ignore
+    from dlt.common.libs.pyarrow import pyarrow
+
+    for i, field in enumerate(table.schema):
+        if pymongoarrow.types._is_objectid(field.type):
+            col_values = [str(value) for value in table[field.name]]
```
rudolfix, line 342: OK I see we'll not be able to convert to hex in arrow. Still it would be good to avoid creating python lists and loops.

you could use numpy.vectorize with `hex` function. you may need to convert to binary first and then go to numpy array

rudolfix (changes requested):
we can move forward now
1. pls try to optimize this object_id conversion, see my comment
2. use alpha 0.5.1a1 as dlt dependency, your fix is there. make sure all tests are passing

rudolfix (approved):
LGTM!

PR #487, opened 2024-06-05
Sql database reflection level

### LINKED ISSUE #430: allow for minimal table reflection level and allow for custom database types to be recognized in `sql_database`
### Source name

sql_database

`sql_database` has two reflection levels: (1) PK + data types (2) 1 + precision information for int/varchar/binary. this is incompatible with old behavior and also is creating problems when arrow/panda data types are different from reflected data types. in the latter case we should allow `dlt` to infer data types from data (we are not coercing content of the data frames)

1. * [ ] drop `with_precision_hints` and introduce `reflect_table_schema` with `minimal`, `full`, `full_with_precision`
2. * [ ] add a new callback `type_conversion_fallback` that will be called BEFORE any sql alchemy type is  converted to dlt type, the callback should let user to cast or ignore custom sql alchemy types
3. * [ ] make bigquery tests to work (right now they fail on JSON fields which should be converted to string, not complex, use (2))
4. * [ ] pass `backend_kwargs` to `create_engine` for special dialect options. one of the use cases is `arraysize` of `cx_orarcle`, currently we hack it like this
```python
def engine_from_credentials(
    credentials: Union[ConnectionStringCredentials, Engine, str]
) -> Engine:
    if isinstance(credentials, Engine):
        return credentials
    if isinstance(credentials, ConnectionStringCredentials):
        credentials = credentials.to_native_representation()
    return create_engine(credentials, arraysize=1)  # HERE!
```
5. * [ ] make sure CI works
6. * [ ] update README

### Are you a dlt user?

Yes, I'm already a dlt user.

Yes, I'm ready.

### dlt destination

_No response_

### Additional information

_No response_

steinitzu:
### Tell us what you do here

### Short description
- Adds `reflection` level argument: `minimal` (column names + pk + nullability), `full` (minimal + data types) , `full_with_precision` (full + precision)
- `type_conversion_callback` argument to cast one sqla type to another (runs before dlt data type is detected from sql type)
- Tests for both

Test pass on dlt devel locally. May not on release version currently.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #430


sources/sql_database/__init__.py
```diff
@@ -30,11 +35,12 @@ def sql_database(
     table_names: Optional[List[str]] = dlt.config.value,
     chunk_size: int = 50000,
     backend: TableBackend = "sqlalchemy",
-    detect_precision_hints: Optional[bool] = dlt.config.value,
```
rudolfix, line 33: now when I think of it.. let's keep this option and mark as deprecated. when true it should set the highest reflection level. we'll have a lot of users complaining that production code stopped working...

sources/sql_database/__init__.py
```diff
@@ -57,9 +63,13 @@ def sql_database(
         table_adapter_callback: (Callable): Receives each reflected table. May be used to modify the list of columns that will be selected.
         backend_kwargs (**kwargs): kwargs passed to table backend ie. "conn" is used to pass specialized connection string to connectorx.
         include_views (bool): Reflect views as well as tables. Note view names included in `table_names` are always included regardless of this setting.
+        type_conversion_callback(Optional[Callable]): Callable to override type inference when reflecting columns.
```
rudolfix, line 66: `detect_precision_hints` - let's mark it as deprecated

sources/sql_database/__init__.py
```diff
@@ -112,10 +123,11 @@ def sql_table(
     incremental: Optional[dlt.sources.incremental[Any]] = None,
     chunk_size: int = 1000,
     backend: TableBackend = "sqlalchemy",
-    detect_precision_hints: Optional[bool] = dlt.config.value,
```
rudolfix, line 115: same here

sources/sql_database/schema_types.py
```diff
@@ -20,20 +23,42 @@
...
+    sql_t = type_conversion_callback(sql_t) if type_conversion_callback else sql_t
+    if type_conversion_callback:
+        sql_t = type_conversion_callback(sql_t)
+        # Check if sqla type class rather than instance is returned
+        if sql_t is not None and isinstance(sql_t, type):
+            sql_t = sql_t()
+
+    if sql_t is None:
+        # Column ignored by callback
+        return None
```
rudolfix, line 59: please take a look #488 - this is actually a bug that will prevent pyarrow backend from working.
we are converting a list of rows into ndarray and then transpose and split it. to do that a list of columns is used - so we may still return a column but without data type.

please take a look and write a test ie. using GEOMETRY or some other postgres type that we do not handle

rudolfix (changes requested):
this 90% good.
1. we have a related bug #488 see in the review
2. please add a test where we have an unsupported type like GEOMETRY in the table. let's try to load it with and without type adapter
3. we need docs update with examples of type adapter, same thing in sql database pipeline demo

steinitzu:
> this 90% good.
> 
>     1. we have a related bug [sql_database source | error with pyarrow BE, if some of the types were not identified correctly #488](https://github.com/dlt-hub/verified-sources/issues/488) see in the review

I added some logic in `row_tuples_to_arrow` to skip unknown type columns. They are returned without data type in dlt schema to preserve the index. The other backends type is still inferred where possible.  

>     2. please add a test where we have an unsupported type like GEOMETRY in the table. let's try to load it with and without type adapter

Added a kinda massive parametrized test for all backends/type adapter/reflection levels. Using `DATERANGE` and `ARRAY` columns. Didn't want to require postgis or anything and make the test setup more complicated. But could be nice to test with snowflake source since they use all custom types that don't inherit regular sqla types.  

Type adapter can work in some cases, but for instance trying to load the `DATERANGE` object to arrow as complex/text type does not work. We would need to run through the data and coerce values for that which would be a big performance hit.  
 
>     3. we need docs update with examples of type adapter, same thing in sql database pipeline demo

https://github.com/dlt-hub/dlt/pull/1467

sources/sql_database/schema_types.py
```diff
@@ -134,6 +178,12 @@ def row_tuples_to_arrow(
 
     arrow_schema = columns_to_arrow(columns, tz=tz)
 
+    columns_list = list(columns.values())
+    included_columns_idx = [
```
rudolfix, line 182: are you removing columns with not recognized data types? why? we should keep them. `dlt` will later try to map missing columns into dlt schema (in extract). so it should work

sources/sql_database/schema_types.py
```diff
@@ -158,4 +211,4 @@ def row_tuples_to_arrow(
             )
             float_array = pa.array(columnar[field.name], type=pa.float64())
             columnar[field.name] = float_array.cast(field.type, safe=False)
-    return pa.Table.from_pydict(columnar, schema=arrow_schema)
```
rudolfix, line 161: why this is removed? here I'm afraid all the precision information etc. will be lost because pyarrow will infer schema from the data and we know way more from sql schema.

ofc we have a problem with unknown data types. but we can create a table for those columns separately and then join it into one. WDYT? should be cheap, arrow is columnar

sources/sql_database/helpers.py
```diff
@@ -254,6 +254,36 @@ def _unwrap(table: TDataItem) -> TDataItem:
...
+def validate_reflection_level(
+    reflection_level: Optional[ReflectionLevel], backend: TableBackend
+) -> ReflectionLevel:
+    if backend == "pyarrow":
+        if reflection_level is None:
+            return "full"
+        if reflection_level == "minimal":
+            raise ValueError(
+                (
+                    "pyarrow backend requires reflection_level to be 'full' or 'full_with_precision' to infer schema. "
```
rudolfix, line 266: why? schema is optional in https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_pydict

what you should do (see my other comments):
1. create arrow table for known data types
2. create arriow table for unknown data types where pyarrow infers the schema from the data
3. create unified table (preserving original column order)

anyway: you have unknown data types, in minimal all data types are uknown ;>

rudolfix (changes requested):
we need to improve how arrow tables are generated. also linter is not passing...

good work with the tests!

sources/sql_database/schema_types.py
steinitzu: Ah mistake there. Didn't mean to keep this change in :grimacing:

sources/sql_database/helpers.py
steinitzu: That's a good idea!  
We do have some problem with objects that pyarrow doesn't support though. E.g. UUID and some dataclass objects that sqlalchemy returns for `Range` types are rejected, they need to be coerced to str/dict first.   
But I think there's not that many such types (built into sqlalchemy at least) so we could run just those columns through some conversion.

sources/sql_database/arrow_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+        for key in list(columnar_unknown_types):
+            try:
+                parr = columnar_unknown_types[key] = pa.array(
+                    columnar_unknown_types[key]
+                )
+            except pa.ArrowInvalid as e:
+                logger.warning(
+                    f"Column {key} contains a data type which is not supported by pyarrow. This column will be ignored. Error: {e}"
+                )
+                del columnar_unknown_types[key]
```
steinitzu, line 102: @rudolfix does it make sense to do anything else here, what do you think?  
Considering running the column through `dlt.common.json.custom_encode` to coerce it to string or better to leave it to the user?

rudolfix:
@steinitzu is there anything you still want to work on? I merged this PR with #478 because it is fixing a lot of config bugs in `sql_database` and I needed the minimal reflection level here.

so: if you are done with the code you can close this PR. and take a look at #478. which is ready to be merged

rudolfix:
@steinitzu OK 0.5 got released and this PR was merged

Issue #494, opened 2024-06-12
rest_api: Allow the REST API config object to exclude rows, columns, and transform data  


francescomucio:
### Source name

rest_api

### Describe the data you'd like to see

I would like to have the possibility to configure a source like:

```
            {
                "name": "my_nicely_named_resource",
                "endpoint": {
                    "path": "endpoint_name",
                },
                "row_filter": lambda x: x["id"] == 3,
                "exclude_columns": ["$.id", "$.another_column"],
                "transform": my_function,
            },
 ```

### Are you a dlt user?

Yes, I'm already a dlt user.

### Do you ready to contribute this extension?

Yes, I'm ready.

### dlt destination

duckdb, snowflake

PR #495, opened 2024-06-12
rest_api: filter, exclude, transform API responses

francescomucio:
### Tell us what you do here

As per the associated issue, [rest_api: Allow the REST API config object to exclude rows, columns, and transform data](https://github.com/dlt-hub/verified-sources/issues/494):
- when specifying a resource it will be possible to pass:
  - a `row_filter`, in the form of a function or lambda
  - a list of columns (`exclude_columns`), as jsonpaths, that need to be removed from the output
  - a `transform`, in the form of a function or lambda

I have added the additional properties to typing.py and added the code to handle them.

### Related Issues

This solves [this issue](https://github.com/dlt-hub/verified-sources/issues/494)


sources/rest_api/__init__.py
```diff
@@ -254,21 +257,39 @@ def paginate_resource(
...
+                        del item[key]
+                    return item
+
                 if incremental_object:
                     params[incremental_param.start] = incremental_object.last_value
                     if incremental_param.end:
                         params[incremental_param.end] = incremental_object.end_value
 
-                yield from client.paginate(
+                for page in client.paginate(
```
z3z1ma, line 274: Is it guaranteed that `page` is an array and never a dict? Say the API returned `{...}` data in a top level dict and there was never an array in the response.

sources/rest_api/__init__.py
```diff
@@ -254,21 +257,39 @@ def paginate_resource(
                 client: RESTClient = client,
                 incremental_object: Optional[Incremental[Any]] = incremental_object,
                 incremental_param: IncrementalParam = incremental_param,
+                row_filter: Any = row_filter,
+                transform: Any = transform,
+                exclude_columns: List[str] = exclude_columns,
             ) -> Generator[Any, None, None]:
+                def exclude_elements(item: Any, exclude_columns: List[str]) -> Any:
+                    for key in exclude_columns:
+                        del item[key]
```
z3z1ma, line 266: I think a jsonpath would be significantly more powerful exclusion mechanism here

sources/rest_api/__init__.py
```diff
@@ -211,6 +211,9 @@ def create_resources(
         request_params = endpoint_config.get("params", {})
         request_json = endpoint_config.get("json", None)
         paginator = create_paginator(endpoint_config.get("paginator"))
+        row_filter = endpoint_resource.pop("filter", None)
```
z3z1ma, line 214: Add an identity function that just returns the input or is an effective no op for each functional param. This removes the need for the ternaries in the function body.

sources/rest_api/__init__.py
```diff
@@ -254,21 +257,39 @@ def paginate_resource(
...
                     method=method,
                     path=path,
                     params=params,
-                    json=json,
                     paginator=paginator,
                     data_selector=data_selector,
                     hooks=hooks,
-                )
+                ):
+                    filtered_page = filter(row_filter, page) if row_filter else page
```
z3z1ma, line 282:
```python
yield map(exclude_elements, map(transform, filter(row_filter, page)), exclude_columns)
```

Given an identity function which removes the need for ternaries, I think this is more aesthetic :) 
Closer to an applicative functor on the list monad, though itd look more accurate if python supported map/filter as methods on the list class. Then itd be left to right instead of bubbling out.

francescomucio: not in one line, but I did something similar

sources/rest_api/__init__.py
francescomucio: done

sources/rest_api/__init__.py
francescomucio: used jsonpath

burnash:
@francescomucio thank you for your contribution and for proposing the enhancement to the `rest_api` configuration. These are valuable suggestions and they can can certainly improve the flexibility of resource configuration.

However, I see a few potential drawbacks with the proposed interface:

1. Flexibility in defining the precedence of filters and transforms: the current approach of specifying `row_filter`, `exclude_columns`, and `transform` separately might lead to confusion regarding the order of operations. It's crucial to have a clear order of applying these operations to ensure predictable results.

2. Consistency with existing dlt resource methods: one possible improvement is to align with the existing DltResource's methods `add_map()` and `add_filter()`. This could  make the configuration more intuitive for users who are already familiar with the `dlt`. (Most likely the implementation should also be based on add_map() and add_filter() where possilbe.)

To address these issues, I suggest an alternative format that puts all operations in one `operations` list. This makes it clear and flexible by showing the exact order of operations. Taking the example from #494, e.g.:

```python
{
    "name": "my_nicely_named_resource",
    "endpoint": {
        "path": "endpoint_name"
    },
    "operations": [
        {"filter": lambda x: x["id"] == 3},
        {"map": "delete_fields", "fields": ["id", "another_column"]},
        {
            "map": "rename_fields",  
            "fields": {
                "user_id": "my_user_id",
                "timestamp": "my_timestamp"
            }
        },
        {"map": my_function}
    ]
}
```

Some other pre-defined possible filters:

```python
{"filter": "range", "field": "date", "from": "2021-01-01", "to": "2021-12-31"}
```
or
```python
{"filter": "in_set", "field": "category", "values": ["tech", "finance", "health"]}
```
and so on. 

Again, thank you for the suggestion and valuable input. What do you think about this? Looking forward to your feedback.

rudolfix:
Overall this looks good
1. We call those functions `transforms` not `operations`. Maybe we could rename to `transform`?
2. The structure as a list of functions is OK
3. I agree with Francesco: let's start with python functions. If we want to implement a few standard transformations we keep them here: `dlt.sources.helpers.transform` I'm totally for adding functions to rename or delete fields
4. If we do (3) we can skip discussion how to declare functions with arguments in dicts/yaml which we then should reuse everywhere

burnash:
I'm not very firm on "operations", but I was looking for a keyword that could work for both "map" and "filter". In my opinion "transform" is could be a bit specific so it's hard to fit "filter" under it. But I may be wrong. So the alternatives to "operations" could be:

1. "transform" / "transformations"
2. "processing_steps" to better align with underlying API. 

Other than that also agreed with @francescomucio we can start with Python functions.

francescomucio:
I went with `processing_steps`, I have also added an initial test case for a filter, I will add one for a map and maybe something a bit more complex

sources/rest_api/__init__.py
```diff
@@ -253,6 +256,14 @@ def create_resources(
             endpoint_resource, {"endpoint", "include_from_parent"}
         )
 
+        def process(resource, processing_steps) -> Any:
+            for step in processing_steps:
+                if "filter" in step:
+                    resource.add_filter(step["filter"])
+                if "map" in step:
+                    resource.add_map(step["map"])
+            return resource
```
willi-mueller, line 265: awesome!

sources/rest_api/__init__.py
```diff
@@ -278,15 +289,15 @@ def paginate_resource(
                         incremental_cursor_transform,
                     )
 
-                yield from client.paginate(
+                for page in client.paginate(
```
burnash, line 292: Do we still need this change?

sources/rest_api/typing.py
```diff
@@ -254,6 +259,10 @@ class ResourceBase(TypedDict, total=False):
     table_format: Optional[TTableHintTemplate[TTableFormat]]
     selected: Optional[bool]
     parallelized: Optional[bool]
+    processing_steps: Optional[List[ProcessingSteps]]
+    # row_filter: Optional[Callable[[Any], bool]]
+    # transform: Optional[Callable[[Any], Any]]
+    # exclude_columns: Optional[List[jsonpath.TJsonPath]]
```
burnash, line 265: Let's remove the commented code and implement these ready-to-use operation as a follow-up PR

burnash (changes requested):
Looks great, @francescomucio! I've listed a couple of minor improvements. 
Also, let's extend the tests to cover "map" and using processing steps in a child resource. A good improvement is to also extend the "offline"/"mocked" tests (https://github.com/dlt-hub/verified-sources/blob/master/tests/rest_api/test_rest_api_source_offline.py). Let me know if you have capacity for this.

sources/rest_api/__init__.py
```diff
@@ -278,11 +289,10 @@ def paginate_resource(
                         incremental_cursor_transform,
                     )
 
-                yield from client.paginate(
+                yield client.paginate(
                     method=method,
                     path=path,
                     params=params,
-                    json=json,
```
burnash, line 285: Is it a typo or there's a reason to remove `json` param?

sources/rest_api/__init__.py
```diff
@@ -278,11 +289,10 @@ def paginate_resource(
                         incremental_cursor_transform,
                     )
 
-                yield from client.paginate(
```
burnash, line 281: I think we need to keep the `from` statement.

burnash (changes requested):
Thanks for updating the tests. I've to questions about the `yield` in the comments.

sources/rest_api/__init__.py
```diff
@@ -347,7 +361,7 @@ def paginate_dependent_resource(
                         if parent_record:
                             for child_record in child_page:
                                 child_record.update(parent_record)
-                        yield child_page
+                        yield from child_page
```
burnash, line 364:
```suggestion
                        yield child_page
```
In this case we actually need to `yield` the page data

sources/rest_api/__init__.py
```diff
@@ -253,6 +256,18 @@ def create_resources(
             endpoint_resource, {"endpoint", "include_from_parent"}
         )
 
+        def process(
+            resource: DltResource,
+            processing_steps: List[ProcessingSteps] = [],
```
burnash, line 261: Mutable defaults in Python may cause unpredicted behaviour: https://docs.astral.sh/ruff/rules/mutable-argument-default/
Since we always call `processing_steps` the default is not needed here.

burnash (changes requested):
Thanks for the update. Please see my comment about the mutable default.

sources/rest_api/__init__.py
francescomucio: Amazing behaviour :) Fixed

sources/rest_api/__init__.py
```diff
@@ -362,6 +380,10 @@ def paginate_dependent_resource(
                 hooks=hooks,
             )
```
burnash, line 382:
```suggestion
```

sources/rest_api/__init__.py
```diff
@@ -253,6 +256,18 @@ def create_resources(
             endpoint_resource, {"endpoint", "include_from_parent"}
         )
 
+        def process(
+            resource: DltResource,
+            processing_steps: List[ProcessingSteps],
+        ) -> Any:
+
```
burnash, line 263:
```suggestion
```

tests/rest_api/test_rest_api_source_processed.py
```diff
@@ -0,0 +1,256 @@
...
+                    {"filter": lambda x: x["id"] == 1},
+                ],
+            },
+        ],
+    }
+    mock_source = rest_api_source(config)
+
+    data = list(mock_source.with_resources("comments"))
+    assert len(data) == 2
+    # assert data[0]["title"] == "Post 1"
```
burnash, line 211:
```suggestion
```
Remove commented code

tests/rest_api/test_rest_api_source_processed.py
```diff
@@ -0,0 +1,255 @@
...
+                    {"map": extend_body},
+                    {"filter": lambda x: x["body"].startswith("Post 2")},
+                ],
+            },
+        ],
+    }
+    mock_source = rest_api_source(config)
+
+    data = list(mock_source.with_resources("comments"))
+    # assert len(data) == 1
```
burnash, line 254:
```suggestion
```
Remove commented code

burnash (approved):
Great work, thank you @francescomucio!

The examples from the tests are very good we'll use them in a follow-up documentation PR

PR #503, opened 2024-06-19
Test/461 rest api unit tests

### LINKED ISSUE #461: rest api: missing unit tests package
following unit test that check processing of RESTAPIConfig and resource generation

### incremental
- [x] test if Incremental is properly instantiated from IncrementalArgs. This is a unit test for function `setup_incremental_object` and we test that (`Incremental`, `IncrementalParam`) are properly constructed from the config dict.
- [x] test if incremental is properly bound to start_param and end_param (IncrementalConfig). Similar to before, just the second branch of the function.
- [x] test many incrementals in single resource. This should fail.
- [x] incremental bound via parameter. (probably line 166)
- [x] incremental bound via dlt.sources.incremental (probably line 164)

### resource generation
- [x] check if resource receives all resource hints possible (table_name, parallel, max nesting etc.)

### transformers (child resource) generation (declarative interface)
- [x] binding to non existing resource 
- [x] two dependent resources to one resource
- [x] one dependent resources binding two resources (should fail)
- [x] I assume we do not support resource -> transformer -> transformer but also this must be tested,. **Answer: we do support that!**
- [x] circular dependencies. Should fail with `CycleError`, probably already raised by `topological_sort()`.
- [x] test include fields from parent, also non existing fields

### test binding of params
- [x] path params vs query params vs body params (what if path param is not bound) (path and query params are tested)

willi-mueller:
### Tell us what you do here
Implementing rest_api source unit tests #461 

### Related Issues

- Resolves #461

### DIFF
--- a/sources/rest_api/config_setup.py
+++ b/sources/rest_api/config_setup.py
@@ -78,6 +78,7 @@ AUTH_MAP: Dict[AuthType, Type[AuthConfigBase]] = {
 
 PARAM_TYPES: List[ParamBindType] = ["incremental", "resolve"]
 
+
 class IncrementalParam(NamedTuple):
     start: str
     end: Optional[str]
@@ -163,7 +164,9 @@ def setup_incremental_object(
         if isinstance(value, dlt.sources.incremental):
             return value, IncrementalParam(start=key, end=None)
         if isinstance(value, dict) and value.get("type") not in PARAM_TYPES:
-            raise ValueError( f"Invalid param type: {value.get('type')}. Available options: {PARAM_TYPES}")
+            raise ValueError(
+                f"Invalid param type: {value.get('type')}. Available options: {PARAM_TYPES}"
+            )
         if isinstance(value, dict) and value.get("type") == "incremental":
             config = exclude_keys(value, {"type"})
             # TODO: implement param type to bind incremental to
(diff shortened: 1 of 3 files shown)


tests/rest_api/test_configurations.py
```diff
@@ -531,3 +541,47 @@ def test_resource_schema() -> None:
...
+def test_incremental_from_invalid_request_param():
+    request_params = {
+        "foo": "bar",
+        "since": {
+            "type": "no_incremental",
+            "cursor_path": "updated_at",
+            "initial_value": "2024-01-01T00:00:00Z",
+        },
+    }
+    (incremental_config, incremental_param) = setup_incremental_object(request_params)
```
burnash, line 571: I think it should actually throw in this case. Maybe we need to lift this check up the call stack

willi-mueller: I implemented a bugfix in 4e594cc6f198bad5c065fa0fd660233c1d6f9ecb. Is that sufficient?

sources/rest_api/config_setup.py
```diff
@@ -163,6 +162,8 @@ def setup_incremental_object(
     for key, value in request_params.items():
         if isinstance(value, dlt.sources.incremental):
             return value, IncrementalParam(start=key, end=None)
+        if isinstance(value, dict) and value.get("type") not in PARAM_TYPES:
```
burnash, line 165: I think we need to find another place for this check: the problem with the check here is that now `setup_incremental_object()` knows about other PARAM_TYPES: which violates encapsulation. A better way is to have the check up the calling stack before setup_incremental_object() gets called.

burnash (changes requested):
Great progress! Please see my suggestion to improve the checks for param types.

burnash (approved):
Looks good, thank you very much!

PR #504, opened 2024-06-19
fix(mongodb): improve data converter

IlyaFaer:
Towards #484

### DIFF
--- a/sources/mongodb/helpers.py
+++ b/sources/mongodb/helpers.py
@@ -6,12 +6,14 @@ from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Tuple
 import dlt
 from bson.decimal128 import Decimal128
 from bson.objectid import ObjectId
+from bson.regex import Regex
+from bson.timestamp import Timestamp
 from dlt.common import logger
 from dlt.common.configuration.specs import BaseConfiguration, configspec
 from dlt.common.time import ensure_pendulum_datetime
 from dlt.common.typing import TDataItem
 from dlt.common.utils import map_nested_in_place
-from pendulum import _datetime
+from pendulum import _datetime, timezone
 from pymongo import ASCENDING, DESCENDING, MongoClient
 from pymongo.collection import Collection
 from pymongo.cursor import Cursor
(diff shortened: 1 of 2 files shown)


sources/mongodb/helpers.py
```diff
@@ -235,9 +237,53 @@ def convert_mongo_objs(value: Any) -> Any:
         return str(value)
     if isinstance(value, _datetime.datetime):
         return ensure_pendulum_datetime(value)
+    if isinstance(value, Regex):
+        return value.try_compile().pattern
+    if isinstance(value, Timestamp):
+        date = value.as_datetime()
+        date = date.replace(tzinfo=timezone("UTC"))
```
rudolfix, line 244: why `replace` is needed? this just forces UTC over any timezone that was there without conversion.
just run it via `ensure_pendulum_datetime`, it will add UTC timezone if necessary

sources/mongodb/helpers.py
```diff
@@ -235,9 +237,53 @@ def convert_mongo_objs(value: Any) -> Any:
...
+        table (pyarrow.lib.Table): The table to convert.
+
+    Returns:
+        pyarrow.lib.Table: The table with the columns converted.
+    """
+    from pymongoarrow.types import _is_binary, _is_code, _is_decimal128, _is_objectid  # type: ignore
+    from dlt.common.libs.pyarrow import pyarrow
+
+    for i, field in enumerate(table.schema):
+        if _is_objectid(field.type) or _is_decimal128(field.type):
```
rudolfix, line 263: 1. we'll need to port the conversion from #486 maybe you should merge this PR with it so it is easier to manage?
2. there's arrow decimal128 type so why is there a problem with conversion?

sources/mongodb/helpers.py
```diff
@@ -235,9 +237,53 @@ def convert_mongo_objs(value: Any) -> Any:
...
+            col_values = [str(value) for value in table[field.name]]
+            table = table.set_column(
+                i,
+                pyarrow.field(field.name, pyarrow.string()),
+                pyarrow.array(col_values, type=pyarrow.string()),
+            )
+        else:
+            type_ = None
+            if _is_binary(field.type):
+                type_ = pyarrow.binary()
```
rudolfix, line 273: a regular cast is not working? I was able to cast object_id directly to binary so that should also work without using any list ie.
```
def convert_time_to_us(table):
    """map transform converting time column to microseconds (ie. from nanoseconds)"""
    import pyarrow as pa
    from pyarrow import compute as pc

    time_ns_column = table['time_col']
    time_us_column = pc.cast(time_ns_column, pa.time64('us'), safe=False)
    new_table = table.set_column(
        table.column_names.index('time_col'),
        'time_col',
        time_us_column,
    )
    return new_table
```
try to use `pc.cast(` and IMO it will work

tests/mongodb/test_mongodb_source.py
```diff
@@ -243,3 +245,80 @@ def test_order(row_order, last_value_func):
...
+    res = mongodb_collection(collection="types_test", data_item_format="arrow")
+    types = list(res)[0]
+
+    for field, value in {
+        "field2": pyarrow.scalar(True),
+        "field3": pyarrow.scalar(1, type=pyarrow.int32()),
+        "field4": pyarrow.scalar(1),
+        "field5": pyarrow.scalar(1.2),
+        "field6": pyarrow.scalar("text"),
+        "field7": pyarrow.scalar([1, 2, 3], type=pyarrow.list_(pyarrow.int32())),
```
rudolfix, line 298: is this thing really loading?

rudolfix (changes requested):
IMO you should move this code to #486 or branch from it. how are you running the tests without arrow code?

sources/mongodb/helpers.py
IlyaFaer: The thing is that values originally written in the database are returned in another type after this converter. `decimal128` is one of those cases. To ensure that they written and read data is exactly the same, I had to add these lines for decimal and ids.

sources/mongodb/helpers.py
IlyaFaer: Done

sources/mongodb/helpers.py
IlyaFaer: No, it's not working because of the same reasons I used column replacement in the #486 

If we just cast a column, it may turn all the values into `null`, or, in case of binary, it'll not load into the destination (it'll be converted, but fail on load)

IlyaFaer:
The PR is merged into #486

PR #508, opened 2024-06-24
rest_api: unit tests (part 2)

### LINKED ISSUE #461: rest api: missing unit tests package

willi-mueller:
### Tell us what you do here

Implementing tests.

### Short description
Unit test package 2.

### Related Issues

- Resolves #461

### DIFF
--- a/tests/rest_api/test_configurations.py
+++ b/tests/rest_api/test_configurations.py
@@ -5,6 +5,8 @@ from unittest.mock import patch
 from copy import copy, deepcopy
 from typing import cast, get_args
 
+from graphlib import CycleError
+
 import dlt
 from dlt.common.utils import update_dict_nested, custom_environ
 from dlt.common.jsonpath import compile_path
@@ -673,6 +675,47 @@ def test_resource_hints():
             assert arg in kwargs.items()
 
 
+def test_two_resources_dependent_on_one_parent_resource():
+    config: RESTAPIConfig = {
+        "client": {
+            "base_url": "https://api.example.com",
+        },
+        "resources": [
+            "users",
+            {
+                "name": "user_details",
+                "endpoint": {
+                    "path": "user/{user_id}/",
+                    "params": {
+                        "user_id": {
+                            "type": "resolve",
+                            "field": "id",
+                            "resource": "users",
+                        },
+                    },
+                },
+            },
+            {
+                "name": "meetings",
+                "endpoint": {
+                    "path": "meetings/{user_id}/",
+                    "params": {
+                        "user_id": {
+                            "type": "resolve",
+                            "field": "id",
+                            "resource": "users",
+                        },
+                    },
+                },
+            },
+        ],
+    }
+    resources = rest_api_resources(config)
+
+    resource_names = [resource.name for resource in resources]
+    assert ["meetings", "user_details", "users"] == sorted(resource_names)
+
+
 def test_dependent_resource_with_multiple_parameter_bindings():
     config: RESTAPIConfig = {
         "client": {
@@ -684,8 +727,6 @@ def test_dependent_resource_with_multiple_parameter_bindings():
                 "name": "user_details",
                 "endpoint": {
                     "path": "user/{user_id}/{group_id}",
-                    "paginator": None,
-                    "data_selector": None,
                     "params": {
                         "user_id": {
                             "type": "resolve",
@@ -725,8 +766,6 @@ def test_one_resource_binding_two_parents():
                 "name": "user_details",
                 "endpoint": {
                     "path": "user/{user_id}/{group_id}",
-                    "paginator": None,
-                    "data_selector": None,
                     "params": {
                         "user_id": {
                             "type": "resolve",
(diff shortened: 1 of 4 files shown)


tests/rest_api/test_configurations.py
```diff
@@ -653,26 +656,220 @@ def test_resource_hints():
...
+            "columns": {"a_text": {"name": "a_text", "data_type": "text"}},
+            "primary_key": "a_pk",
+            "merge_key": "a_merge_key",
+            "schema_contract": {"tables": "evolve"},
+            "table_format": "iceberg",
+            "selected": False,
+        }
+        for arg in expected_kwargs.items():
+            _, kwargs = mock_resource_constructor.call_args_list[0]
+            assert arg in kwargs.items()
```
willi-mueller, line 675: as requested by Marcin: instead of constructing the resource we just assert that resource construction was invoked as expected. The core tests then test the resource construction

tests/rest_api/test_configurations.py
```diff
@@ -653,26 +656,220 @@ def test_resource_hints():
...
+                },
+            },
+        ],
+    }
+
+    with pytest.raises(ValueError) as e:
+        rest_api_resources(config)
+
+    error_part_1 = re.escape(
+        "Multiple resolved params for resource user_details: [ResolvedParam(param_name='user_id'"
```
willi-mueller, line 790: We always have the same error message when we have multiple resolved parameters in a child resource – no matter if all come from the same parent or from different parents. I think that's OK.

tests/rest_api/test_configurations.py
```diff
@@ -653,26 +656,220 @@ def test_resource_hints():
...
-        for _, spec in schema["columns"].items()
-        if spec.get("merge_key", False)
-    ]
-    assert merge_keys == ["a_merge_key"]
-    assert resources[0].schema_contract == {"tables": "evolve"}
-    assert schema.get("table_format") == "iceberg"
-    assert resources[0].selected is False
-    # TODO: test if it is parallelized and has spec
+    resource_names = [resource.name for resource in resources]
+    assert ["location_details", "locations", "meetings"] == sorted(resource_names)
```
willi-mueller, line 835: I could not find a way to test that we have parent -> child -> child. 

I tried playing with `resource._get_parent_pipe()` but no success

```
(Pdb) resources[0].name
'locations'
(Pdb) resources[1].name
'location_details'
(Pdb) p resources[1]._get_parent_pipe("foo", resources[0])
Pipe locations [steps: 2] at 5486406096
(Pdb) p resources[1]._get_parent_pipe("foo", resources[1])
Pipe location_details [steps: 2] at 5401533648 data bound to Pipe locations [steps: 2] at 5486406096
```

Do you know of a way to assert that a resource has a specific parent?

burnash: Try `resources[0]._pipe.parent.name`, and check from the innermost child resource (e.g. "meetings")

tests/rest_api/test_configurations.py
burnash: Yes, I think so too

tests/rest_api/test_configurations.py
```diff
@@ -653,26 +656,220 @@ def test_resource_hints():
...
+                        },
+                    },
+                },
+            },
+        ],
+    }
+    resources = rest_api_resources(config)
+
+    resource_names = [resource.name for resource in resources]
+    assert ["meetings", "user_details", "users"] == sorted(resource_names)
```
burnash, line 716: Let's also check if the dependency has been established by checking `resource._pipe.parent.name`

burnash (changes requested):
Great progress! Let's try to explicitly check the relationship (e.g. `resources._pipe.parent.name`) and I think it's good to go.

tests/rest_api/test_configurations.py
willi-mueller: great, this is what I was looking for. The test is much better now!

willi-mueller:
@burnash I just rebased on master. Here are the changes after your review: https://github.com/dlt-hub/verified-sources/pull/508/commits/aebe65a701472e4456761343ba66893c6b694976

burnash (approved):
Looks good! Great job!

PR #511, opened 2024-06-26
rest_api: Test/505 hooks in response actions

### LINKED ISSUE #505: rest_api: Extend `response_actions` to accept hook functions
### Background

Currently, the `response_actions` configuration within the REST API source only supports basic string actions like "ignore".

```py
{
    "path": "issues",
    "response_actions": [
        {"status_code": 404, "action": "ignore"},
        {"content": "Not found", "action": "ignore"},
        {"status_code": 200, "content": "some text", "action": "ignore"},
    ],
}
```

### Proposal

Extend the `response_actions` configuration to:

1. Accept Python function as "action" to allow for more complex actions:

```py

def custom_action(response):
    # Custom logic here
    return response

# ...
{
    "path": "issues",
    "response_actions": [
        {"status_code": 404, "action": custom_action},
    ],
}
```

2. Accept a function as an action for all response statuses and content:

```py

def custom_action(response):
    # Custom logic here
    return response

# ...
{
    "path": "issues",
    "response_actions": [
        custom_action,
    ],
}
```

3. Accept multiple functions as an action :

```py

def custom_action_1(response):
    # Custom logic here
    return response

def custom_action_2(response):
    # Custom logic here
    return response

def custom_action_3(response):
    # Custom logic here
    return response

# ...
{
    "path": "issues",
    "response_actions": [
        custom_action_1,
        custom_action_2
    ],
}

{
    "path": "issues",
    "response_actions": [
        custom_action_1,
        {"status_code": 200, "action": custom_action_2 }
    ],
}

{
    "path": "issues",
    "response_actions": [
        custom_action_1,
        {"status_code": 200, "action": [custom_action_2, custom_action_3] }
    ],
}
```

### Task 

- [x] Update `source/rest_api/config_setup.py` to accept hook functions as actions
- [x] Update the documentation
- [x] Add tests

willi-mueller:
### Tell us what you do here

- improvements: allows one or more callable response_actions to be executed (#505)

- fixing defects:
  - `removes "response_actions":  {"status_code": 429, "action": "retry" }` from the docs because only `action == ignore"` was implemented
  - only the first element in `response_actions` was registered as a hook
  

### Short description

#505 

### Related Issues

- Resolves #505


sources/rest_api/config_setup.py
```diff
@@ -349,50 +350,96 @@ def _find_resolved_params(endpoint_config: Endpoint) -> List[ResolvedParam]:
...
-        action_type: str = action.get("action")
+        content_substr = action.get("content")
+        response_action = action.get("action")
+        if isinstance(response_action, str):
+            action_type = response_action
+        elif callable(response_action):
+            custom_hook = [response_action]
+        elif isinstance(response_action, list) and all(
+            callable(action) for action in response_action
+        ):
```
willi-mueller, line 393: We probably can't replace these conditionals on runtime type checks with polymorphism – unless we require the user to pass in our custom classes which would be cumbersome for them.

sources/rest_api/config_setup.py
```diff
@@ -403,16 +450,30 @@ def create_response_hooks(
...
+            modified_content: bytes = json.dumps(payload).encode("utf-8")
+            response._content = modified_content
+            return response
+
         response_actions = [
+            set_encoding,
             {"status_code": 404, "action": "ignore"},
             {"content": "Not found", "action": "ignore"},
-            {"status_code": 429, "action": "retry"},
-            {"status_code": 200, "content": "some text", "action": "retry"},
```
willi-mueller, line 410: "retry" is not implemented and thus it's removed from the docs

sources/rest_api/config_setup.py
```diff
@@ -403,16 +450,30 @@ def create_response_hooks(
...
+            response.encoding = 'windows-1252'
+            return response
+
+        def remove_field(response: Response, *args, **kwargs) -> Response:
+            payload = response.json()
+            for record in payload:
+                del record["email"]
+            modified_content: bytes = json.dumps(payload).encode("utf-8")
+            response._content = modified_content
+            return response
```
willi-mueller, line 463: I find it a bit cumbersome to unpack the response manually – the detector does it so well and now the user has to do it again by hand. 

Maybe users would prefer a different API for the use cases of adding or deleting fields. Instead of operating on the raw response I'd prefer to operate on the parsed records.

sources/rest_api/config_setup.py
```diff
@@ -403,16 +450,30 @@ def create_response_hooks(
...
-            {"status_code": 429, "action": "retry"},
-            {"status_code": 200, "content": "some text", "action": "retry"},
+            {"status_code": 200, "content": "some text", "action": "ignore"},
+            {"status_code": 200, "action": remove_field},
         ]
         hooks = create_response_hooks(response_actions)
     """
     if response_actions:
-        return {"response": [_create_response_actions_hook(response_actions)]}
+        hooks = [_create_response_action_hook(action) for action in response_actions]
```
willi-mueller, line 475: now, we do not only register the first action but all of them

PR #515, opened 2024-07-02
Feat/507 transformation function for incremental cursor

### LINKED ISSUE #507: rest_api: Allow to specify a conversion function for the cursor field in incremental load configuration
### Background

Currently when defining incremental load configuration, the value type of the cursor field is expected to be the same as the value of the parameter passed to the API. 

For example, if the API expects a numeric timestamp value (e.g. 1718991859) for the `date_from` parameter, the cursor field should also be a numeric timestamp value.

### Problem

Some APIs expect a date string (e.g. "2022-01-01") for the querystring parameter, but the response includes a numeric timestamp value. 

### Proposal

Allow users to specify a transformation function that will be applied to the cursor field value before it is used to query the API.

```python
"from_date": {
    "type": "incremental",
    "cursor_path": "time",
    "start_value": date_from,
    "transform": "epoch_to_date"
},
```

Or using a Python function:

```python
def epoch_to_date(epoch):
    return datetime.fromtimestamp(epoch).strftime('%Y-%m-%d')
```

```python
"from_date": {
    "type": "incremental",
    "cursor_path": "time",
    "start_value": date_from,
    "transform": epoch_to_date
},
```

Relevant Slack discussions: 
- https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1718943814032609
- https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1719391770505319

### Implementation
- [x] implement feature
- [x] implement test suite
- [x] write documentation

willi-mueller:
### Tell us what you do here

### Related Issues

- Resolves #507 

### TODO
- [x] implement an integration test with a full source config so that we are sure that the request is actually altered.
- [x] squash and rebase onto master (resolve conflict)

### DIFF
--- a/sources/rest_api/__init__.py
+++ b/sources/rest_api/__init__.py
@@ -222,6 +222,7 @@ def create_resources(
         (
             incremental_object,
             incremental_param,
+            incremental_cursor_transform,
         ) = setup_incremental_object(request_params, endpoint_config.get("incremental"))
 
         client = RESTClient(
@@ -249,12 +250,13 @@ def create_resources(
                 hooks: Optional[Dict[str, Any]],
                 client: RESTClient = client,
                 incremental_object: Optional[Incremental[Any]] = incremental_object,
-                incremental_param: IncrementalParam = incremental_param,
+                incremental_param: Optional[IncrementalParam] = incremental_param,
+                incremental_cursor_transform: Optional[Callable[..., Any]] = incremental_cursor_transform
             ) -> Generator[Any, None, None]:
                 if incremental_object:
-                    params[incremental_param.start] = incremental_object.last_value
-                    if incremental_param.end:
-                        params[incremental_param.end] = incremental_object.end_value
+                    params = _set_incremental_params(
+                        params, incremental_object, incremental_param, incremental_cursor_transform
+                    )
 
                 yield from client.paginate(
                     method=method,
@@ -296,12 +298,13 @@ def create_resources(
                 resolved_param: ResolvedParam = resolved_param,
                 include_from_parent: List[str] = include_from_parent,
                 incremental_object: Optional[Incremental[Any]] = incremental_object,
-                incremental_param: IncrementalParam = incremental_param,
+                incremental_param: Optional[IncrementalParam] = incremental_param,
+                incremental_cursor_transform: Optional[Callable[..., Any]] = incremental_cursor_transform
             ) -> Generator[Any, None, None]:
                 if incremental_object:
-                    params[incremental_param.start] = incremental_object.last_value
-                    if incremental_param.end:
-                        params[incremental_param.end] = incremental_object.end_value
+                    params = _set_incremental_params(
+                        params, incremental_object, incremental_param, incremental_cursor_transform
+                    )
 
                 for item in items:
                     formatted_path, parent_record = process_parent_data_item(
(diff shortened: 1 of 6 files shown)


sources/rest_api/__init__.py
```diff
@@ -232,6 +222,7 @@ def create_resources(
         (
             incremental_object,
             incremental_param,
+            incremental_cursor_transform,
```
willi-mueller, line 225: this value here is not accessible inside the `paginate_resource()` and `paginate_dependent_resource()`

sources/rest_api/__init__.py
```diff
@@ -259,12 +250,12 @@ def paginate_resource(
...
                 incremental_object: Optional[Incremental[Any]] = incremental_object,
-                incremental_param: IncrementalParam = incremental_param,
+                incremental_param: Optional[IncrementalParam] = incremental_param,
             ) -> Generator[Any, None, None]:
                 if incremental_object:
-                    params[incremental_param.start] = incremental_object.last_value
-                    if incremental_param.end:
-                        params[incremental_param.end] = incremental_object.end_value
+                    params = _set_incremental_params(
+                        params, incremental_object, incremental_param, incremental_cursor_transform
```
willi-mueller, line 257: `incremental_cursor_transform` is always `None` at runtime outside of the unit tests. 
Is extending the function signatures the only way?

burnash: Most likely, there's an alternative way, but I haven't discovered it yet. Let's use explicit params and experiment with the alternative implementation as a follow-up.

tests/rest_api/test_configurations.py
```diff
@@ -600,10 +612,11 @@ def test_constructs_incremental_from_endpoint_config_incremental():
             "end_param": "until",
             "cursor_path": "updated_at",
             "initial_value": "2024-01-25T11:21:28Z",
+            "transform": lambda epoch: pendulum.from_timestamp(int(epoch)),
```
burnash, line 615: We may also need to keep the case where `transform` is missing or None.

willi-mueller: Good idea! Now we test with a callable argument and without.

sources/rest_api/__init__.py
willi-mueller: excellent, thank you for the clarification.

tests/rest_api/test_rest_api_source_offline.py
```diff
@@ -288,3 +289,32 @@ def test_load_mock_api_typeddict_config(mock_api_server):
 
     assert table_counts["posts"] == 100
     assert table_counts["post_comments"] == 5000
+
+
+def test_posts_with_inremental_date_transformation(mock_api_server) -> None:
```
burnash, line 294: This is great!

burnash (approved):
Great work, thank you @willi-mueller!

tests/rest_api/test_rest_api_source_offline.py
willi-mueller: An even better way might be to spy on the request client and assert that it was called with the transformed query string params. If you agree, we could do that in a refactoring PR and get the feature shipped for now.

willi-mueller: done in https://github.com/dlt-hub/verified-sources/pull/519

PR #526, opened 2024-07-11
derive decimal type from sql alchemy before attempting integers

rudolfix:
### Tell us what you do here

### Short description
Oracle NUMBER type derives both from Numeric and Integer and is used to represent decimals and float (integers have separate type)
```py
class NUMBER(sqltypes.Numeric, sqltypes.Integer):
    __visit_name__ = "NUMBER"
```
type inference was changed so Numeric is checked first and then integers


steinitzu (approved):
Perfect!

PR #529, opened 2024-07-15
feat(mongodb): implement data filter for mongodb source

### LINKED ISSUE #521: add filter to mongodb source
### Source name

mongodb

Add a filter parameter to the MongoDB source that will be applied before reading data from MongoDB. This feature is useful for creating incremental sources based on fields that could be null or may not exist (e.g., callback_date).

### Are you a dlt user?

Yes, I'm already a dlt user.

Yes, I'm ready.

### dlt destination

_No response_

### Additional information

_No response_

IlyaFaer:
Closes #521

### DIFF
--- a/tests/mongodb/test_mongodb_source.py
+++ b/tests/mongodb/test_mongodb_source.py
@@ -6,6 +6,7 @@ from pendulum import DateTime, timezone
 from unittest import mock
 
 import dlt
+from dlt.pipeline.exceptions import PipelineStepFailed
 
 from sources.mongodb import mongodb, mongodb_collection
 from sources.mongodb_pipeline import (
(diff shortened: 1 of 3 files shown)


tests/mongodb/test_mongodb_source.py
```diff
@@ -356,3 +357,50 @@ def test_arrow_types(destination_name):
...
+    pipeline = dlt.pipeline(
+        pipeline_name="mongodb_test",
+        destination=destination_name,
+        dataset_name="mongodb_test_data",
+        full_refresh=True,
+    )
+    movies = mongodb_collection(
+        collection="movies",
+        incremental=dlt.sources.incremental("runtime", initial_value=20),
+        filter_={"runtime": {"$gte": 20}},
```
Pipboyguy, line 402: would it be worth adding tests for when $lts < $gte filter for the same field or thinking about whether we should handle this in some way?

Pipboyguy (approved):
There's no exception or warning raised when $lte < $gte for example:

```python
    incremental = dlt.sources.incremental(
        "released",
        initial_value="1917-01-01T00:00:00Z"
    )

    user_filter = {
        "released": {
            "$lte": "1800-06-01T00:00:00Z"
        }
    }

    # with pytest.raises(PipelineStepFailed) as excinfo:
    movies = mongodb_collection(
        collection="movies",
        incremental=incremental,
        filter_=user_filter
    )
```

Is this something we want to test or leave as is?

Otherwise LGTM

rudolfix:
@Pipboyguy thanks for checking this out! I think we should let people to put any condition they want even if that makes sense. we do the same with sql alchemy AFAIK

PR #532, opened 2024-07-24
rest_api: Fix dlt init JSONLinkPaginator import error for 0.5.2a1

burnash:


willi-mueller (approved):
Works for me. If I have a dlt version < 0.5.2a1 I need to update dlt after having initialized the rest_api source.

PR #538, opened 2024-07-24
Implement additional test for default resource settings merging

burnash:
Additional tests for #535


willi-mueller (approved):
Very good, these test cases were missing indeed!

PR #540, opened 2024-07-25
fixes unused-ignore errors under mypy 1.10

willi-mueller:
### Short description
This PR proposes to increased developer happiness by fixing these type errors occurring under mypy v1.10.0. This is the version that we require in our pyproject.toml.


willi-mueller:
It's now part of https://github.com/dlt-hub/verified-sources/pull/545

PR #542, opened 2024-07-29
rest_api: mask secrets in config

willi-mueller:
### Tell us what you do here
fixing a bug (please link a relevant bug report)

### Short description

The masking of secrets works as follows:
1. deepcopy the configuration dict
2. take the `auth` configuration and replace all values for allowed sensitive keys with masked versions
3. validate the deep copy where sensitive values are replaced

Assumption:
Authentication objects that are not dictionaries are expected not to print their sensitive values when `__str__()` is called.

Complexity:
- The auth configuration allows both config dictionaries and specific classes. Thus, we need to do runtime type checks.
- implementing polymorphic dispatch, e.g. `auth_config.mask_secrets()` for dicts that are also callable (HttpBasicAuth, BearerTokenAuth, APIKeyAuth) requires an update in the dlt core library. See implementation here: https://github.com/dlt-hub/dlt/issues/1474

### Related Issues

- Fixes https://github.com/dlt-hub/dlt/issues/1474

### DIFF
--- a/sources/rest_api/__init__.py
+++ b/sources/rest_api/__init__.py
@@ -17,6 +17,7 @@ from dlt.sources.helpers.rest_client import RESTClient
 from dlt.sources.helpers.rest_client.paginators import BasePaginator
 from dlt.sources.helpers.rest_client.typing import HTTPMethodBasic
 from .typing import (
+    AuthConfig,
     ClientConfig,
     ResolvedParam,
     ResolveParamConfig,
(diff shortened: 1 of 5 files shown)


sources/rest_api/__init__.py
```diff
@@ -350,6 +358,60 @@ def paginate_dependent_resource(
...
+        or has_sensitive_key
+    ):
+        return _mask_secrets_dict(auth_config)
+    # Here, we assume that OAuth2 and other custom classes that don't implement __get__()
+    # also don't print secrets in __str__()
+    # TODO: call auth_config.mask_secrets() when that is implemented in dlt-core
+    return auth_config
+
+
+SENSITIVE_KEYS: List[str] = [
```
burnash, line 390: Let's move this to the top of the file.

sources/rest_api/__init__.py
```diff
@@ -350,6 +358,60 @@ def paginate_dependent_resource(
...
+            auth_config[sensitive_key] = _mask_secret(auth_config[sensitive_key])  # type: ignore[literal-required, index]
+        except KeyError:
+            continue
+    return auth_config
+
+
+def _mask_secret(secret: Optional[str]) -> str:
+    if secret is None:
+        return "None"
+    if len(secret) < 3:
```
burnash, line 410: Potential improvement is to make "3" a `MIN_SECRET_LENGTH = 3` module-level constant

burnash (changes requested):
Great work, I really like the implementation. I left a couple of commends regarding some small improvements.

sources/rest_api/__init__.py
willi-mueller: Oh, I wanted to have a <3 in the code, but it's better to refactor magic numbers to telling constant names

PR #547, opened 2024-08-02
fix: [sql_database] Handle type ARRAY

vivingo:
Handle type ARRAY in source sql_database and maps it to `data_type` "complex"

### DIFF
--- a/tests/sql_database/sql_source.py
+++ b/tests/sql_database/sql_source.py
@@ -165,6 +165,7 @@ class SQLAlchemySourceDB:
                 Column("json_col", JSONB, nullable=nullable),
                 Column("bool_col", Boolean, nullable=nullable),
                 Column("uuid_col", Uuid, nullable=nullable),
+                Column("array_col", ARRAY, nullable=nullable),
             )
 
         _make_precision_table("has_precision", False)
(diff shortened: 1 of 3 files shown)


tests/sql_database/sql_source.py
```diff
@@ -165,6 +165,7 @@ def _make_precision_table(table_name: str, nullable: bool) -> Table:
                 Column("json_col", JSONB, nullable=nullable),
                 Column("bool_col", Boolean, nullable=nullable),
                 Column("uuid_col", Uuid, nullable=nullable),
+                Column("array_col", ARRAY, nullable=nullable),
```
rudolfix, line 168: You need to provide `ARRAY` type in the constructor. I enabled CI for your PR all tests are failing here

rudolfix (changes requested):
thanks for this fix! Implementation itself is good and I expect it to work. Please see the review, tests should run automatically this time. Let's make sure that arrow backend also works...

tests/sql_database/sql_source.py
vivingo: your're right, sorry ! just updated that

rudolfix:
@vivingo are we able to push to your fork? we are happy to look at the tests and fix them

vivingo:
@rudolfix yes, I enabled the "Allow edits by maintainers" checkbox. 
Do I need to give write access on my fork repository, or will it be done automatically ? 
Thank you ! 

(I don't understand why my tests succeed when I test locally, sorry for that. I just pushed a new change but I guess I might  have some further errors)

vivingo:
@rudolfix any update on this ? 
Do I need to give other authorizations on my fork ?

PR #548, opened 2024-08-05
Update the dlt version to stable 0.5.2

burnash:


willi-mueller (approved):
Good! I synced my poetry env and everything is still working. 
Feels great to be back to stable!

Issue #531, opened 2024-07-22
Add column selection and filtering to sql_database source


arjun-panchmatia-mechademy:
### Source name

sql_database

### Describe the data you'd like to see

Background:
A fairly common use-case when dealing with clients is having a direct database exposed for extraction. Considering the non-technical nature of these clients, these DBs often have lots of columns that may be irrelevant during the extraction process AND/OR require custom column-wise filtering to make the relevant data available for ingestion. This is best expressed as a SQL query where you can select individual columns and apply filters as needed.
Currently, the `sql_database` source does not have the ability to do level of filtering or column selection. This makes it hard to use directly in pipelines, as often applying a map function still results in a full scan, slowing down critical loads. We often have instances of extreme volumes of data for even 15 minutes of incremental loads and cannot use this source as it would be far too slow, and potentially eat up a lot of in-memory store. We've been working around this by writing custom .sql files and yielding them from a resource, but that can be quite cumbersome from a maintenance perspective and does not allow us to leverage the full breadth of dlt's features.

Request:
It would be great if we can get some level of control on the data loaded by the sql_database source (or the sql_table) resource. Potentially letting us select our required columns (maybe via a data_selector as is configred in the rest api source) and allow a level of filtering to occur. Aggregation would also be a generally useful function to have but that may be overkill and is probably better suited to post processing / ELT methadology anyway.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Do you ready to contribute this extension?

No.

### dlt destination

filesystem

### Additional information

I think just having the ability to select columns and apply filters would go a long way. A big benefit of using ORMs is being DB agnostic. When trying to write an orchestration package, we need to necessarily account for variance and try to be as agnostic as possible, and this would go a long way towards helping us achieve that goal.

Current workaround is as described above: writing raw SQL / using an ORM and then yielding the data from the resource.

rudolfix:
@arjun-panchmatia-mechademy we have `table_adapter` callback that let's you modify the SQL Alchemy Table before it is used for select. that allows to remove certain columns ie. (it is in the sql_database_pipeline.py demo)
```py
def select_columns() -> None:
    """Uses table adapter callback to modify list of columns to be selected"""
    pipeline = dlt.pipeline(
        pipeline_name="rfam_database",
        destination="duckdb",
        dataset_name="rfam_data_cols",
        full_refresh=True,
    )

    def table_adapter(table: Table) -> None:
        print(table.name)
        if table.name == "family":
[...]

PR #550, opened 2024-08-08
SQL database: included_columns option

steinitzu:
### Tell us what you do here

### Short description

This adds `included_columns` option for sql database ( I would have named it `columns` but that was conflicting with the `$COLUMNS` shell variable in config resolution)   

Can be set with argument or config in in the standalone resource, and with config in the database source, e.g.:

```toml
[sources.sql_database.campaign]
included_columns = ["id", "name", "created_at", "updated_at"]
```

### Related Issues
One part of https://github.com/dlt-hub/verified-sources/issues/531#issuecomment-2267646842  
The query editing callback I would do in another PR

### DIFF
--- a/sources/sql_database/__init__.py
+++ b/sources/sql_database/__init__.py
@@ -101,22 +101,13 @@ def sql_database(
         tables = list(metadata.tables.values())
 
     for table in tables:
-        if not defer_table_reflect:
-            default_table_adapter(table)
-            if table_adapter_callback:
-                table_adapter_callback(table)
-
-        yield dlt.resource(
-            table_rows,
-            name=table.name,
-            primary_key=get_primary_key(table),
-            spec=SqlDatabaseTableConfiguration,
-            # columns hint will be set at runtime, after included_columns setting has been resolved
-        )(
-            engine,
-            table,
-            chunk_size,
-            backend,
+        yield sql_table(
+            credentials=credentials,
+            table=table.name,
+            schema=table.schema,
+            metadata=metadata,
+            chunk_size=chunk_size,
+            backend=backend,
             reflection_level=reflection_level,
             defer_table_reflect=defer_table_reflect,
             table_adapter_callback=table_adapter_callback,
@@ -187,7 +178,7 @@ def sql_table(
     engine.execution_options(stream_results=True, max_row_buffer=2 * chunk_size)
     metadata = metadata or MetaData(schema=schema)
 
-    table_obj = Table(
+    table_obj = metadata.tables.get("table") or Table(
         table, metadata, autoload_with=None if defer_table_reflect else engine
     )
     if not defer_table_reflect:
--- a/sources/sql_database/__init__.py
+++ b/sources/sql_database/__init__.py
@@ -182,7 +182,7 @@ def sql_table(
         table, metadata, autoload_with=None if defer_table_reflect else engine
     )
     if not defer_table_reflect:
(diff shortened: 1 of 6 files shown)


sources/sql_database/__init__.py
```diff
@@ -111,7 +111,7 @@ def sql_database(
             name=table.name,
             primary_key=get_primary_key(table),
             spec=SqlDatabaseTableConfiguration,
-            columns=table_to_columns(table, reflection_level, type_adapter_callback),
+            # columns hint will be set at runtime, after included_columns setting has been resolved
```
steinitzu, line 114: @rudolfix is there any sense in trying to resolve the config there before resource is called? The column hints are now added with filtered columns in `table_rows` (same as with deferred reflection mode before)

rudolfix: hmmm `sql_table` actually resolves config before we create resource. maybe we just call it here and yield what it returns?

sources/sql_database/__init__.py
```diff
@@ -142,6 +142,7 @@ def sql_table(
     table_adapter_callback: Callable[[Table], None] = None,
     backend_kwargs: Dict[str, Any] = None,
     type_adapter_callback: Optional[TTypeAdapter] = None,
+    included_columns: Optional[List[str]] = None,
```
rudolfix, line 145: maybe we should just name it columns? WDYT?

sources/sql_database/schema_types.py
```diff
@@ -139,13 +139,15 @@ def table_to_columns(
...
     type_conversion_fallback: Optional[TTypeAdapter] = None,
+    included_columns: Optional[List[str]] = None,
 ) -> TTableSchemaColumns:
     """Convert an sqlalchemy table to a dlt table schema."""
     return {
         col["name"]: col
         for col in (
             sqla_col_to_column_schema(c, reflection_level, type_conversion_fallback)
             for c in table.columns
+            if included_columns is None or c.name in included_columns
```
rudolfix, line 150: this certainly works. maybe we could remove columns from the Table object? so for example when table adapter is called user sees right set of columns?

or rename included_columns to selected_columns :)

rudolfix (changes requested):
this is almost LGTM!

sources/sql_database/__init__.py
steinitzu: Hmm maybe that would work. I'll try it.

sources/sql_database/__init__.py
steinitzu: I wanted to but when there's a `COLUMNS` shell/env variable then everything fails. The config resolver tries to read that as a fallback when you don't set the option.

sources/sql_database/schema_types.py
steinitzu: yep deleting the sqla columns is simpler. Doing this now in the default table adapter https://github.com/dlt-hub/verified-sources/pull/550/files#diff-5d691408095c848d7a6e68ba31e82c21c396fb280c655ad5f09d68f579bba94eR41

sources/sql_database/__init__.py
steinitzu: Ok re-using `sql_table` resource now, seems to work fine and reduces duplication a bit.

steinitzu:
@rudolfix please check the updated version and let me know if any other changes needed

rudolfix (approved):
LGTM!

PR #556, opened 2024-08-14
rest_api: pluggable custom auth (Feat/524)

### LINKED ISSUE #524: rest_api: Pluggable Custom Authentication/Authorization
### Background

Currently in `rest_api` source, we have a predefined set of auth methods (`BearerTokenAuth`, `APIKeyAuth`,  `HttpBasicAuth`). This works well for built-in auth it lacks flexibility for users who wish to introduce custom auth logic.

### Problem

Users must modify the `rest_api.config_setup.AUTH_MAP` to add new paginators.

Also, we have implemented `OAuth2ClientCredentials` and `OAuthJWTAuth` but users cannot configure them with the declarative interface because they are not present in the `AUTH_MAP`.

### Proposal

Introduce a pluggable auth system that allows users to define and register custom auth methods dynamically. This should have a registration function to add new auth methods without adding duplicates.

Also, add the 

#### Example

```python
from dlt.sources.helpers.rest_client.auth import AuthConfigBase

class CustomAuth(AuthConfigBase):
    ...

# Register the custom paginator
register_paginator("custom_auth", CustomAuth)
```

### Tasks

- [x] Update the `rest_api` source to support pluggable auth.
- [x] Add tests.
- [x] Update the docs in the dlt repo.

Follow-up:
- #557

willi-mueller:
### Short description
This PR:
- Updates the rest_api source to support pluggable authorization/authentication with custom auth classes
- renames test_paginators.py to test_config_custom_paginators.py to clarify that we don't test paginators but only the configurations of custom paginators.

### Related Issues

- Resolves #524 
- See docs changes: https://github.com/dlt-hub/verified-sources/issues/524


burnash (approved):
Awesome work, @willi-mueller!

PR #570, opened 2024-09-28
Read start offset from partition instead of 0

### LINKED ISSUE #569: Kafka: fails to load partitions record with non-zero start offsets
### dlt version

1.1.0

### Source name

dlt init kafka snowflake

### Describe the problem

By default, kafka resource starts consuming from offset 0 and fetches up to 3000 records at a time when you do not set the `start_from`.

https://github.com/dlt-hub/verified-sources/blob/59749e6829180804f966a61665d156cf64272411/sources/kafka/__init__.py#L87-L89

I have a partition that only retains data for 7 days, and its starting offset is now 103,445. Due to the current implementation, kafka resource is unable to read any records from this partition, because it attempts to read from offset 0, fetches no records, and then exits.

### Expected behavior

It should adjust to the real starting offset and reading data from there.

### Steps to reproduce

Use Kafka resources to run any partition where the start offset is 3000 or higher.

If there is no similar partition, can manually simulate one by following the steps provided [here](https://dev.to/de_maric/how-to-delete-records-from-a-kafka-topic-464g#kafkadeleterecords).

### Operating system

Linux, macOS

### Runtime environment

Kubernetes

### Python version

3.11

### dlt destination

Snowflake

### Additional information

PR is incoming.

kang8:
### Tell us what you do here

### Short description

See #569
### Related Issues

- Fixes #569

### DIFF
--- a/sources/kafka/helpers.py
+++ b/sources/kafka/helpers.py
@@ -134,7 +134,7 @@ class OffsetTracker(dict):  # type: ignore
 
             # designate current and maximum offsets for every partition
             for i, part in enumerate(parts):
-                max_offset = self._consumer.get_watermark_offsets(part)[1]
+                [min_offset, max_offset] = self._consumer.get_watermark_offsets(part)
 
                 if start_from is not None:
                     if ts_offsets[i].offset != -1:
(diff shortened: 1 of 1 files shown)


kang8:
I realized that I made a mistake regarding this issue. See the [issue comment](https://github.com/dlt-hub/verified-sources/issues/569#issuecomment-2380624897) for more details.

PR #592, opened 2025-01-12
mongodb: Allow projection and mongoarrow schema

### LINKED ISSUE #577: Add support for pymongoarrow schema and projection parameters in the mongo CollectionLoaders
### Source name

mongodb

There are 2 parameters that would be useful to be user-configurable in the mongo `CollectionLoaders`:
1. A `projection` parameter for `find_raw_batches`/`find`, which allows to optionally limit which data will be exported from mongo (e.g. to remove at the source columns with sensitive data/reduce data size if not all columns are needed)
2. A `pymongoarrow_schema` to be used in PyMongoArrowContext to enforce a schema in `process_bson_stream` in case it is needed. This means that instead of the current call with the schema set to `None ` as done in
        `context = PyMongoArrowContext.from_schema(None, codec_options=self.collection.codec_options)` , one would be able to use a pymongoarrow schema:
    ```
            pymongoarrow_schema = pymongoarrow.api.Schema(arrow_schema)
            context = PyMongoArrowContext.from_schema(pymongoarrow_schema, codec_options=self.collection.codec_options)
    ```
    Without this schema, in one of our use cases `data_item_format = "arrow"` fails with the error `extraction of resource transaction in generator collection_documents caused an exception: value too large to convert to int32_t`. This error is due to the fact that the schema is wrongly inferred to be int32, but setting pyarrow type `pa.float64()` in the `pymongoarrow_schema `things work as expected
    

### Are you a dlt user?

I'm considering using dlt, but this bug is preventing this.

Yes, I'm ready.

### dlt destination

duckdb/s3

### Additional information

_No response_

zilto:
## Tell us what you do here
- improving, documenting, or customizing an existing source (#577)

## Short description

This implements two sets of features:
- Improve default behavior for `objectId` type and add `pymongoarrow_schema` for a user-facing type-casting API for `mongodb -> pyarrow`
- Add `projection` kwarg: allows to select the fields to include/exclude when pulling the data from MongoDB. This reduces egress volume and helps with regulatory compliance (e.g., don't pull sensitive information)

## Notes
- Using `pymongoarrow_schema` can intersect with the column selection in `projection`. The `pymongoarrow_schema` overrides the projection kwarg.
- The `pymongoarrow_schema` will "fail silently" if you cast to an invalid type (e.g., `ObjectId` to `int`): https://github.com/mongodb-labs/mongo-arrow/issues/246
- `ObjectId` is represented as a `binary[12]` type in Arrow using `pymongoarrow`. There's no vectorized operation to convert this to a string since the buffer contains ASCII characters.
- Invalid types in `pymongoarrow_schema` will make `pipeline.run()` fail with an hard to trace error `Out of buffer` coming from arrow (e.g., `_id: pyarrow.string()`.
  ```python
  E           dlt.pipeline.exceptions.PipelineStepFailed: Pipeline execution failed at stage load when processing package 1736799421.5290868 with exception:
  E           
  E           <class 'dlt.load.exceptions.LoadClientJobRetry'>
  E           Job for movies.e27d2e9dab.parquet had 5 retries which a multiple of 5. Exiting retry loop. You can still rerun the load package to retry this job. Last failure message was Out of buffer
  ```
- `pymongoarrow.schema.Schema` validates that its type annotations are valid a `__init__()`

## Future work
- As `pymongoarrow` is updated, we can raise failed type casting.
- Vectorize the operation to cast `ObjectId` to string. 
- If we can't vectorize type casting from `ObjectId` to `string`, then we can't efficiently handle nested arrays and structs without looping through each

## Open questions

1. ~~`projection` was added to `mongodb_collection` resource. Should it be added to `mongodb` source? Then, the projection would be applied to all the generated resources.~~
2. ~~`projection` + incremental loading: should we raise an exception if the `primary_key` is excluded? or force its inclusion and log a warning (current implementation)?.~~
3. ~~`pymongoarrow_schema` is applied at extract-time and it's hard to surface schema issues to the user. How much validation needs to happen within the dlt source code vs. user code?~~
4. ~~Should we upgrade from `pymongoarrow == 1.4.0` to `== 1.6.0` given their better type support?~~

### Related Issues

- Fixes #577

### DIFF
--- a/sources/mongodb/__init__.py
+++ b/sources/mongodb/__init__.py
@@ -1,6 +1,6 @@
 """Source that loads collections form any a mongo database, supports incremental loads."""
 
-from typing import Any, Dict, Iterable, List, Optional
+from typing import Any, Dict, Iterable, List, Optional, Union, Mapping
 
 import dlt
 from dlt.common.data_writers import TDataItemFormat
@@ -74,6 +74,7 @@ def mongodb(
             parallel=parallel,
             limit=limit,
             filter_=filter_ or {},
+            projection=None,
         )
 
 
@@ -93,6 +94,7 @@ def mongodb_collection(
     chunk_size: Optional[int] = 10000,
     data_item_format: Optional[TDataItemFormat] = "object",
     filter_: Optional[Dict[str, Any]] = None,
+    projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
 ) -> Any:
     """
     A DLT source which loads a collection from a mongo database using PyMongo.
@@ -112,6 +114,12 @@ def mongodb_collection(
                 object - Python objects (dicts, lists).
                 arrow - Apache Arrow tables.
         filter_ (Optional[Dict[str, Any]]): The filter to apply to the collection.
+        projection: (Optional[Union[Mapping[str, Any], Iterable[str]]]): The projection to select columns
+            when loading the collection. Supported inputs:
+                include (list) - ["year", "title"]
+                include (dict) - {"year": 1, "title": 1}
+                exclude (dict) - {"released": 0, "runtime": 0}
+            Note: Can't mix include and exclude statements '{"title": 1, "released": 0}`
 
     Returns:
         Iterable[DltResource]: A list of DLT resources for each collection to be loaded.
@@ -139,4 +147,5 @@ def mongodb_collection(
         chunk_size=chunk_size,
         data_item_format=data_item_format,
         filter_=filter_ or {},
+        projection=projection,
     )
--- a/sources/mongodb/__init__.py
+++ b/sources/mongodb/__init__.py
@@ -25,6 +25,8 @@ def mongodb(
     parallel: Optional[bool] = dlt.config.value,
     limit: Optional[int] = None,
     filter_: Optional[Dict[str, Any]] = None,
+    projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
+    pymongoarrow_schema: Optional[Any] = None
 ) -> Iterable[DltResource]:
     """
     A DLT source which loads data from a mongo database using PyMongo.
@@ -42,6 +44,13 @@ def mongodb(
             The maximum number of documents to load. The limit is
             applied to each requested collection separately.
         filter_ (Optional[Dict[str, Any]]): The filter to apply to the collection.
+        projection: (Optional[Union[Mapping[str, Any], Iterable[str]]]): The projection to select fields of a collection
+            when loading the collection. Supported inputs:
+                include (list) - ["year", "title"]
+                include (dict) - {"year": True, "title": True}
+                exclude (dict) - {"released": False, "runtime": False}
+            Note: Can't mix include and exclude statements '{"title": True, "released": False}`
+        pymongoarrow_schema (pymongoarrow.schema.Schema): Mapping of expected field types of a collection to convert BSON to Arrow
 
     Returns:
         Iterable[DltResource]: A list of DLT resources for each collection to be loaded.
@@ -74,7 +83,8 @@ def mongodb(
             parallel=parallel,
             limit=limit,
             filter_=filter_ or {},
-            projection=None,
+            projection=projection,
+            pymongoarrow_schema=pymongoarrow_schema,
         )
 
 
@@ -94,7 +104,8 @@ def mongodb_collection(
     chunk_size: Optional[int] = 10000,
     data_item_format: Optional[TDataItemFormat] = "object",
     filter_: Optional[Dict[str, Any]] = None,
-    projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
+    projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = dlt.config.value,
+    pymongoarrow_schema: Optional[Any] = None
 ) -> Any:
     """
     A DLT source which loads a collection from a mongo database using PyMongo.
@@ -114,12 +125,13 @@ def mongodb_collection(
                 object - Python objects (dicts, lists).
                 arrow - Apache Arrow tables.
         filter_ (Optional[Dict[str, Any]]): The filter to apply to the collection.
-        projection: (Optional[Union[Mapping[str, Any], Iterable[str]]]): The projection to select columns
+        projection: (Optional[Union[Mapping[str, Any], Iterable[str]]]): The projection to select fields
             when loading the collection. Supported inputs:
                 include (list) - ["year", "title"]
-                include (dict) - {"year": 1, "title": 1}
-                exclude (dict) - {"released": 0, "runtime": 0}
-            Note: Can't mix include and exclude statements '{"title": 1, "released": 0}`
+                include (dict) - {"year": True, "title": True}
+                exclude (dict) - {"released": False, "runtime": False}
+            Note: Can't mix include and exclude statements '{"title": True, "released": False}`
+        pymongoarrow_schema (pymongoarrow.schema.Schema): Mapping of expected field types to convert BSON to Arrow
 
     Returns:
         Iterable[DltResource]: A list of DLT resources for each collection to be loaded.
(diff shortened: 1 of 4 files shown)


sources/mongodb/helpers.py
```diff
@@ -106,6 +107,41 @@ def _filter_op(self) -> Dict[str, Any]:
...
+
+        Returns:
+            Tuple[str, ...] | Dict[str, Any]: A tuple or dictionary with the projection operator.
+        """
+        if projection is None:
+            return None
+
+        projection_dict = dict(_fields_list_to_dict(projection, "projection"))
+
+        # NOTE we can still filter on primary_key if it's excluded from projection
```
rudolfix, line 126: Mongo has "_id" column which is a kind of primary key. Currently when we create a resource for a collection, we set the primary_key to "_id". if that field is removed, dlt will later complain that primary key is not assuming any value so I'd add it back with a warning (like you do below). that we can do way earlier (when we create a resource)

In very rare cases when someone sets the primary key to something else (on resource / incremental) your code blow is fine. Adding back primary key and warning is good.

cool you thought about this edge case!

and in 99% of cases this is what you'll find in the incremental. Now the question is if

rudolfix (commented):
this looks really good! and thanks for writing tests.
your questions:
1 - 2: answered in the review

`pymongoarrow_schema` is not yet committed so I'll guess the answer to 3:
- I think you can reuse validation for projection to also validate columns in the mongo schema.
- this is really power user feature so we do not need to do a deep validation
- it is OK to fail in extract phase but we always try to provide useful exception messages for the most common case. If you write test properly you'll have those cases there ie. schema has some crazy type mismatch or refers to unknow columns - just wrap those in 1-2 meaningful exception (`Out of buffer` is not really helpful)

Yeah please upgrade the pymongoarrow - AFAIK the newest arrow is handling UUIDs as a type so maybe we do not need to do this crazy duckdb conversion from the other ticket

sources/mongodb/__init__.py
```diff
@@ -90,6 +91,7 @@ def mongodb_collection(
     chunk_size: Optional[int] = 10000,
     data_item_format: Optional[TDataItemFormat] = "object",
     filter_: Optional[Dict[str, Any]] = None,
+    projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
```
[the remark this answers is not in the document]
zilto: > 3. What is the reason for not mixing exclude and include projections?

The `projection` argument is implemented the same as `pymongo` (so users can copy-paste from their existing code). AFAIK, the query engine won't let you specify "include" and "exclude" clauses at once. It's either / or, so allowing two kwargs `include_projection` and `exclude_projection` would make it ambiguous which is used

zilto:
# Update and PR summary
## Original issue
This follows [this issue](https://github.com/dlt-hub/verified-sources/issues/577) which is specific to MongoDB source with pyarrow data loading, which uses `pymongoarrow`.

## TODOs
1. convert MongoDB's `ObjectId` type to dlt `data_type="string"`, which would be `pyarrow.string()` without looping over rows.
2. add `projection` kwarg to `mongodb_collection` to include/exclude columns. This is a native MongoDB operation available through `pymongo`
3. add `pymongoarrow_schema` kwarg to cast types when data is loaded to pyarrow format via `pymongoarrow`

## This PR
1. It's not possible because `ObjectId` is represented as a `pyarrow.FixedSizeBinary[12]` in arrow that contains ASCII characters. Therefore, it can't be casted to `pyarrow.String()`. Looping in Python or Pyarrow values is seemingly equivalent.

2. This is fully implemented. The official specs describe [3 ways to specify projections](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/read/project/#projection-types):
  a. **include**: only the specified fields are included (implicitly include `_id`)
  b. **include without** `_id`: like include, but explicitly exclude `_id`
  c. **exclude**: only the fields not specified are included
We support list and dictionary inputs like `pymongo`, and `pymongo` does the clause validation. We properly support nested projections and tests were added.

3.  `pymongoarrow_schema` is intended to cast MongoDB types to `pyarrow` types (this is akin to the core dbt conversion to arrow logic). This would be before the dlt mongodb specific type narrowing. The issue is that `pymongoarrow_schema` doesn't throw exceptions to catch if it fails to convert an arrow type. It will silently convert things to nan and nulls https://github.com/mongodb-labs/mongo-arrow/issues/246. The feature is implemented, but should be marked as experimental. I don't know what to test since this type conversion happens internally before a second conversion by dlt.

## Other notes
- the original issue reported with `int32` and `float64` seems to be at the `pymongoarrow` inference level. Our role is to provide a path for the user to manually correct `pymongoarrow` errors.
- `pymongoarrow` is rapidly evolving, but only supports the latest `pyarrow` version. e.g., they recently added support for `decimal128`. However, this conflicts with dlt core upper bound for `pyarrow`
- since it's impossible to vectorize casting `ObjectId` to string, it's also very difficult to handle nested objects that contain `ObjectId` types. The solution is to keep them as `binary` type for the unnesting, and potentially do all type casting after normalization.
- the `convert_arrow_columns` logic could extend the dlt core arrow conversion logic for maintainability

sources/mongodb/helpers.py
```diff
@@ -283,11 +361,68 @@ class CollectionArrowLoaderParallel(CollectionLoaderParallel):
...
+        limit: Optional[int] = None,
+        projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
+        pymongoarrow_schema: Any = None,
+    ) -> Iterator[TDataItem]:
+        """Load documents from the collection in parallel.
+
+        Args:
+            filter_ (Dict[str, Any]): The filter to apply to the collection.
+            limit (Optional[int]): The number of documents to load.
+            projection: selection of fields to create Cursor
```
anuunchin, line 376: typing 👀

anuunchin: will not repeat this comment anymore

sources/mongodb/helpers.py
```diff
@@ -348,6 +490,13 @@ def collection_documents(
             Supported formats:
                 object - Python objects (dicts, lists).
                 arrow - Apache Arrow tables.
+        projection: (Optional[Union[Mapping[str, Any], Iterable[str]]]): The projection to select fields
```
anuunchin, line 493: should this be moved above? - just to match the order of arguments, but probably not too important

sources/mongodb/helpers.py
```diff
@@ -372,11 +533,26 @@ def collection_documents(
...
+        yield from loader.load_documents(
+            limit=limit, filter_=filter_, projection=projection
+        )
 
 
 def convert_mongo_objs(value: Any) -> Any:
+    """MongoDB to dlt type conversion when using Python loaders.
+
+    Notes:
+        The method `ObjectId.__str__()` creates an hexstring using `binascii.hexlify(__id).decode()`
```
anuunchin, line 553:
```suggestion
        The method `ObjectId.__str__()` creates a hex string using `binascii.hexlify(__id).decode()`
```

tests/mongodb/test_mongodb_source.py
```diff
@@ -409,6 +409,113 @@ def test_filter_intersect(destination_name):
         pipeline.run(movies)
 
 
+@pytest.mark.parametrize("destination_name", ALL_DESTINATIONS)
+def test_projection_list_inclusion(destination_name):
+    pipeline = dlt.pipeline(
+        pipeline_name="mongodb_test",
+        destination=destination_name,
+        dataset_name="mongodb_test_data",
+        full_refresh=True,
```
anuunchin, line 418: I think full_refresh in deprecated, needs to be replaced with dev_mode=True in all tests

anuunchin (dismissed):

rudolfix (changes requested):
@anuunchin we do not have any tests for `pymongoarrow_schema` is there a reason?

rudolfix (approved):
this is so good. thx @anuunchin and @zilto

PR #601, opened 2025-02-28
swithch to uv

rudolfix:
### Tell us what you do here
this PR will drop `poetry` and switch to `uv` for project/dependency management
- also bumps `dlt` to `1.7.x` and fixes some issues
- also fixes how configuration is handled in `mongodb` (so it behaves exactly like a regular resource)

NOTE: FORKS workflow will not pass because we need to merge it into master for the changes to be visible...


.github/workflows/test_all_sources_nigthly.yml
anuunchin, line 1: Wondering why this is entirely removed

PR #604, opened 2025-03-11
schema pg replication

rudolfix:
### Tell us what you do here
1. allows to create a publication for a schema and create initial snapshots via list of tables taken from `sql_database`
2. skips dlt tables from being replicated
3. a test that demonstrates schema replication and auto-adding of new tables


sources/pg_replication/helpers.py
```diff
@@ -340,17 +404,19 @@ def persist_snapshot_table(
     col_str = "*"
     if include_columns is not None:
         col_str = ", ".join(map(escape_postgres_identifier, include_columns))
-    snapshot_table_name = f"{table_name}_snapshot_{snapshot_name}"
+    # make sure to shorten identifier
+    naming = NamingConvention(63)
```
VioletM, line 408: Just a small comment, let's put this value to some constant

VioletM (approved):
Looks good, works for me locally, added small comment. Also tests are failing

PR #606, opened 2025-04-02
rest_api -> sql_database

djudjuu:


burnash (approved):
Thanks for fixing this @djudjuu

PR #607, opened 2025-04-03
Adding Kafka end_time support for batch processing and SASL optionality. 

lfagliano:
### Tell us what you do here

### Short description

This PR enhances the Kafka source by adding an end_time parameter to the kafka_consumer function, allowing users to specify an upper time boundary for message consumption. This complements the existing start_from parameter, providing more precise control over the time range of messages to be consumed, allowing for batch processing to be idempotent. 

### Related Issues

Key changes:
1. Added end_time parameter to limit message consumption to a specific time
2. Made SASL authentication parameters optional in KafkaCredentials

### Additional Context

- The changes maintain backward compatibility - existing code using the Kafka source will continue to work without modification

### DIFF
--- a/sources/kafka/__init__.py
+++ b/sources/kafka/__init__.py
@@ -35,6 +35,7 @@ def kafka_consumer(
     batch_size: Optional[int] = 3000,
     batch_timeout: Optional[int] = 3,
     start_from: Optional[TAnyDateTime] = None,
+    end_time: Optional[TAnyDateTime] = None,
 ) -> Iterable[TDataItem]:
     """Extract recent messages from the given Kafka topics.
 
@@ -78,7 +79,18 @@ def kafka_consumer(
     if start_from is not None:
         start_from = ensure_pendulum_datetime(start_from)
 
-    tracker = OffsetTracker(consumer, topics, dlt.current.resource_state(), start_from)
+    if end_time is not None:
+        end_time = ensure_pendulum_datetime(end_time)
+
+    if end_time is not None and start_from is None:
+        raise ValueError("`start_from` must be provided if `end_time` is provided")
+    elif end_time is not None and start_from is not None:
+        if start_from > end_time:
+            raise ValueError("`start_from` must be before `end_time`")
+
+    tracker = OffsetTracker(
+        consumer, topics, dlt.current.resource_state(), start_from, end_time
+    )
 
     # read messages up to the maximum offsets,
     # not waiting for new messages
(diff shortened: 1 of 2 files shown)


sources/kafka/__init__.py
```diff
@@ -78,7 +79,18 @@ def kafka_consumer(
     if start_from is not None:
         start_from = ensure_pendulum_datetime(start_from)
 
-    tracker = OffsetTracker(consumer, topics, dlt.current.resource_state(), start_from)
+    if end_time is not None:
```
VioletM, line 82: Just thinking out loud Would this be more readable?

```
if end_time is not None:
    end_time = ensure_pendulum_datetime(end_time)

    if start_from is None:
        raise ValueError("`start_from` must be provided if `end_time` is provided")

    if start_from > end_time:
        raise ValueError("`start_from` must be before `end_time`")
```

sources/kafka/__init__.py
```diff
@@ -35,6 +35,7 @@ def kafka_consumer(
     batch_size: Optional[int] = 3000,
     batch_timeout: Optional[int] = 3,
     start_from: Optional[TAnyDateTime] = None,
+    end_time: Optional[TAnyDateTime] = None,
```
VioletM, line 38: Also needs to be described in docs!

sources/kafka/helpers.py
```diff
@@ -128,27 +135,38 @@ def _init_partition_offsets(self, start_from: pendulum.DateTime) -> None:
                 for part in topic.partitions
             ]
 
-            # get offsets for the timestamp, if given
-            if start_from is not None:
-                ts_offsets = self._consumer.offsets_for_times(parts)
+            # get offsets for the timestamp ranges, if given
+            if start_from is not None and end_time is not None:
```
VioletM, line 139: What if 
```
(start_from is not None) and (end_time is None)
```

VioletM (changes requested):
Thanks a lot for contributing!
Left a couple of comments :)

sources/kafka/helpers.py
lfagliano: Hmm this one I am not that sure, I dont think it really makes it more readable (it is a simple `AND`). I made a quick check on the folder as well, and I think it all without the parenthesis (in `helpers.py ` for instance.

sources/kafka/__init__.py
lfagliano: Added!

sources/kafka/__init__.py
lfagliano: True! Nice catch, adding it now

sources/kafka/helpers.py
VioletM: oh, sorry, I wasn't clear. I meant what would happen in the case if `start_from` is specified and `end_time` is None?
It seems that this case is not covered

lfagliano: Done! I added back the original fall back scenario when start_time is the only param provided. Basically we read from that start to the max offset. Let me know your thoughts!

VioletM (approved):

lfagliano:
@VioletM Do I need to do anything re the lint failures? Looks like a problem in the postgres source

PR #608, opened 2025-04-15
feat: add reference LLM-enabled IDE setup files

zilto:
Relates to PR https://github.com/dlt-hub/dlt/pull/2503


rudolfix (dismissed):
LGTM! should we update our README on how to change / add more rules also for other editors?

PR #611, opened 2025-04-23
in dlt changed

anuunchin:
The location of the exception changed in dlt, and test utils were importing from the old location.


anuunchin:
This is a separate PR because I wanted to see smth in the ci

PR #613, opened 2025-05-06
fix: tests import exception from correct location

djudjuu:
tests did no longer run because they were trying to import this exception from a wrong location.
after moving it, most passed except the workflow `tests dlt init compatibility`.

with [refactors iceberg and duckdb cache support](https://github.com/dlt-hub/dlt/pull/2430/files#top) 
the exception imported in the utils file moved.
most workflows worked correctly, except the one in `init.yml` which didn't do make dev but only installed `dltpure`-dep-group 
```
[dependency-groups]
dltpure = ["dlt>=1.7.0"]
```
I dont understand why this didnt pick up the latest then, but apparrenlty it didnt. changing it to 1.10. works though

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -26,7 +26,7 @@ from dlt.common.storages import FileStorage
 from dlt.common.schema.typing import TTableSchema
 from dlt.common.utils import set_working_dir
 
-from dlt.pipeline.exceptions import SqlClientNotAvailable
+from dlt.common.destination.exceptions import SqlClientNotAvailable
 
 TEST_STORAGE_ROOT = "_storage"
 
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -26,7 +26,7 @@
 from dlt.common.schema.typing import TTableSchema
 from dlt.common.utils import set_working_dir
 
-from dlt.pipeline.exceptions import SqlClientNotAvailable
+from dlt.common.destination.exceptions import SqlClientNotAvailable
```
rudolfix, line 29: this is ok because this is a part of tests! just remember that we need to be backward compatible with old dlt versions in the main code so you'd need to try both options there

rudolfix (approved):

djudjuu:
what can I do to make that pass?

PR #614, opened 2025-05-12
fix(inbox): senders filter is optional

joscha:
the function (`inbox_source`) this argument is passed to, has this parameter optional. It's the first error one runs into as most likely most inboxes do not have a sender of `dlthub@dlthub.com`, meaning the generated sample code always yields an empty result.


rudolfix (approved):
LGTM!

PR #625, opened 2025-05-23
Handling of trailing undetected empty columns

### LINKED ISSUE #619: Google Sheets Columns Fetching Issue
### dlt version

dlt 1.8.1

### Describe the problem

When fetching the data from Google Sheets, if the last columns do not have a value in the first row, they will not be fetched...
until a last column with a value in the first row is added, then the columns will be fetched
so in short, the last column in the sheet should have a value in its first row to be able to bring all the data in the sheet...  I have a header for the column, but the first value of the column (after the header) is empty

am I missing something?

Case 1:
Image

Image

Case 2:

Image

Image

Case 3:

Image

Image

My code:
```
 try:
        range_names = []
            
        url = "https://docs.google.com/spreadsheets/d/1MOm8eaIFbpBf8wEUU4YjqAn7dqeWvbn5sliK1bk5uyo/edit?usp=sharing"
        print("load_pipeline_with_sheets")
        load_data = google_spreadsheet(
            url,
            get_sheets=True,
            get_named_ranges=False,
            range_names=range_names,

        )
        pipeline = dlt.pipeline(
            pipeline_name="DLT_google_sheets_pipeline",
            destination='snowflake',
            dataset_name="google_sheet_testing",
        )
        load_info = pipeline.run(load_data)
        print(load_info)
    except Exception as e:
        print(e)

```
### Expected behavior

Get all the data in the sheet

### Steps to reproduce

In your Google Sheet, create some columns and the last column keep the first row of it empty, and then use DLT to get the sheet, then the last column will not be fetched even if it has data in the other rows (but not the first row)

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

Google Sheets

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
### Description

This PR adjusts how the `process_range` function handles trailing empty column entries that are silently dropped in the zipping operation of the inner loop. Before handling the row values, the function now aligns the row with explicit empty values and adjusts the data_types list with None values accordingly. A test was added with cases provided by the user.

### Related Issues

- Fixes #619

### DIFF
--- a/sources/google_sheets/helpers/data_processing.py
+++ b/sources/google_sheets/helpers/data_processing.py
@@ -251,6 +251,9 @@ def process_range(
         # empty row; skip
         if not row:
             continue
+        # align trailing empty columns
+        data_types += [None] * (len(headers) - len(row))
+        row += [""] * (len(headers) - len(row))
         table_dict = {}
         # process both rows and check for differences to spot dates
         for val, header, data_type in zip(row, headers, data_types):
--- a/sources/pipedrive/__init__.py
+++ b/sources/pipedrive/__init__.py
@@ -174,7 +174,7 @@ def parsed_mapping(
 
 
 @dlt.resource(primary_key="id", write_disposition="merge")
-def  leads(
+def leads(
     pipedrive_api_key: str = dlt.secrets.value,
     update_time: dlt.sources.incremental[str] = dlt.sources.incremental(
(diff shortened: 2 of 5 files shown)


rudolfix (dismissed):
LGTM!

ShreyasGS (dismissed):

PR #634, opened 2025-07-18
Fix Google Sheets source to not dropping columns if 1st value is null

### LINKED ISSUE #591: in Google Sheets column droped if first value is null
### dlt version

1.4.1

### Source name

google_sheets

### Describe the problem

**Issue:**

When working with this table in Google Sheets:

| Col0 | Col1 |
|------|------|
| a    |      |
| b    | c    |

The second column (`Col1`) droped by source.

**What should happen:**

Both columns (`Col0` and `Col1`) should keep their data as entered.

**What actually happens:**

The second column (`Col1`) gets dropped.

### Expected behavior

_No response_

### Steps to reproduce

load this csv to google sheets
col0, col1
a,	
b, c
then read it with
google_spreadsheet(
        spreadsheet_url_or_id=url,
        credentials=creds,
        range_names=["list1!a1:b3"]
    )

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12.3

### dlt destination

snowflake

### Additional information

_No response_

franloza:
### Tell us what you do here

I'm fixing a bug in Google Sheets that causes a column dropped if first value is null

### Short description

Fixed an issue where Google Sheets columns were being dropped when the first data row contained null/empty values in those columns. Added a new `metadata_scan_rows` parameter to the `google_spreadsheet` source that allows scanning multiple data rows for metadata detection, ensuring all columns are preserved even when early rows have sparse data.

The fix maintains full backwards compatibility by defaulting `metadata_scan_rows=1`, while allowing users to increase it for better metadata detection in spreadsheets with sparse data patterns.

Key changes:
- Added `metadata_scan_rows` parameter to `google_spreadsheet()` function
- Created `get_data_types_from_multiple_rows()` helper function
- Updated metadata processing logic to scan multiple rows and pad results to match header count
- Added comprehensive tests to verify column preservation
- Improved error handling and user guidance

### Related Issues

- Fixes #591

### Additional Context

This was a bug affecting users with spreadsheets containing empty cells in the first data row. The issue occurred because the metadata detection only examined the first row, causing columns with null values to be completely omitted from the extracted data.

The solution scans the specified number of data rows (`metadata_scan_rows`) to build a complete picture of all columns, ensuring that even columns missing from early rows are preserved in the final output. Users experiencing this issue can now simply increase `metadata_scan_rows` to a higher value (e.g., 5-10) for robust column detection.

All existing functionality remains unchanged, making this a safe upgrade for all users.

### DIFF
--- a/sources/google_sheets/__init__.py
+++ b/sources/google_sheets/__init__.py
@@ -11,6 +11,7 @@ from .helpers import api_calls
 from .helpers.api_calls import api_auth
 from .helpers.data_processing import (
     get_data_types,
+    get_data_types_from_multiple_rows,
     get_range_headers,
     get_spreadsheet_id,
     process_range,
@@ -27,6 +28,7 @@ def google_spreadsheet(
     get_sheets: bool = False,
     get_named_ranges: bool = True,
     max_api_retries: int = 5,
+    metadata_scan_rows: int = 1,
 ) -> Iterable[DltResource]:
     """
     The source for the dlt pipeline. It returns the following resources:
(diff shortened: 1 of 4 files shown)


franloza:
I'm not sure this PR is needed. More context here:  https://github.com/dlt-hub/verified-sources/issues/591#issuecomment-3090016373

Issue #639, opened 2025-08-01
Salesforce OAuth w/ simple_salesforce


neal301:
### Documentation description

DLT uses the excellent simple_salesforce api under the hood to access salesforce data. Most production deployments of salesforce use SSO for verification, physical passwords are mostly a thing of the past and ESPECIALLY when it comes to authentication in deployed data pipelines.

Simple salesforce supports OAuth, I have been able to get this functionality to work on my end with our salesforce deployment, albeit on our sandbox instance (simply not ready to deploy DLT yet, still testing).

This requires a good bit of config on the salesforce end of things and does require admin privileges to work properly. I am happy to provide my notes here:

Oauth examples for simple_salesforce found here:
https://pypi.org/project/simple-salesforce/

The method I personally opted to use (read: was able to get to work) is the private key connected app method:
`from simple_salesforce import Salesforce
sf = Salesforce(username='myemail@example.com', consumer_key='XYZ', privatekey_file='filename.key')`

On the salesforce end, one must create a new "EXTERNAL CLIENT APP" and give it access to:

<img width="1786" height="670" alt="Image" src="https://github.com/user-attachments/assets/df05a0bb-37f4-4aad-a620-433204d9b0b3" />

The consumer_key is found by clicking the link shown above.

One must also create a key using openssl:
`openssl req -x509 -sha256 -nodes -days 36500 -newkey rsa:2048 -keyout salesforce.key -out salesforce.crt
`

I believe the certificate must be uploaded to the external app. Then, in python as shown above, the path to the private key must be given.

Aside from all of the ssl mess, once that is done, it is actually quite simple to switch to oauth using simple salesforce - we just have to modify the salesforce_source() function to take in the required arguments as shown in the image above instead of username and password and the other secret string it required (dont remember).

VERY VERY IMPORTANT NOTE THAT TOOK ME OVER 2 HOURS TO FIGURE OUT (salesforce noob here):
If you are connecting to ANY SANDBOX INSTANCE OF SALESFORCE (if your company is using dev, or test, or QA, or UAT its LIKELY A SANDBOX) you MUST ENTER AN ADDITINAL ARGUMENT IN THE FUNCTION TO CREATE THE SALESFORCE CLIENT - but it couldn't be any easier:

`domain='test'` - THAT'S IT!

### Are you a dlt user?

Yes, I'm already a dlt user.

PR #641, opened 2025-08-07
Feat: Salesforce different auth methods

anuunchin:
This PR enables different auth methods for the salesforce source that are supported by the underlying simple_salesforce package.

Additionally, the standard api is enabled if bulk api is not available in Salesforce - this applies to Salesforce trial accounts that users may test this source with.  

Relates to #639

### DIFF
--- a/sources/salesforce/README.md
+++ b/sources/salesforce/README.md
@@ -37,7 +37,7 @@ Here, we chose BigQuery as the destination. Alternatively, you can also choose r
 
 ## Setup verified source and pipeline example
 
-To grab credentials and initialize the verified source, please refer to the [full documentation here.](https://dlthub.com/docs/dlt-ecosystem/verified-sources/salesforce)
+To initialize the verified source, please refer to the [full documentation here.](https://dlthub.com/docs/dlt-ecosystem/verified-sources/salesforce)
 
 ## Add credentials and configs
 
@@ -51,7 +51,7 @@ To grab credentials and initialize the verified source, please refer to the [ful
     security_token = "please set me up!" # Salesforce security token generated
     ```
 
-    Alternatively, you can authenticate using one of the following methods:
+   Alternatively, you may authenticate using any method supported by the underlying `simple_salesforce` library. For instructions on obtaining these credentials, please consult Salesforce’s documentation.
 
     - IP-whitelist + organization ID:
         ```toml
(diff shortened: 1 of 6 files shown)


anuunchin:
@neal301 Would be awesome if you test this out and see if the desired auth method works 🙂

neal301:
@anuunchin Hey! I've actually been working on testing & implementing this myself, and would love to contribute if my solution works!

I can confirm that I have gotten Oauth to work with this, I'm just debugging switching the Bulk API to Bulk2.0, Bulk is causing some malformed requests when I try to query

sources/salesforce/README.md
```diff
@@ -39,19 +39,74 @@ Here, we chose BigQuery as the destination. Alternatively, you can also choose r
...
     ```toml
     # put your secret values and credentials here. do not share this file and do not push it to github
-    [sources.salesforce]
+    [sources.salesforce.credentials]
     user_name = "please set me up!" # Salesforce user name
     password = "please set me up!" # Salesforce password
     security_token = "please set me up!" # Salesforce security token generated
     ```
+
+    Alternatively, you can authenticate using one of the following methods:
```
djudjuu, line 54: is there a link to hubsport where the user can read up on those methods? (or me)
just feel this woulb be useful context

neal301: This is for salesforce, not hubspot. If you are curious about salesforce methods, the documentation is here on github. The package DLT uses is called simple_salesforce

sources/salesforce/helpers/records.py
```diff
@@ -0,0 +1,121 @@
...
+                record[field] = pendulum.from_timestamp(
+                    record[field] / 1000,
+                ).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+            else:
+                # Standard API return ISO strings, ensure consistent format
+                if record[field]:
+                    # pendulum.parse can also return Date, Time or Duration
+                    # Salesforce date/datetime fields are always DateTime
+                    dt = cast(pendulum.DateTime, pendulum.parse(record[field]))
+                    record[field] = dt.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
djudjuu, line 45: so Z means UTC timezone. so we are converting all things to utc timezone here. is that intended?
not sure what would be a good solution here

djudjuu (dismissed):
good work!
just some minor adjustments then this is good to go.

and I am unsure about the timezones...

PR #657, opened 2025-09-18
Fix: Timestamp without timezone, regular json data types in pg_replication

### LINKED ISSUE #652: bug: pg_replication don't handle timestamp (no tz) and regular json data types
### dlt version

1.15.0

### Source name

pg_replication

### Describe the problem

`pg_replication` can't handle `json`(OID 114) and `timestamp`(OID 1114) data types.
If there is a column with json data type, it will be treated as text and the data in the destination will be written wrong.
If there is a column with timestamp data type, it will try to treat it as text, and it will crash during the `load` step (the destination won't be able to convert text to timestamp).

### Expected behavior

It will use the correct data type.

### Steps to reproduce

Create a table using these data types and try to replicate it (make sure to use `json`, not `jsonb`).

I'm considering using this source in my work, but bug is preventing this.

### Operating system

macOS

### Runtime environment

Kubernetes

### Python version

3.12.7

### dlt destination

Snowflake

### Additional information

_No response_

anuunchin:
This PR is a continuation of #652 that addresses the following problem: `pg_replication` source doesn't handle json (OID 114) and timestamp (OID 1114) data types correctly.

The types were added to the type mapper, along with appropriate tests.

Resolves #652

### DIFF
--- a/tests/pg_replication/test_pg_replication.py
+++ b/tests/pg_replication/test_pg_replication.py
@@ -6,7 +6,6 @@ from psycopg2.errors import InsufficientPrivilege
 
 import dlt
 from dlt.destinations.job_client_impl import SqlJobClientBase
-from pendulum import DateTime, timezone
 
 from tests.utils import (
     ALL_DESTINATIONS,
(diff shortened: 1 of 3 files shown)


tests/pg_replication/test_pg_replication.py
```diff
@@ -50,7 +50,10 @@ def tbl_y(data):
     snapshots = init_replication(
         slot_name=slot_name,
         pub_name=pub_name,
-        table_names=None,  # will initialize schema replication and get list of snapshots from sql_database
+        table_names=[
+            "tbl_x",
+            "tbl_y",
+        ],  # will initialize schema replication and get list of snapshots from sql_database, set to None if creds have superuser privileges
```
anuunchin, line 56: The adjustment in this specific test was added because the creds don't have super user rights, and we can't set table_names to none

PR #670, opened 2025-11-17
Remove default breakdowns and action_breakdowns from Facebook Ads insights source - #644

anuunchin:
This PR is a continuation of #644, which basically turns off breakdowns and action breakdowns that were enabled by default and were therefore confusing.

Additionally fixes some imports.


rudolfix (approved):
LGTM!

PR #680, opened 2026-01-20
[FEAT] support extracting all google_sheets values as text

acaruso7:
### Tell us what you do here

### Short description
support extracting all google_sheets values as text, represented exactly as in source spreadsheet

This is helpful for destinations where users wish to handle type casting themselves, for example via tools like dbt

Default behavior is unchanged; this option can be used by adding `extract_all_as_text=True` parameter to source

### Additional Context

https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1768500985331239


rudolfix (approved):
LGTM!

PR #686, opened 2026-05-27
Fix Pipedrive set custom field string mapping

srl-git:
### Tell us what you do here

- fixing a bug
- improving an existing source

### Short description

Fixes Pipedrive V1 custom field mapping for `set` fields returned as comma-separated strings.

Previously, values like `"62,63,65"` were treated as generic iterables, so the mapper iterated over each character and produced values like `"6"`, `"2"`, `","`, etc. This change normalizes string `set` values into option IDs before mapping them to readable option labels.

Added a regression test covering comma-separated Pipedrive `set` custom field values.

### Related Issues

No related issue. This was found while debugging Pipedrive custom field loading locally.


dat-a-man:
It is tested once and works as expected.

PR #687, opened 2026-05-28
Map Pipedrive V2 custom fields

### LINKED ISSUE #683: Pipedrive V2: custom field renaming equivalent to V1 `custom_fields_mapping`
### Source name

pipedrive

Are there plans to add custom field renaming support to the Pipedrive V2 source, similar to the V1 `custom_fields_mapping` resource?

In V1, `custom_fields_mapping` works well: custom field hash keys are mapped to readable field names, and enum/set option IDs can be mapped to labels.

In V2, custom fields currently land using raw Pipedrive hash/key names, for example:

```text
custom_fields__<hash>
custom_fields___<hash>
v2_deals__custom_fields___<hash>
```

For enum/set fields, nested tables contain option IDs rather than labels, so downstream SQL has to manually join back to field metadata to decode values.

This makes V2 difficult to use for analytics, even though V2 appears to be the better source for complete deal extracts. (V1 deals are loaded via `/v1/recents`, so this doesn't provide a complete historical snapshot)

Using V1 long-term also seems risky because Pipedrive has deprecated selected V1 endpoints now that V2 replacements exist. 

Desired V2 behavior:

- load field metadata from endpoints such as `dealFields`, `personFields`, `organizationFields`, etc.
- map custom field hash keys to readable field names
- map enum/set option IDs to labels
- ideally flatten/rename `custom_fields` before dlt normalization, so resulting tables use readable names instead of hashes

Would a PR adding V2 custom field mapping/renaming support be welcome?

### Are you a dlt user?

Yes, I'm already a dlt user.

Yes, but I need help.

### dlt destination

duckdb

### Additional information

_No response_

srl-git:
### Tell us what you do here

- improving, documenting, or customizing an existing source

### Short description

Adds optional custom field mapping for Pipedrive V2 resources.

When `map_custom_fields=True` is passed to `pipedrive_v2_source`, supported V2 resources fetch their field metadata from the corresponding V2 field endpoint and use it to rename hashed `custom_fields` keys into readable top-level columns. Enum and set option IDs are mapped to labels using the existing custom field mapping behavior.

The mapping is opt-in to avoid changing existing V2 destination schemas by default. Field mappings are also stored in source state, so if a Pipedrive custom field is renamed later, future loads continue writing to the original normalized column name instead of creating a new column.

### Related Issues

Fixes #683

### Additional Context

I have read the contributing guide and run the tests locally.

Local checks run:

- `uv run pytest tests/pipedrive/test_pipedrive_source.py`
- `make lint-code`

### DIFF
--- a/sources/pipedrive/helpers/custom_fields_munger.py
+++ b/sources/pipedrive/helpers/custom_fields_munger.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, TypedDict, Optional
+from typing import Any, Dict, TypedDict, Optional, cast
 
 import dlt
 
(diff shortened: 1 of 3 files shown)


dat-a-man:
it is tested once and works as expected.

PR #693, opened 2026-07-02
fix(shopify): remove stray debug print from partner GraphQL pagination

Pawansingh3889:
`ShopifyPartnerApi.get_graphql_pages` prints every raw API response page to stdout. For any real partner query that means pages of raw JSON echoed to the console, or to an orchestrator's logs, on every scheduled run — and partner transaction data isn't something to write into logs by default.

One line removed, no behaviour change otherwise. Spotted while building a Shopify pipeline with dlt and reading the verified source for reference.


rudolfix (approved):
LGTM! thanks!

